PR #13, opened 2025-10-08
add neptune logging support

mzio:
For folks that use [Neptune](https://docs.neptune.ai/api_overview), adds logging support automatically if  `NEPTUNE_API_TOKEN` is specified as an environment variable.

A bit hacky, but for now avoided doing a bigger refactor to be logger-agnostic (where you'd need to rename or add equivalents for all instances of `wandb_name` and `wandb_project`; probs worth discussing at some point)

Instead this just updates `tinker_cookbook/utils/ml_log.py`:
* Neptune uses the same W&B project and name
* `wandb_project` can be specified as `workspace-name/project-name` to work with Neptune [conventions](https://docs.neptune.ai/run/#:~:text=Name%20of%20a%20project,-in%20the%20form) 
* If `WANDB_API_KEY` also specified, can log to both W&B and Neptune 

Example:
```bash
python -m tinker_cookbook.recipes.math_rl.train \
  env=math \
  model_name="Qwen/Qwen3-8B" \
  group_size=16 \
  groups_per_batch=64 \
  learning_rate=2e-5 \
  max_tokens=512 \
  wandb_project="<workspace-name>/<project-name>"
  ```

### DIFF
--- a/tinker_cookbook/utils/ml_log.py
+++ b/tinker_cookbook/utils/ml_log.py
@@ -268,10 +268,8 @@ class NeptuneLogger(Logger):
     ) -> None:
         """Log metrics to neptune."""
         if self.run and NeptuneRun is not None:
-            if step is not None:
-                self.run.log_metrics(metrics, step=step)
-            else:
-                self.run.log_metrics(metrics)
+            assert step is not None, "step is required to be int or float for Neptune logging."
+            self.run.log_metrics(metrics, step=step)
             logger.info("Logging to: %s", self.run.get_run_url())
 
     def close(self) -> None:

### CODE REVIEW

**Comment by Tiiiger** [comment]
hi @mzio , let me figure out soon when we can merge public PRs

**Comment by Tiiiger** [comment]
ready to merge once if you can help rebase and run pyright precommit

**Comment by mzio** [comment]
TY! thank you for sharing the amazing repo and letting us all tinker!

File: pyproject.toml [hunk 1]
```diff
@@ -46,6 +46,9 @@ vector-search = [
 wandb = [
     "wandb",
 ]
+neptune_scale = [
```
sergei-something-ai, line 49: sorry for a late comment, just one nit: we use `vector-search` with `-`, so we should probably use `-`. I opened https://github.com/thinking-machines-lab/tinker-cookbook/pull/14

Issue #6, opened 2025-10-01
Would it be possible to get `trackio` experiment training logs for the finetuning runs?


abidlabs:
Thanks for the release of Tinker! Would you be open to adding a [Trackio](https://github.com/gradio-app/trackio) integration for logging and monitoring fine-tuning runs with Tinker?

Having access to training metrics would help researchers compare different runs (e.g. easily compare multiple fine-tuning experiments with different hyperparameters), debug training issues (identify problems like overfitting or gradient explosions early), and generally provide valuable insights to researchers to optimize their fine-tuning workflows.

Tiiiger:
Hi @abidlabs

Great suggestion!

We currently have a wandb integration and can be turned on in every recipe example. Will need to think a bit more between balancing cook book complexity and more integrations.

abidlabs:
Thanks for considering @Tiiiger! One benefit of Trackio is that it can run completely locally (without needing to make an account, etc.)

Tiiiger:
hi @abidlabs if you want to follow the pattern in #13 and help contribute a PR, I can help review and merge it

abidlabs:
thanks @Tiiiger here ya go! https://github.com/thinking-machines-lab/tinker-cookbook/pull/15

PR #15, opened 2025-10-09
Add Trackio integration for experiment tracking

abidlabs:
Adds logging support with [Trackio](https://github.com/gradio-app/trackio). Since Trackio runs locally, not api token needs to be specified. Logging with trackio happens if `trackio` is installed as a package. Summary of changes:

- Adds Trackio integration for experiment tracking and logging.
- Implements `TrackioLogger` class following the same pattern as the Neptune integration from PR #13
- Enables simultaneous logging to multiple platforms (W&B, Neptune, and Trackio)

Closes: #6

### DIFF
--- a/tinker_cookbook/utils/ml_log.py
+++ b/tinker_cookbook/utils/ml_log.py
@@ -315,7 +315,7 @@ class TrackioLogger(Logger):
 
         assert trackio is not None
         self.run = trackio.init(
-            project=project,
+            project=project or "default",
             name=trackio_name,
             config=dump_config(config) if config else None,
         )


tinker_cookbook/utils/ml_log.py
```diff
@@ -139,7 +148,7 @@ def __init__(self):
 
     def log_hparams(self, config: Any) -> None:
         """Print configuration summary."""
-        config_dict = chz.asdict(config)
+        config_dict = dump_config(config)
```
abidlabs, line 155: Unrelated, but all the other loggers use `dump_config`, so I changed this for consistency and because it works more generally. For example, when I   tested with a plain dictionary `({'learning_rate': 0.001,  'batch_size': 32, ...})`, it raised a `RuntimeError: {...} is not a  chz object`.

Tiiiger:
Will try locally over the weekend! sorry for the delay

abidlabs:
Hi @Tiiiger just a friendly ping, let me know if you have any feedback on the PR. Thanks again!

Tiiiger:
Apologies for the delay! I will fix the CI pattern first and test this out over the weekend ...

tinker_cookbook/utils/ml_log.py
Tiiiger: yea thanks! this is good

Tiiiger:
Thanks @abidlabs ! LGTM and I have tried quickly and it seems to work.

Happy to merge. Can you help:
1. update pyproject.toml to include trackio as an optional dependency? 
   note: #35 needs to be merged in first and after that it will install all optional dependencies.
2. pass pyright check (it should pass but there might be something minor)

Sorry again for the delay!

I'll review again after pyright check passes

abidlabs:
Should be all set @Tiiiger thanks so much!

Tiiiger (approved):

PR #17, opened 2025-10-10
Remove duplicate training_client in rl_loop

ekzhang:
Noticed as I was debugging another thing that this training client has been duplicated, so the program creates two LoRAs on initialization and only uses one.

Already reviewed in monorepo, that PR is closed as public repo is the source of truth.


ekzhang:
My GitHub is a bit wonky and can't request reviewers on this PR but could you give a ✔️  @devendrachaplot ?

ekzhang:
image

Hm I'm in the org but I can't merge

PR #24, opened 2025-10-14
[tinker-cookbook] utils: add lightweight tracing library

kennyyu:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.12.0) (oldest at bottom):
* #25
* __->__ #24

This adds a lightweight library to trace sync + async functions,
and generate perfetto-compatible traces. This helps with debugging
performance issues, especially for RL runs. The next PR uses this
to add tracing for RL runs.

See unit test for example usages. Implementation notes:
- Each track in the trace corresponds to a coroutine, and each group of tracks represents a group of coroutines managed by a specific thread.
- The `@scope` decorator works for both sync and async functions. We will log a span for the duration of this function call
- You can use `get_scope_context()` to modify attributes for this span to include additional debug info
- This also includes a script `tinker_cookbook/utils/trace.py` to convert the trace events jsonl format into a perfetto/chrome-tracing compatible format.

Example of a trace from the unit test:

Screenshot 2025-10-13 at 9 42 27 PM

### DIFF
--- a/tinker_cookbook/recipes/math_rl/math_env.py
+++ b/tinker_cookbook/recipes/math_rl/math_env.py
@@ -143,9 +143,10 @@ class MathDataset(RLDataset):
         renderer: renderers.Renderer,
         convo_prefix: list[renderers.Message] | None = None,
         split: Literal["train", "test"] = "train",
+        seed: int = 0,
     ):
         if split == "train":
-            self.ds = _get_hendrycks_math_train().shuffle(seed=0)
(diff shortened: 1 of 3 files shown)


kennyyu:
ah it seems `ghstack land` doesn't work with branch protection (https://github.com/ezyang/ghstack/issues/50) -- I'll reopen the PR without ghstack

kennyyu:
Replaced with https://github.com/thinking-machines-lab/tinker-cookbook/pull/28

PR #27, opened 2025-10-14
Environments Hub integration

willccbb:
This recipe allows using the Tinker RL trainer with environments from the Prime Intellect [Environments Hub](https://app.primeintellect.ai/dashboard/environments?ex_sort=by_sections), which features a few hundred community-contributed RL environment implementations built with the [verifiers](https://github.com/PrimeIntellect-ai/verifiers) library. 

**Structure:**

In `tinker_openai.py` we define a Tinker-backed AsyncOpenAI client which is used by `verifiers` environments during rollouts.

In `verifiers_env.py` we create an RLDatasetBuilder which contains a global Environment object (which can manage heavier resources, e.g. code sandboxes, MCP servers), and we return a dummy EnvGroupBuilder, which is ignored by our trainer patch for `do_group_rollouts`.

In `train.py` we adopt the basic setup from other recipes, but we patch the `do_rollout_group` method to inject our rollout logic (whose two key methods are `rollout` and `score_rollout`, both operating over dataset rows). Turn-level responses are collected with a hook for per-rollout AsyncOpenAI clients (with a shared SamplingClient), and we then construct a TrajectoryGroup from the resulting trajectories and rewards.

An optional dependency group ('verifiers') is added to the top-level `pyproject.toml`.

### DIFF
--- a/tinker_cookbook/recipes/math_rl/train.py
+++ b/tinker_cookbook/recipes/math_rl/train.py
@@ -140,6 +140,7 @@ async def cli_main(cli_config: CLIConfig):
         )
         if cli_config.max_steps_off_policy is not None
         else None,
+        loss_fn=cli_config.loss_fn,
     )
 
     cli_utils.check_log_dir(log_path, behavior_if_exists=cli_config.behavior_if_log_dir_exists)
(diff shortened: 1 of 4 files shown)


joschu (approved):
Looks great, thanks for the contribution.

It's unfortunate that this requires monkey-patching train.py, and I expect us to inadvertently break this occasionally. If this pattern starts to get popular -- i.e., defining a custom rollout function, rather than subclassing our Env, then we can make this more of a first-class feature.

PR #36, opened 2025-10-23
Integrate with GEM

lkevinzc:
Hi tinker maintainers,

Thanks for the great tool! I would like to contribute the integration of Tinker x GEM (a collection of diverse environments for LLM RL) in the PR. We provide training recipes for both using tinker-cookbook and using tinker directly.


Tiiiger:
Will test but let's fix pyright first ... thanks!

PR #37, opened 2025-10-24
Add logtree utility, use it to create HTML file from RL rollouts, including logging from the environments

joschu:
Add a new `logtree` module, which is a library for tracing-based logging, which allows you to create an html file with the same tree structure of the call tree. See attached screenshots from a few RL training scripts.

Each iteration, it writes a new file with a name like iteration_000000.html, containing a subset of the rollout (groups). You can put logging code into the Env.step or EnvGroupBuilder.compute_group_rewards, and it'll be appear in the right place in the html.

While I was at it, I switched a few recipe scripts to use the 4B instruct model instead of the 8B hybrid-thinking model -- this should improve speed and simplify some renderer issues.

Math (arithmetic):

Screenshot 2025-10-24 at 3 08 54 PM

20 questions:

Screenshot 2025-10-24 at 2 06 37 PM

RLHF with tournament pairwise comparisons:

Screenshot 2025-10-24 at 2 06 34 PM

### DIFF
--- a/tinker_cookbook/rl/preference_envs.py
+++ b/tinker_cookbook/rl/preference_envs.py
@@ -195,16 +195,6 @@ class PairwisePreferenceGroupBuilder(EnvGroupBuilder):
             matchup_count[i] += 1
         format_coef = 1.0
 
-        # Log final rewards
-        with logtree.scope_header("Final Rewards"):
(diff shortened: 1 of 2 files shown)


joschu:
Closing because I did something wrong with git

PR #39, opened 2025-10-25
Add logtree utility, use it to create HTML file from RL rollouts, including logging from the environments

joschu:
Add a new logtree module, which is a library for tracing-based logging, which allows you to create an html file with the same tree structure of the call tree. See attached screenshots from a few RL training scripts.

Each iteration, it writes a new file with a name like iteration_000000.html, containing a subset of the rollout (groups). You can put logging code into the Env.step or EnvGroupBuilder.compute_group_rewards, and it'll be appear in the right place in the html.

While I was at it, I switched a few recipe scripts to use the 4B instruct model instead of the 8B hybrid-thinking model -- this should improve speed and simplify some renderer issues.

*Math*

Screenshot 2025-10-24 at 3 08 54 PM

*Twenty Question*

Screenshot 2025-10-24 at 2 06 37 PM

*RLHF*

Screenshot 2025-10-24 at 2 06 34 PM

### DIFF
--- a/tinker_cookbook/utils/logtree.py
+++ b/tinker_cookbook/utils/logtree.py
@@ -29,7 +29,7 @@ from contextvars import ContextVar
 from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Callable, Iterator, Mapping, Protocol, Sequence, TypeVar
+from typing import Any, Callable, Iterator, Mapping, Protocol, Sequence, TypeVar, overload
 
 # Context variables for task-local state
 _current_trace: ContextVar["Trace | None"] = ContextVar("lt_current_trace", default=None)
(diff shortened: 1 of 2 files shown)


tinker_cookbook/utils/logtree.py
```diff
@@ -0,0 +1,1045 @@
...
+
+
+@contextmanager
+def scope_header(title: str, **attrs: Any) -> Iterator[None]:
+    """
+    Open a section with an auto-leveled header.
+
+    Args:
+        title: Text for the header
+        **attrs: HTML attributes (use class_="foo" for class, data__x="y" for data-x)
```
kennyyu, line 447: (perhaps in a future change) Would it be helpful to add a layer on direction to allow swapping out the html rendering with something else in future? Example:

- Current PR: annotated code -> produces html
- Potential future: annotated code -> produces structured data (e.g. json) -> default rollout viewer would render with html.
  - However, if folks want to change how the rollout viewer looks in the future, they can keep the code-as-is and use the json format to create a new rollout viewer.

kennyyu (approved):
This is awesome!!

tinker_cookbook/utils/logtree.py
joschu: yeah that would be great!

PR #41, opened 2025-10-26
Modify supervised/train.py to be async and use pipelining

joschu:
As explained in https://tinker-docs.thinkingmachines.ai/under-the-hood, we should get better performance by pipelining the requests, so that we send the next iteration's fwdbwd+optim requests while waiting for the current one.

I had previously implemented this, but it wasn't faster, possibly due to suboptimal scheduling. However, I just tried it again on prod, and it does seem to be faster.

image

### DIFF
--- a/tinker_cookbook/supervised/train.py
+++ b/tinker_cookbook/supervised/train.py
@@ -1,5 +1,10 @@
 """
 Supervised fine-tuning (SFT)
+
+This module implements a pipelined supervised learning training loop. For background on
+why we pipeline requests, see https://tinker-docs.thinkingmachines.ai/under-the-hood.
+For a minimal, pedagogical example of SL training without these optimizations,
+refer to `tinker_cookbook/recipes/sl_loop.py`.
 """
 
 import asyncio
@@ -186,9 +191,7 @@ async def main(config: Config):
         if data:
             logger.info(colorize_example(data[0], tokenizer))
 
-        fwd_bwd_future = await training_client.forward_backward_async(
-            data, loss_fn="cross_entropy"
-        )
+        fwd_bwd_future = await training_client.forward_backward_async(data, loss_fn="cross_entropy")
         optim_step_future = await training_client.optim_step_async(adam_params)
 
         return SubmittedBatch(


tinker_cookbook/supervised/train.py
```diff
@@ -227,30 +166,114 @@ def main(config: Config):
...
+            num_loss_tokens=sum(
+                sum(datum.loss_fn_inputs["weights"].data) for datum in submitted.data
+            ),
+            train_mean_nll=train_nll,
+        )
+        metrics["time/total"] = time.time() - submitted.batch_start_time
+
+        if evaluators and config.eval_every > 0 and submitted.step % config.eval_every == 0:
+            with timed("evals", metrics):
+                eval_metrics = await run_evals(evaluators, training_client, submitted.step)
```
kennyyu, line 243: If running evals synchronously on the main training loop is a bottleneck in the future, we could run the evals on a background coroutine -- this would allow overlapping the next step to train while we do evals on the current step. The downside here is that we won't be able to log the eval metrics in the same event as the normal training step metrics (but that might be OK from wandb perspective).

tinker_cookbook/supervised/train.py
```diff
@@ -227,30 +166,114 @@ def main(config: Config):
         f"Training for {n_batches} batches x {config.num_epochs} epochs = {n_batches * config.num_epochs} steps"
     )
 
-    # Training loop
+    async def submit_batch(epoch_idx: int, batch_idx: int) -> SubmittedBatch:
```
kennyyu, line 169: To visualize the performance before/after, we can add tracing (`@scope`) to all of these functions: https://github.com/thinking-machines-lab/tinker-cookbook/pull/28

I can do that in a followup :)

kennyyu (approved):
Awesome!

tinker_cookbook/supervised/train.py
joschu: yeah that's a good idea. wandb might already handle it correctly. the jsonl loading would need to do something slightly more complicated, but it's not too bad.

tinker_cookbook/supervised/train.py
joschu: great!

tinker_cookbook/supervised/train.py
joschu: i'd be in favor of that -- @rafailovtm needed and implemented this previously, but at the time it caused too much complexity to merge.

PR #44, opened 2025-10-27
Add DeepSeek-V3.1 and gpt-oss model-info and renderers

opherlieber:


tinker_cookbook/tests/test_renderers.py
```diff
@@ -6,12 +6,26 @@
...
+)
 def test_against_hf_chat_templates(model_name: str):
     tokenizer = AutoTokenizer.from_pretrained(model_name, use_fast=True)
     # not using get_tokenizer(model_name)
     # because we want to test against the original tokenizer from HF, not the mirror
-    cookbook_renderer = get_renderer(get_recommended_renderer_name(model_name), tokenizer)
+    # gpt_oss HF matches gpt_oss_medium_reasoning and not the default gpt_oss
+    render_name = (
+        get_recommended_renderer_name(model_name)
+        if not model_name.startswith("openai")
```
devendrachaplot, line 25: Maybe put this inside get_recommended_renderer_name?

opherlieber: Currently the one we're recommending first for gpt-oss in 'get_recommended_renderer_name' is gpt_oss_no_sysprompt (Since most users wont be training 'reasoning_effort' by default), but for this test we need to use gpt_oss_medium_reasoning to match the HF template.

PR #46, opened 2025-10-27
Fix DPO dataset builder for `helpsteer3` and `ultrafeedback`

janos-perczel:
`helpsteer3` and `ultrafeedback` currently use `ChatDatasetBuilderFromComparisons`, which yields one `tinker.Datum` (with optional A/B swap for data augmentation), which is incorrect for DPO. Depending on data and batch size this may run without crashes (e.g. it fails silently with `ultrafeedback` on Llama-3.2-1B), but will ingest unrelated accepted and rejected samples leading to no meaningful learning (margin and accuracy hover both around zero and 0.5 respectively). 

DPO needs two datums as seen in the correct implementation for `hhh` using `DPODatasetBuilderFromComparisons`. 

After fix, both `helpsteer3` and `ultrafeedback` and run as expected using the [demo code](https://tinker-docs.thinkingmachines.ai/preferences/dpo-guide) in the documentation (note the small typo in the docs `python -m tinker_cookbook.recipes.preference.train` should read `python -m tinker_cookbook.recipes.preference.dpo.train`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 **/__pycache__
-.DS_Store


rafailovtm (approved):
Goog catch!

tinker_cookbook/recipes/preference/datasets.py
```diff
@@ -174,7 +174,7 @@ def example_to_labeled_comparison(self, example: dict) -> LabeledComparison | No
             completion_B=[{"role": "assistant", "content": response2}],
         )
         return LabeledComparison(
-            comparison=comparison, label="A" if overall_preference > 0 else "B"
```
janos-perczel, line 177: Code gets the comparison backwards. In `HelpSteer3` a positive `overall_preference` indicates preference for `Response 2` (see [data info page on HF](https://huggingface.co/datasets/nvidia/HelpSteer3)), which is mapped to `completion_B` in the code above. 

The current implementation leads to the wrong response being labelled as preferred/rejected as `DPODatasetBuilderFromComparisons` handles labels as follows:

```
            chosen_completion = (
                labeled_comparison.comparison.completion_A
                if labeled_comparison.label == "A"
                else labeled_comparison.comparison.completion_B
            )
```

PR #51, opened 2025-10-28
Support analysis channel in gptoss renderer

sarahegler:
Add support for CoT in the gptoss renderer based on https://cookbook.openai.com/articles/gpt-oss/fine-tune-transfomers and https://cookbook.openai.com/articles/gpt-oss/handle-raw-cot

An example like this: 
```
{"role": "user", "content": "What is 2+2?"},
{
    "role": "assistant",
    "thinking": "The user wants me to calculate 2+2. Let me compute: 2+2 = 4.",
    "content": "The answer is 4.",
},
```

Now renders to the harmony format like this:
```
<|start|>system<|message|>You are ChatGPT, a large language model trained by OpenAI.
Knowledge cutoff: 2024-06
Current date: 2025-10-28

Reasoning: medium

# Valid channels: analysis, commentary, final. Channel must be included for every message.<|end|><|start|>user<|message|>What is 2+2?<|end|><|start|>assistant<|channel|>analysis<|message|>The user wants me to calculate 2+2. Let me compute: 2+2 = 4.<|end|><|channel|>final<|message|>The answer is 4.<|return|>
```

Left commentary channel / tooling for a separate PR.

### DIFF
--- a/tinker_cookbook/tests/test_renderers.py
+++ b/tinker_cookbook/tests/test_renderers.py
@@ -10,7 +10,7 @@ from transformers.models.auto.tokenization_auto import AutoTokenizer
     "model_name",
     [
         "meta-llama/Llama-3.2-1B-Instruct",
-        "Qwen/Qwen3-30B-A3B-Instruct-2507",
+        # "Qwen/Qwen3-30B-A3B", TODO: This was broken, will address in another PR.
         "deepseek-ai/DeepSeek-V3.1",
         "openai/gpt-oss-20b",
     ],
@@ -45,6 +45,7 @@ def test_generation_against_hf_chat_templates(model_name: str):
     elif model_name.startswith("deepseek-ai"):
         aug_convo = convo
     elif model_name.startswith("openai"):
+        # Thinking field should not be rendered in this case as it is not the last message.
         convo[1]["thinking"] = "The user is sharing a greeting. We should respond politely."
         aug_convo = convo
     else:
@@ -65,7 +66,7 @@ def test_generation_against_hf_chat_templates(model_name: str):
     "model_name",
     [
         "meta-llama/Llama-3.2-1B-Instruct",
-        "Qwen/Qwen3-30B-A3B-Instruct-2507",
+        "Qwen/Qwen3-30B-A3B",
         "deepseek-ai/DeepSeek-V3.1",
         "openai/gpt-oss-20b",
     ],
@@ -100,7 +101,7 @@ def test_supervised_example_against_hf_chat_templates(model_name: str):
     elif model_name.startswith("deepseek-ai"):
         aug_convo = convo
     elif model_name.startswith("openai"):
-        # Test thinking field for GPT-OSS
+        # Test thinking field for GPT-OSS is rendered.
         convo[1]["thinking"] = "The user is sharing a greeting. We should respond politely."
         aug_convo = convo
     else:
@@ -171,12 +172,8 @@ def test_eot_parsing(model_name: str, renderer_name: str):
 
 
 if __name__ == "__main__":
-    test_generation_against_hf_chat_templates("meta-llama/Llama-3.2-1B-Instruct")
-    test_generation_against_hf_chat_templates("Qwen/Qwen2.5-VL-3B-Instruct")
-    test_generation_against_hf_chat_templates("deepseek-ai/DeepSeek-V3.1")
+    # test_against_hf_chat_templates("meta-llama/Llama-3.2-1B-Instruct")
+    # test_against_hf_chat_templates("Qwen/Qwen2.5-VL-3B-Instruct")
     test_generation_against_hf_chat_templates("openai/gpt-oss-20b")
-    test_supervised_example_against_hf_chat_templates("meta-llama/Llama-3.2-1B-Instruct")
-    test_supervised_example_against_hf_chat_templates("Qwen/Qwen3-30B-A3B-Instruct-2507")
-    test_supervised_example_against_hf_chat_templates("deepseek-ai/DeepSeek-V3.1")
     test_supervised_example_against_hf_chat_templates("openai/gpt-oss-20b")
     test_eot_parsing("Qwen/Qwen3-30B-A3B", "qwen3")


opherlieber:
Thanks for the PR!
I think each channel should be formatted like a new assistant message, so you are missing the `<|start|>assistant` in the final channel, i.e. in your example it should be:
```
<|start|>assistant<|channel|>analysis<|message|>The user wants me to calculate 2+2. Let me compute: 2+2 = 4.<|end|><|start|>assistant<|channel|>final<|message|>The answer is 4.<|return|>
```
That's according to [this](https://cookbook.openai.com/articles/openai-harmony#example-output), and also that's what I get in the HF chat template for this example (If you can also add a version of test_against_hf_chat_templates that tests the thinking it would be great too)

Also, if you can add an assert in the other render_message functions that the new message.thinking field is None, at least if/until we support those there too (Also maybe worth asserting that it's None also in gpt-oss if it's a user message)

joschu:
According to @opherlieber's comment, the gpt-oss thinking block is in its own Message. So are you sure we want thinking as a separate attribute of the Message, as opposed to just using the content field?

That said, having thinking as a separate field of the Message could actually make sense for models like the Qwen3 ones where every message includes thinking followed by text. Currently, for these models, content field includes the <think></think> tags rendered as text, which requires parsing the string, and could lead to prompt injection issues.

joschu:
It's worth looking at other libraries like https://github.com/UKGovernmentBEIS/inspect_ai, which work with various models with different thinking conventions, to see how they deal with these issues.

opherlieber:
I think it might make sense to have the channel as a separate message in this specific gpt-oss case, but might still be preferable to have it in a single message like the current PR does since:

- It can allow the same thinking messaging format support for other models like you wrote
- It matches the format of the HF chat template (i.e. single message with 'content' and 'thinking' fields)
- The harmony spec defines to remove the analysis/thinking from all but the last assistant message, which complicates and can be confusing if it's a separate message and not part of the last one.

sarahegler:
Thanks for the review @opherlieber @joschu! I was probably oversimplifying for the gpt-oss case a bit. I based this off the gpt-oss [chat template](https://huggingface.co/openai/gpt-oss-20b/blob/main/chat_template.jinja#L268) and [fine-tuning cookbook](https://cookbook.openai.com/articles/gpt-oss/fine-tune-transfomers) where `thinking` is a separate field within the message, parallel to `content`. OpenRouter also has a reasoning_details in parallel to the content field in the message. 

It looks like Inspect used to do the same, but [switched to defining a ContentReasoning type](https://github.com/UKGovernmentBEIS/inspect_ai/pull/1390) alongside other Content blocks, e.g. ContentText, ContentReasoning, ContentToolUse, ContentImage, etc. (They changed it to support multiple fields in the reasoning content as with [Anthropic’s API](https://docs.claude.com/en/docs/build-with-claude/extended-thinking)) In that case, the reasoning is still part of the same message, but a separate content block. Happy to try something more like that as 1) I like the extensibility for things like multimodal content, 2) user messages won’t have a “thinking” field so might make more sense to not have that explicitly defined on the message (or else to define a separate subclass for assistant messages like they also do in Inspect), 3) and like you said, it might fit nicely with some of the other models where the tags are currently part of the content field.

So for now my thinking is either 1) have reasoning in a separate field like this or 2) define a separate Content block type similar to Inspect and allow the content field to be an array of such but let me know what you prefer! Happy to also include changes for other renders in this PR if it makes sense to see it across the board before approving.

sarahegler:
jinx, what @opherlieber said :)

tinker_cookbook/tests/test_renderers.py
```diff
@@ -44,12 +45,80 @@ def test_against_hf_chat_templates(model_name: str):
     elif model_name.startswith("deepseek-ai"):
         aug_convo = convo
     elif model_name.startswith("openai"):
+        convo[1] = {
```
opherlieber: Maybe just set `convo[1]["thinking"] = "..."`
Also, I think it's not actually tested here since it's not the last message so we don't render the thinking?

tinker_cookbook/tests/test_renderers.py
```diff
@@ -10,12 +10,13 @@
     "model_name",
     [
         "meta-llama/Llama-3.2-1B-Instruct",
-        "Qwen/Qwen3-30B-A3B",
+        "Qwen/Qwen3-30B-A3B-Instruct-2507",
```
opherlieber: Why was it changed?

sarahegler: This test case was failing on `main`, I believe because there isn't yet support for removing thinking spans in multi-turn convos (https://github.com/thinking-machines-lab/tinker-cookbook/blob/main/tinker_cookbook/renderers.py#L327) so there is an extra `<thinking>/n` compared to the HF version. I changed it to use the instruct version without thinking in the meantime instead of letting it fail/leaving commented out.

tinker_cookbook/tests/test_renderers.py
sarahegler: I wanted to test that it doesn't get rendered since it's not the last message. I updated to use `convo[1]["thinking"] = "..."`

tinker_cookbook/tests/test_renderers.py
sarahegler: Let me know if you prefer I address this separately. Happy to also take a look at removing the thinking spans in multi-turn qwen convos.

opherlieber: Yes, it seems that broke, we can address that in a different PR

tinker_cookbook/tests/test_renderers.py
opherlieber: Testing it's not rendered is good, but we should also test case where it is rendered.
Let's maybe add a 4th message to the convo where the assistant answers the question with thinking text, and then run the test with and without the last message i.e. on `convo[:-1]` and `convo`

tinker_cookbook/tests/test_renderers.py
sarahegler: sounds good, will leave it unchanged / commented out

sarahegler: @opherlieber  for that test, the two reasons for failure are 1) we need to remove <thinking> span from everything except the last message as mentioned, and 2) we intentionally add an extra <thinking> tag (https://github.com/thinking-machines-lab/tinker-cookbook/blob/main/tinker_cookbook/renderers.py#L335) which is not added by HF, so it seems intentional to not match the HF implementation (in which case we can modify the test to make that clear)

tinker_cookbook/tests/test_renderers.py
sarahegler: The test below with the supervised example tests with the assistant message as the last one where thinking is rendered. For this generation prompt test, I was thinking it didn't make sense to add an assistant message as the last, but are you suggesting to test with assistant prefills?

opherlieber: Right sorry, I missed the separate supervised test, looks good then!

PR #54, opened 2025-10-30
[logtree for RLTestEvaluator]

ruiqi-zhong:
- Adding logtree for evaluator
- Now the `scope_header` decorator can automatically adds `enable_logging` arguments.
- Remove redundant overloads of `scope_header` function to simplify code.

Ran the arithmetic env example and the logging is correct.

### DIFF
--- a/tinker_cookbook/utils/logtree.py
+++ b/tinker_cookbook/utils/logtree.py
@@ -477,12 +477,12 @@ def scope_header(title: str, **attrs: Any) -> Iterator[None]:
 F = TypeVar("F", bound=Callable[..., Any])
 
 
-# Overloads to support both bare decorator and parameterized usage
-# More specific overloads must come first
+# Overloads the parameterized usage
 @overload
 def scope_header_decorator(title: str) -> Callable[[F], F]: ...  # String title
 
 
+# Overloads the bare usage
 @overload
 def scope_header_decorator(title: F) -> F: ...  # Bare: @scope_header_decorator
 
@@ -495,15 +495,14 @@ def scope_header_decorator(
 
     Args:
         title: String or function returning string
-        **attrs: HTML attributes
 
     Examples:
         @logtree.scope_header_decorator
         async def process_batch():
             ...
 
-        @logtree.scope_header_decorator(lambda self, x: f"Item {x}")
-        def handle_item(self, x):
+        @logtree.scope_header_decorator("Handling item")
+        def handle_item():
             ...
     """
     title_str = title if isinstance(title, str) else title.__name__


tinker_cookbook/utils/logtree.py
```diff
@@ -483,48 +480,35 @@ def scope_header(title: str, **attrs: Any) -> Iterator[None]:
...
     Examples:
         @logtree.scope_header_decorator
         async def process_batch():
             ...
 
-        @logtree.scope_header_decorator("Processing")
-        def work():
-            ...
-
         @logtree.scope_header_decorator(lambda self, x: f"Item {x}")
```
kennyyu: Looks like the code to handle lambdas was removed -- do we need to update the comment?

tinker_cookbook/utils/logtree.py
```diff
@@ -483,48 +480,35 @@ def scope_header(title: str, **attrs: Any) -> Iterator[None]:
...
-        title: String, callable returning string, or None (use function name)
+        title: String or function returning string
         **attrs: HTML attributes
 
     Examples:
         @logtree.scope_header_decorator
         async def process_batch():
             ...
 
-        @logtree.scope_header_decorator("Processing")
```
kennyyu, line 518: From the unit tests, it looks like this still works!

kennyyu (approved):

tinker_cookbook/utils/logtree.py
ruiqi-zhong: nice catch! will fix

tinker_cookbook/utils/logtree.py
ruiqi-zhong: yeah, I only deleted the usages like `logtree.scope_header_decorator(lambda x: "Processing")` , which we don't really need right now and adds code complexity.

tinker_cookbook/rl/metric_util.py
```diff
@@ -102,15 +103,29 @@ def dataset_to_env_group_builders(dataset: RLDataset) -> list[EnvGroupBuilder]:
...
 
 class RLTestSetEvaluator(SamplingClientEvaluator):
-    def __init__(self, dataset: RLDataset, max_tokens: int, name: str | None = None):
+    def __init__(
+        self,
+        dataset: RLDataset,
+        max_tokens: int,
+        name: str | None = None,
+        num_groups_to_log: int = 4,
+        log_path: str | None = None,
```
joschu, line 112: is this argument used?

tinker_cookbook/rl/train.py
```diff
@@ -41,10 +40,24 @@
...
+from contextlib import contextmanager
+
 
 logger = logging.getLogger(__name__)
 
 
+@contextmanager
+def get_logtree_scope(
+    log_path: str | None, num_groups_to_log: int, f_name: str, scope_name: str
+) -> Iterator[None]:
```
joschu, line 52: could you add a docstring please?

joschu: please note that it creates the file specified, and explain what the arguments do

PR #58, opened 2025-10-30
Fix Qwen rendering and parsing for tool calls

eddie-scio:
1. Rendering: `Message` can have `tool_calls` which are currently ignored -- this renders them in with how Qwen docs refer to them.
2. Parsing: `<tool_call>` seems more widely used in official Qwen docs than `<function_call>`, so switched the `parse_response` implementation to use that.

Looked at the unit tests to update but they seem commented out for Qwen models, so didn't touch them

### DIFF
--- a/tinker_cookbook/renderers.py
+++ b/tinker_cookbook/renderers.py
@@ -336,21 +336,20 @@ class Qwen3Renderer(Renderer):
             # <think> in the assistant messages, we so don't need to re-add it in those cases.
             ob_str += "<think>\n"
         # Observation (prompt) part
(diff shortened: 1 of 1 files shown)


eddie-scio:
Hi @Tiiiger -- are you the right person to review this contribution?  Or maybe @kzl-tml

Tiiiger (approved):
Thanks! Makes  sense. Original I used the <functional_call> tag because Qwen2.5 models were not trained with <tool_call> but Qwen3 models are so ... this is good!

PR #59, opened 2025-10-30
fix resume checkpoint conflict in frontend

Xiuyu-Li:
This PR fixes a checkpoint conflict (409 error) that can occur when resuming training from an existing checkpoint. For example:
```bash
...
  File "/rscratch/xiuyu/anaconda3/envs/tinker/lib/python3.12/site-packages/tinker/_base_client.py", line 1578, in request
    raise self._make_status_error_from_response(err.response) from None
tinker.ConflictError: Error code: 409 - {'detail': "Checkpoint '000060' already exists for model 951d7759-2db7-4f44-9f36-c6d090bad0a4 in weights. Please choose a different name to avoid overwriting."}
```
The issue likely originates from a backend caveat. This PR proposes a frontend fix that resolves it by skipping the initial save when resuming—matching the behavior of training from scratch.

Specifically, it changes
```python
kind="both" if (i_batch > 0 and i_batch % save_every == 0) else "sampler"
```
to
```python
kind="both" if (i_batch > start_batch and i_batch % save_every == 0) else "sampler"
```
in `save_checkpoint_and_get_sampling_client`, ensuring consistent behavior across resumed and new runs.

### DIFF
--- a/tinker_cookbook/rl/train.py
+++ b/tinker_cookbook/rl/train.py
@@ -257,7 +257,7 @@ async def do_sync_training_with_stream_minibatch(
     """
     # Initial sampling client
     sampling_client, _ = await save_checkpoint_and_get_sampling_client(
-        training_client, start_batch, cfg.log_path, cfg.save_every
+        training_client, start_batch, cfg.log_path, cfg.save_every, start_batch
     )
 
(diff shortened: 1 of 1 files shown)


Tiiiger (approved):
LGTM thanks!

PR #60, opened 2025-10-31
parallelize evaluation in rl-training

ruiqi-zhong:
Parallelize different evaluators during a RL training loop.

### DIFF
--- a/tinker_cookbook/rl/train.py
+++ b/tinker_cookbook/rl/train.py
@@ -288,7 +288,7 @@ async def run_evaluations_parallel(
         ev_name = _get_evaluator_name(evaluator)
         task = asyncio.create_task(
             run_single_evaluation(evaluator, cfg, i_batch, sampling_client),
-            name=f"eval_task_{ev_name or i}_{i_batch}",
+            name=f"eval_{ev_name or i}_iteration_{i_batch:06d}",
         )
         tasks.append(task)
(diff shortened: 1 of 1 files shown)


tinker_cookbook/rl/train.py
```diff
@@ -254,6 +254,46 @@ class Config:
...
+            num_groups_to_log=cfg.num_groups_to_log,
+            f_name=f"eval_{ev_name}_iteration_{i_batch:06d}",
+            scope_name=f"Running evaluation {ev_name} {i_batch}",
+        ):
+            eval_metrics = await evaluator(sampling_client)
+            return {f"test/{k}": v for k, v in eval_metrics.items()}
+
+    # Create tasks for all evaluators
+    tasks = [
+        run_single_evaluation(evaluator, cfg, i_batch, sampling_client) for evaluator in evaluators
```
kennyyu: nit: can we use `asyncio.create_task(..., name=f"evals_task_{ev_name}")` to give these coroutines a name? This makes it easier to identify these workloads in traces!

Example: https://github.com/thinking-machines-lab/tinker-cookbook/blob/main/tinker_cookbook/rl/train.py#L347-L350

ruiqi-zhong: good catch! just addressed.

PR #64, opened 2025-10-31
add support for model_path to verifiers evaluate script

tmabraham:
right now the verifiers evaluate script doesn't take in the `model_path` argument so you cannot evaluate the model checkpoints. Adding support similar to how it's in the [run_inspect_evals.py](https://github.com/thinking-machines-lab/tinker-cookbook/blob/main/tinker_cookbook/eval/run_inspect_evals.py) script.

cc: @willccbb


joschu:
@tmabraham sorry for the delay -- reviewing now

joschu (approved):

PR #72, opened 2025-11-03
remove redundant variables defined in search_env.py

mzio:
Remove hardcoded hf_repo_id and parquet_filename variable which were already specified

Was tracing thru this file for reference impl of multi-turn tool call; minor nit but you defined some vars twice 

- Also maybe want to generalize this with a config or chz thing,


kennyyu (approved):
Thank you!

PR #79, opened 2025-11-10
make deepseekv3 renderer work with system messages, add renderer that forces thinking

joschu:


tinker_cookbook/renderers.py
```diff
@@ -564,6 +590,31 @@ def build_generation_prompt(
         return super().build_generation_prompt(messages, role, prefill)
 
 
+class DeepSeekV3ForceThinkingRenderer(DeepSeekV3Renderer):
+    """
+    Renderer that forces inclusion of a thinking block for DsV3 models.
+    """
+
+    def _render_message(self, message: Message) -> tuple[list[int], list[int], list[int]]:
```
joschu, line 598: I copied this from the NoThinking renderer. Not sure exactly why it's here.

tinker_cookbook/renderers.py
```diff
@@ -564,6 +590,31 @@ def build_generation_prompt(
...
+        if message["role"] == "assistant":
+            content = message["content"]
+            new_content = content
+            if new_content.startswith("</think>"):
+                new_content = new_content[len("</think>") :]
+            if not new_content.startswith("<think>"):
+                new_content = "<think>" + new_content
```
opherlieber, line 605: Do we want to support/use the Message's 'thinking' field here like gpt-oss? If we do want to support `<think>` also directly in the content, should we enforce here that there is also a `</think>` followed by an actual message after it? And not sure if for multi-message interaction the thinking tokens should be rendered in all previous assistant messages, or be kept only in the last message like in gpt-oss?

joschu: Good questions. Haven't thought these things through yet. No rush to merge this one -- still experimenting with it in my use case.

joschu: Re: thinking field, we might as well prioritize supporting the most advanced tool use systems that are being released, which would include gpt-oss and kimi-k2-thinking, which use interleaved CoT + tool calls. So I'm in favor of whatever improves the support for those models.

joschu: I'm not sure what's the right policy for possibly hiding CoT -- in gpt-oss, I assume we include all the thinking traces from the current turn (i.e., back until the last user message), in cases where we have multiple assistant messages including tool calls?

opherlieber: Yes, in gpt-oss we include thinking tokens from the current turn, even if there are multiple assistant channel messages

PR #84, opened 2025-11-11
Use save_weights_and_get_sampling_client_async in rl/train.py

andriigrynenko:
This enables use of ephemeral checkpoints which eventually can be made more performant.


andriigrynenko:
Note, this is blocked on SDK update to stop requiring `name` for ephemeral sampling clients.

kennyyu (approved):

PR #85, opened 2025-11-11
[Train-on-what; human simulator]

ruiqi-zhong:
Adding options to train on user-specified messages.

Ran 

```bash
python -m tinker_cookbook.recipes.chat_sl.train
    model_name=Qwen/Qwen3-8B-Base \
    dataset=no_robots \
    learning_rate=5e-4 \
    batch_size=64 \
    lora_rank=64 \
    eval_every=20 \
    save_every=20 \
    wandb_project=cookbook_sl
```

and I can reproduce the performance.


joschu:
In particular, try both the role_colon renderer (nontrivial ac_tail) and another special token renderer.

ruiqi-zhong:
> viz_sft_dataset

Thanks for checking! yeah it works. 

image

PR #88, opened 2025-11-12
[tinker-cookbook] supervised: add tracing annotations

kennyyu:
This adds tracing annotations to make it easier to identify where time is spent in SFT workloads. To run it:

- set `enable_trace=True`
- Run sl_basic:
```
python -m tinker_cookbook.recipes.sl_basic enable_trace=true
```
- While the job is running, generate the trace:
```
python -m tinker_cookbook.utils.trace /tmp/tinker-examples/sl_basic/trace_events.jsonl /tmp/sl_basic.json
```
- And then visualize with perfetto:

Screenshot 2025-11-21 at 7 44 13 AM

### DIFF
--- a/tinker_cookbook/supervised/train.py
+++ b/tinker_cookbook/supervised/train.py
@@ -16,6 +16,7 @@ from dataclasses import dataclass
 import chz
 import tinker
 from tinker.lib.public_interfaces import APIFuture
+
 from tinker_cookbook import checkpoint_utils
 from tinker_cookbook.display import colorize_example
 from tinker_cookbook.eval.evaluators import (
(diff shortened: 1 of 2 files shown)


tinker_cookbook/supervised/train.py
```diff
@@ -166,8 +196,12 @@ async def main(config: Config):
         f"Training for {n_batches} batches x {config.num_epochs} epochs = {n_batches * config.num_epochs} steps"
     )
 
+    @scope
     async def submit_batch(epoch_idx: int, batch_idx: int) -> SubmittedBatch:
         step = epoch_idx * n_batches + batch_idx
+        context = get_scope_context()
+        context.attributes["step"] = step
```
joschu, line 229: nit: could be nice to have a helper to do
`context.set_attribute('step', step)`
or even
`set_scope_attribute('step', step)`
which does the get_scope_context() internally

joschu (approved):

kennyyu:
Pre-existing type check/uv errors

PR #90, opened 2025-11-12
supervised/train.py: shift evals so they occur after 0, k, 2k, ... steps instead of 1, k+1, 2k+1, etc where k=config.eval_every

joschu:


tinker_cookbook/tests/test_utils.py
```diff
@@ -42,5 +42,6 @@ def log_metrics(metrics: dict[str, Any], step: int):
 
     mock_logger.log_metrics = log_metrics
     mock_logger.close = MagicMock()
+    mock_logger.get_logger_url = MagicMock(return_value=None)
```
joschu, line 45: needed to make the resumption test pass

kennyyu (approved):

PR #95, opened 2025-11-14
Adjust sample command line

gklab:
The sample Python command was missing a slash, causing the following arguments to be ignored, so the training used the default parameters in the code.


kennyyu (approved):
Thank you!

PR #99, opened 2025-11-16
Fix advantage computation + add advantage to metric logs

sdan:
When a trajectory group is 1, centering advantage within that 1 group is: reward - reward = 0.

Fix is to instead use batch mean as baseline when our group size is 1

Also added advantage stats logging (mean/std/min/max)


Tiiiger:
hi @sdan thank you for your contribution! but I don't think this is a common/standard enough technique to be merged into recipe. please fork if needed!

PR #101, opened 2025-11-18
qwen_instruct and search_env recipe fixes

eddie-scio:
This is the same idea as https://github.com/thinking-machines-lab/tinker-cookbook/pull/58/files but for `Qwen3InstructRenderer`, which overrides `_render_message`.  Also, the `search_env` recipe has `Qwen/Qwen3-4B-Instruct-2507` as default, so changing the tool description / sys prompt there to s/function_call/tool_call as well


eddie-scio:
cc @Tiiiger

PR #103, opened 2025-11-20
fix: on-policy distillation missing temperature parameter fix

Andrewzh112:
The do_group_rollout_and_filter_constant_reward sampling step requires a temperature parameter, right now it is not specified in `distillation/train_on_policy.py` so the runs will fail. Added the `temperature` parameter to config and specifying it in do_group_rollout_and_filter_constant_reward for `distillation/train_on_policy.py` and the recipes fixes the problem.

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -1,0 +1,82 @@
+name: Claude Code
+
+permissions:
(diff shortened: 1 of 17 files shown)


joschu:
@claude please merge in main then review

joschu (approved):

PR #104, opened 2025-11-20
[adding name to NLL evaluator]

ruiqi-zhong:
adding name to NLL evaluator. otherwise we cannot have two different nll evaluators and log their results properly.


tinker_cookbook/supervised/nll_evaluator.py
```diff
@@ -7,7 +7,8 @@
 
 
 class NLLEvaluator(TrainingClientEvaluator):
-    def __init__(self, data: list[tinker.Datum]):
```
joschu, line 10: actually wouldn't it be better to make the name default to "test", and remove the prefixing of "test" from the training loop?

ruiqi-zhong: yeah that's better -- I thought the original code meant to always want "test/" to be the prefix of the log name but that's indeed a bit cumbersome.

PR #106, opened 2025-11-21
[tinker-cookbook] set_scope_context: add helper

kennyyu:
This adds a helper to make it easier to set context when tracing, and changes the existing callsites to use the new helper.


joschu:
Renamed on claude's recommendation.

PR #107, opened 2025-11-22
[tinker-cookbook] rl: avoid hanging in async runs when we run out of data

kennyyu:
Previously, on async RL runs, we can hang in shutdown if we run out of data. This fixes it to ensure proper shutdown and that all data in queues are drained when the dataloader loop terminates:

Overview of the shutdown workflow:
1. dataloader terminates, enqueues sentinel values in the trajectory worker queues
2. trajectory worker sees sentinel values and terminates
3. last trajectory worker enqueues a sentinel value to terminate the training loop
4. training loop sees a sentinel value and terminates, and signals the evals loop
5. evals loop is signaled and terminates

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -14,6 +14,7 @@ on:
 jobs:
   claude_review:
     runs-on: ubuntu-latest
+    environment: claude-review
     steps:
       - name: Run Claude Code review
         uses: anthropics/claude-code-action@v1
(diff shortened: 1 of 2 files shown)


joschu:
@kennyyu I think there was a bug, which was caught by claude and codex reviews. I think I fixed it in my last push. Now trying to fix the auto review functionality.

YujiaBao:
https://github.com/thinking-machines-lab/tinker-cookbook/pull/480

PR #123, opened 2025-11-23
Fix TextArena double-wrapper error in TicTacToe example

joschu:
## Summary
Fixes the `ValueError: Environment is already wrapped with LLMObservationWrapper. Double-wrapping is not allowed.` error that prevents the TicTacToe example from running.

## Problem
The multiplayer RL TicTacToe example was failing with the following error:
```
ValueError: Environment is already wrapped with LLMObservationWrapper. Double-wrapping is not allowed.
```

This was reported in thinking-machines-lab/tinker-feedback#30.

## Root Cause
The issue is caused by an upstream change in **textarena 0.7.4**:

1. **Automatic wrapper application**: `ta.make(env_id="TicTacToe-v0")` now automatically applies default wrappers defined during environment registration. For TicTacToe-v0, this includes `LLMObservationWrapper` and `ActionFormattingWrapper`.

2. **Double-wrapping protection**: Textarena added validation to prevent the same wrapper type from being applied twice to an environment.

3. **The conflict**: The tinker-cookbook code in `env.py:199` was manually wrapping environments with `LLMObservationWrapper` after `ta.make()` had already applied it automatically, causing the double-wrapping error.

## Solution
Removed the redundant manual wrapper application on line 199 of `tinker_cookbook/recipes/multiplayer_rl/text_arena/env.py`:

```python
# Before:
shared_env = ta.make(env_id=self.game_name)
shared_env = ta.wrappers.LLMObservationWrapper(shared_env)  # ❌ Double-wrapping
shared_env.reset(num_players=self.num_players)

# After:
shared_env = ta.make(env_id=self.game_name)  # ✅ Already has LLMObservationWrapper
shared_env.reset(num_players=self.num_players)
```

The environment from `ta.make()` already has the correct wrappers applied via textarena's automatic mechanism, so manual wrapping is no longer needed.

## Testing
- ✅ Verified this is the only location in the codebase with this issue (no other multiplayer RL examples use textarena)
- ✅ Tested by @joschu - the example now runs successfully

## Related Issue
Fixes thinking-machines-lab/tinker-feedback#30

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ dependencies = [
     "anyio",
     "pylatexenc",
     "sympy",
-    "textarena",
+    "textarena>=0.7.4",
     "math-verify",
     "scipy",
 ]


joschu:
@claude review please

joschu:
@claude could you update the textarena dependency to require textarena>=0.7.4?

joschu:
Thanks @murphyk for reporting and finding the fix!

PR #126, opened 2025-11-24
Add some more options to play_w_env.py

joschu:
Add a multiline option to play_w_env.
Also add a show_observation parameter, so you can disable the printout. (I found this useful in cases where I had another wrapper doing printouts, and I wanted to get rid of the redundancy.)


joschu:
@claude please make the following changes (based on your review):
1. implement your suggestion re whitespace
2. keep show_observation=True as default
3. add a comment, don't bother changing
4. no need to document show_observation, this only there for an obscure reason. include the doc for multitline though

PR #139, opened 2025-12-02
[allowing system role for deep seek renderer]

ruiqi-zhong:
Allowing an option for the deepseek renderer to convert system message to user messages.

### DIFF
--- a/tinker_cookbook/renderers.py
+++ b/tinker_cookbook/renderers.py
@@ -627,6 +627,7 @@ class DeepSeekV3Renderer(Renderer):
     Format like this (no newlines between messages):
         <|begin_of_sentence|><|User|>What can you help me with?<|Assistant|><think>Thinking...</think>I can help you with...<|end_of_centence|>
     For no-think, just use <|Assistant|></think>
+    Deepseek renderer does not support the system role out of the box. You can set system_role_as_user to True to automatically convert the system role to the user role.
     """
 
     def __init__(self, tokenizer: Tokenizer, system_role_as_user: bool = False):


tinker_cookbook/renderers.py
```diff
@@ -629,9 +629,13 @@ class DeepSeekV3Renderer(Renderer):
...
     """
 
+    def __init__(self, tokenizer: Tokenizer, system_role_as_user: bool = False):
+        super().__init__(tokenizer)
+        self.system_role_as_user = system_role_as_user
+
     def _render_message(self, message: Message) -> tuple[list[int], list[int], list[int]]:
         assert message.get("thinking") is None, "TODO: support CoT in DsV3 renderer"
-        if message["role"] == "user":
+        if message["role"] == "user" or (self.system_role_as_user and message["role"] == "system"):
```
kennyyu, line 639: when would we ever want `self.system_role_as_user == False`? It looks like without this change, we would hit an exception on line 643. So perhaps it's safe to make the True path the default behavior?

kennyyu (approved):

tinker_cookbook/renderers.py
ruiqi-zhong: I am not sure actually. but i am generally inclined not to change the system's default behavior. Just in case somebody might not like this conversion

I guess the downside is just the user sees an error and turns on a flag, which is not that bad. 

don't have strong opinion here though!

opherlieber: I think it makes more sense to keep the old behavior, to make sure the user is aware there is no actual system-role support if they try setting it, and require explicitly setting this flag if they want it converted

PR #142, opened 2025-12-02
feat(qwen3): Add strip_thinking_from_history option to Qwen3Renderer

joschu:
## Summary

Add a `strip_thinking_from_history` parameter to `Qwen3Renderer.__init__`:

- When `True` (default): strips `<think>...</think>` blocks from assistant messages in multi-turn history
- When `False`: preserves thinking blocks in history (useful for certain RL scenarios)

## Background

**Why this is correct:** Qwen3 models were trained this way - they only see their own thinking during the current turn, not from previous turns in multi-turn conversations.

**Why you might disable it:** Stripping leads to higher training cost for multi-turn RL because turns can't be merged into one sequence. Set `strip_thinking_from_history=False` to preserve thinking in history if you need full context or want to reduce rendering overhead.

The default behavior is unchanged (stripping is enabled).

This change is the companion to [monorepo PR #20802](https://github.com/iterationlab/monorepo/pull/20802) which adds the same option to the TML (Rust) renderer.

## Changes

- Added `strip_thinking_from_history: bool = True` parameter to `Qwen3Renderer.__init__`
- Conditioned the thinking stripping logic on this flag
- Added docstring explaining the parameter and tradeoffs


joschu:
@claude please the revised code

joschu:
@claude please review revised code

PR #149, opened 2025-12-04
Allow training runs with saving disabled via `save_every=0`

maitchison:
# Motivation

During testing it can sometimes be useful to disable checkpoint saving. A natural way to do this is to specify `save_every=0`. 

Currently doing this crashes with a divide by zero error. 

This PR modifies the behaviour of `save_every` to instead disable saving, as is already the case for `eval_every`.

### DIFF
--- a/tinker_cookbook/preference/train_dpo.py
+++ b/tinker_cookbook/preference/train_dpo.py
@@ -48,7 +48,7 @@ class Config:
     num_replicas: int = 8
     base_url: str | None = None
 
-    # Checkpointing and evaluation
+    # Checkpointing and evaluation (0 = disabled for *_every fields)
     evaluator_builders: list[EvaluatorBuilder] = chz.field(default_factory=list)
     infrequent_evaluator_builders: list[EvaluatorBuilder] = chz.field(default_factory=list)
     save_every: int = 20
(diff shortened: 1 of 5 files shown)


joschu:
@claude /review please

joschu (approved):
## ✅ Approved

### Summary
This PR correctly fixes a divide-by-zero error when `save_every=0` by adding a guard check (`save_every > 0`) before the modulo operation. The fix is minimal, surgical, and consistent with the existing `eval_every` behavior.

### Changes Reviewed
- **5 files modified** with identical pattern: `if save_every > 0 and ... % save_every == 0`
- Files: `supervised/train.py`, `rl/train.py`, `preference/train_dpo.py`, `recipes/sl_loop.py`, `recipes/rl_loop.py`

### Findings
- **Logic is correct**: The `> 0` check short-circuits before the modulo, preventing the divide-by-zero
- **Consistent with existing patterns**: `eval_every` already uses this same `> 0` guard pattern throughout the codebase
- **Condition ordering**: Minor variation in `rl/train.py` (`i_batch > start_batch` vs `step > 0`) is pre-existing and intentional for proper resume handling

### Documentation Added
I've pushed a follow-up commit adding inline documentation to clarify the `0 = disabled` behavior for `save_every` and related `*_every` fields in all affected Config classes.

LGTM! 🚀

PR #152, opened 2025-12-06
create a changelog

joschu:


CHANGELOG.md
```diff
@@ -0,0 +1,95 @@
...
+Each entry includes:
+- **Title**: A short, human-readable summary (not the commit message)
+- **Date**: When it was merged
+- **Type**: `new` (feature), `improvement` (enhancement to existing functionality), or `fix`
+- **Tags**: What area it touches (e.g., `renderers`, `rl`, `supervised`, `eval`, `datasets`)
+- **PR**: Link to the pull request
+
+---
+
+### Qwen3 thinking blocks can now be preserved in history ([#142](https://github.com/thinking-machines-lab/tinker-cookbook/pull/142))
```
danobi, line 16: i've done it before like this and it works, but confuses a bunch of people. putting a link to the PR makes it difficult to put up a PR with the change and the changelog entry. adds friction to the workflow. 

alternatively, we could have a CI job that comes by for each landed commit and creates a changelog entry PR then tags the author - claude should be able to do that well.

danobi (approved):

CHANGELOG.md
danobi: it should be quite doable - ask claude to make a github actions job to do this

joschu: Ah good point. I think it's OK -- these entries were all auto-generated by claude, so I think we can have a bot do this.

Issue #165, opened 2025-12-13
flawed tinker code for pig latin


samearle-VCConnector:
[...]

PR #169, opened 2025-12-14
fix: correct pig latin training examples for proper consonant cluster…

nathon-lee:
fix: #165 
correct pig latin training examples for proper consonant cluster handling

- Fix "split" → "it-splay" (move entire "spl" consonant cluster)
- Fix "quantum" → "antum-quay" (treat "qu" as single phonetic unit)
- Resolves inconsistent phonological rules in training data
- Addresses Issue #165: flawed tinker code for pig latin

This ensures the model learns consistent linguistic patterns instead of
contradictory character-level rules that lead to training confusion.

### DIFF
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -765,11 +765,11 @@ Let's create some training examples and convert them to a format expected by Tin
 examples = [
     {
         "input": "banana split",
-        "output": "anana-bay plit-say"
+        "output": "anana-bay it-splay"
     },
     {
         "input": "quantum physics",
-        "output": "uantum-qay ysics-phay"
+        "output": "antum-quay ysics-phay"
     },
     {
         "input": "donut shop",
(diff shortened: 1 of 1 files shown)


samearle-VCConnector:
Thanks @joschu - I'd love to have a brief conversation via phone or video. I'm east coast. Lmk how to connect, and when suits you? We're spinning up a company very similar to TML. Happy holidays!!!

samearle-VCConnector:
@leejianwoo-collab 

The Pig Latin examples in the Tinker cookbook are a perfect demonstration of why RL is currently hitting a wall.

The problem isn't the data; it's the methodology. By treating the transform of "split" as an ontological structure (a noun) rather than an etiological progression (a verb), TML is attempting to solve a problem of reality using only locality.

Regarding the 2022 Nobel Prize—the observer defines the attribute. RL can't "learn" a objective etiological path because it is designed for observation, not provenance.

Our genomic database tools already solve this—handling the same polymorphic "split" issues that are currently insoluble in Tinker.

Do you have some time after the holiday? I’d like to show you why "learning" will never solve the "banana split" problem, and how our Computational Semantics engine already has.

PR #171, opened 2025-12-14
fix: fix grammar in Search-R1 README

### LINKED ISSUE #157: Typo in the docs
Not a real big issue, but it should say "we also released" as opposed to "we also release"

Image

nathon-lee:
Change "also releases" to "also released" to use correct past tense
when describing already published research paper and data.

Fixes  #157


Tiiiger (approved):
LGTM

PR #192, opened 2025-12-19
Add checkpoint_name_prefix to RL training Config

pourion:
## Summary
Adds an optional `checkpoint_name_prefix` field to the RL training Config that prefixes all saved checkpoint names.

## Motivation
When running multiple experiments, checkpoints saved to Tinker are difficult to identify because they're named only by batch number (e.g., `000042`). This change allows prefixing with an experiment identifier (e.g., `my_exp_dec19_000042`), making it easy to match checkpoints to WandB runs.

## Changes
- Added `checkpoint_name_prefix: str | None = None` to `Config` class
- Updated `save_checkpoint_and_get_sampling_client` to accept and use the prefix
- Updated all checkpoint save call sites to pass the prefix

## Usage
Config(
    checkpoint_name_prefix="exp_rl_dec19",
    wandb_name="exp_rl_dec19",
    ...
)

## Backward Compatibility
Fully backward compatible - the field defaults to None, preserving existing behavior.

### DIFF
--- a/tinker_cookbook/rl/train.py
+++ b/tinker_cookbook/rl/train.py
@@ -256,6 +256,7 @@ class Config:
     eval_every: int = 20  # 0 = disabled
     save_every: int = 20  # 0 = disabled
     load_checkpoint_path: str | None = None
+    checkpoint_name_prefix: str | None = None
 
     async_config: AsyncConfig | None = None
     stream_minibatch_config: StreamMinibatchConfig | None = None
@@ -328,7 +329,12 @@ async def do_sync_training_with_stream_minibatch(
     """
     # Initial sampling client
     sampling_client, _ = await save_checkpoint_and_get_sampling_client(
-        training_client, start_batch, cfg.log_path, cfg.save_every, start_batch
+        training_client,
+        start_batch,
+        cfg.log_path,
+        cfg.save_every,
+        start_batch,
+        cfg.checkpoint_name_prefix,
     )
 
     for i_batch in range(start_batch, end_batch):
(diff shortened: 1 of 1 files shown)


YujiaBao:
Thanks for the PR! I appreciate the motivation — identifying checkpoints across experiments is a real pain point.

However, after looking at how other training frameworks handle this, I think checkpoint name prefixing isn't the right approach for us:

- **HuggingFace Transformers, TRL, and veRL all use the output directory as the experiment identifier**, not the checkpoint name. Checkpoints are always just `checkpoint-{step}` or `global_step_{N}`, and users distinguish experiments by writing to different output directories. This is a well-established convention.
- **Tinker already scopes checkpoints per training run** — paths are `tinker://<model_id>/<name>`, so different runs are already in separate namespaces.
- **`log_path` already serves this role locally** — all artifacts (metrics, transcripts, `checkpoints.jsonl`) live there, and users already set distinct `log_path` values per experiment.

There are also a couple of implementation concerns:
- This only covers RL training — SL, DPO, and distillation all use the same `f"{step:06d}"` naming without a prefix, which would create inconsistency.
- The `"final"` checkpoint isn't prefixed, so the feature is incomplete even within RL.

I'd suggest using distinct `log_path` values per experiment to keep things organized, which aligns with how the rest of the ecosystem works. If checkpoint browsing on the Tinker service side is the core issue, that might be better addressed at the service/CLI level.

Going to close this for now, but happy to discuss if you see a use case that the above doesn't cover!

PR #200, opened 2025-12-20
Add learning rate config support for Moonshot Kimi models

bledden:
## Summary

The `get_lr()` function in `hyperparam_utils.py` was missing support for Moonshot Kimi models, causing it to fail with an `AssertionError` when users tried to get learning rate recommendations for `moonshotai/Kimi-K2-Thinking`.

This PR adds the missing configuration:
- Added hidden size lookup for Kimi-K2 models (hidden_size=7168)
- Added learning rate scaling exponent for Kimi-K2 (0.0775, matching other MoE architectures)

Uses specific model name matching (`moonshotai/Kimi-K2` and `kimi-k2`) to avoid conflicts if Moonshot releases other model families in the future.

## Test plan

- [x] Verified `get_lr("moonshotai/Kimi-K2-Thinking")` returns a valid learning rate instead of raising an error
- [x] Confirmed existing model support (llama, qwen) is unaffected

## Note on exponent value

The exponent value (0.0775) is based on Qwen's MoE models since Kimi-K2 is also an MoE architecture. The existing exponents for Llama (0.781) and Qwen (0.0775) were empirically derived through hyperparameter sweeps. A follow-up eval run would help validate this is optimal for Kimi-K2 specifically.

Related to thinking-machines-lab/tinker-feedback#49

### DIFF
--- a/tinker_cookbook/hyperparam_utils.py
+++ b/tinker_cookbook/hyperparam_utils.py
@@ -86,6 +86,9 @@ def _get_hidden_size(model_name: str) -> int:
             "meta-llama/Llama-3.3-70B-Instruct": 8192,
         }[model_name]
 
+    if model_name == "moonshotai/Kimi-K2-Thinking":
+        return 7168
+
     config = AutoConfig.from_pretrained(model_name)
     return config.hidden_size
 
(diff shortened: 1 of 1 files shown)


tinker_cookbook/hyperparam_utils.py
```diff
@@ -86,6 +86,10 @@ def _get_hidden_size(model_name: str) -> int:
             "meta-llama/Llama-3.3-70B-Instruct": 8192,
         }[model_name]
 
+    if "moonshotai" in model_name:
```
joschu: can you make it check for the full name, in case we add another moonshot model later?

bledden: @joschu ah nice catch. Updated this and the [DeepSeek PR](https://github.com/thinking-machines-lab/tinker-cookbook/pull/199/files)

Issue #74, opened 2025-11-05
qwen3 tool rendering


csmith23:
I've been using tinker to finetune qwen3 30b a3b instruct on conversations involving tool use, and just wanted to bring up a discrepancy in the qwen3 renderer in the cookbook vs. the jinja templates on huggingface.

In the [jinja template](https://huggingface.co/Qwen/Qwen3-30B-A3B-Instruct-2507?chat_template=default), there are the lines
```
{%- elif message.role == "tool" %}
        {%- if loop.first or (messages[loop.index0 - 1].role != "tool") %}
            {{- '<|im_start|>user' }}
        {%- endif %}
```
which render tool responses with the `user` role, while in the cookbook the role is used as-is,
```
ob_str = f"{maybe_newline}<|im_start|>{message['role']}\n"
```
in `renderers.py`, lines 332 and 445.

The rendering of elements under the `"tools"` key of a conversation as part of the system prompt is also not implemented, but that seems to be the case for all renderers, not just the qwen ones

justinchiu:
what did you end up doing here? did you hack it (both role and system tools) in yourself or just apply a chat template => use the rendered system prompt

csmith23:
I just made a copy of the qwen3 renderer in the cookbook with these changes

joschu:
Sorry for the delay here -- working on a fix

joschu:
> The rendering of elements under the "tools" key of a conversation as part of the system prompt is also not implemented, but that seems to be the case for all renderers, not just the qwen ones

This is now implemented by the method `create_conversation_prefix_with_tools` that's now part of the renderer base class.

The role issue should also be fixed, and there's a unit test checking equivalence with the HF chat template. Le me know if you still see discrepancies.

PR #219, opened 2025-12-24
Fix tool calling bugs

joschu:
## Human-Written Overview

There were a few PRs pointing out issues with how tool calls were rendered and parsed. I asked claude to fix them, with reference to the Qwen3 and Kimi K2 documentation. I've reviewed the code changes and they look reasonable, but they could use another pair of eyes.

I think the tool use requires a bit more of a refactor to make sure we can render the tool specs in a model-agnostic way, and I'd like to write some tests that actually query the service with a tool-using prompt and check that we get valid tool calls, but that'll be saved for a future PR.

## Summary

This PR fixes several tool calling bugs identified in GitHub issues:

### Bug Fixes

- **Qwen3 tool response role (Issue #74)**: Tool messages with `role="tool"` now correctly render as `<|im_start|>user` with content wrapped in `<tool_response>` tags, matching HuggingFace's chat template behavior
- **Qwen3 multiple tool calls**: `parse_response` now finds all `<tool_call>` blocks instead of just the first one
- **Qwen3 content stripping**: Tool call blocks are now stripped from message content after parsing
- **Kimi K2 tool name extraction**: Function name is now correctly extracted from the `functions.{name}:{idx}` format

### Renderers Updated

- `Qwen3Renderer`
- `Qwen3InstructRenderer`
- `Qwen3VLRenderer`
- `Qwen3VLInstructRenderer`
- `KimiK2Renderer`

## Test plan

- [ ] Run existing renderer tests to ensure no regressions
- [ ] Test with search_env recipe to verify tool use still works

### DIFF
--- a/tinker_cookbook/tests/test_tool_calling.py
+++ b/tinker_cookbook/tests/test_tool_calling.py
@@ -174,18 +174,3 @@ def test_qwen3_parse_invalid_tool_call_json():
     message, success = renderer.parse_response(response_tokens)
 
     assert success is False
-
-
-def test_qwen3_parse_no_tool_call():
-    """Test parsing response without tool calls."""
-    model_name = "Qwen/Qwen3-8B"
-    tokenizer = get_tokenizer(model_name)
-    renderer = get_renderer("qwen3", tokenizer)
-
-    response_text = "I don't need to use any tools for this.<|im_end|>"
-    response_tokens = tokenizer.encode(response_text, add_special_tokens=False)
-    message, success = renderer.parse_response(response_tokens)
-
-    assert success is True
-    assert "tool_calls" not in message
-    assert message["content"] == "I don't need to use any tools for this."
--- a/tinker_cookbook/tests/test_tool_calling.py
+++ b/tinker_cookbook/tests/test_tool_calling.py
@@ -9,6 +9,7 @@ These tests verify:
 """
 
 import pytest
+import tinker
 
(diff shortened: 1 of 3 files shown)


tinker_cookbook/tests/test_tool_calling.py
```diff
@@ -0,0 +1,176 @@
...
+)
+def test_qwen3_parse_single_tool_call(model_name: str, renderer_name: str):
+    """Test parsing a single tool call from Qwen3 response."""
+    tokenizer = get_tokenizer(model_name)
+    renderer = get_renderer(renderer_name, tokenizer)
+
+    # Simulate model response with tool call
+    response_text = """I'll search for that information.
+<tool_call>
+{"name": "search", "args": {"query": "weather in NYC"}}
```
tyler-griggs, line 81: I believe Qwen tool calling format expects `arguments` instead of `args` (see docs [here](https://qwen.readthedocs.io/en/latest/getting_started/concepts.html#tool-calling) and tokenizer config [here](https://huggingface.co/Qwen/Qwen3-8B/blob/main/tokenizer_config.json)). This is a pre-existing issue, I can send out a follow-up PR to fix.

tyler-griggs (approved):

tinker_cookbook/tests/test_tool_calling.py
joschu: That'd be great, I'll merge this, and let's fix in a followup.

PR #220, opened 2025-12-24
Rename RenderedMessage fields: prefix/content/suffix → header/output/stop_overlap

joschu:
## Human-Written Overview

I noticed that in the large PR #161, `render_message` was upgraded to be a part of the `Renderer` interface, and its return value was changed to `RenderedMessage`. But when I was reading `RenderedMessage`, it wasn't clear to me what should go in the `prefix` / `content` / `suffix` members -- the docstring wasn't very clear. Later, I realized that `content` didn't just correspond to `message["content"]` -- it usually included the stop token, which I found confusing. So I renamed the fields on `RenderedMessage` and added more documentation in various places. I also realized that `Renderer` was a protocol but would make more sense as an `ABC`, so I changed that too.

It looks like a big diff, but most of the diff is from renaming.

Below summary is written by Claude.

---

## Summary

Two related improvements to the Renderer protocol:

### 1. Rename RenderedMessage fields for clarity

| Old Name | New Name | Why |
|----------|----------|-----|
| `prefix` | `header` | Clearly indicates this is the introductory role identifier |
| `content` | `output` | Clarifies this is what the model *generates*, not the message content |
| `suffix` | `stop_overlap` | Explicitly documents its edge-case nature |

**Problem with old names:**
- `content` was misleading—it includes end-of-turn tokens, not just message text
- `prefix`/`suffix` implied false symmetry, but `suffix` is an obscure edge case only RoleColonRenderer uses

### 2. Convert Renderer from Protocol to ABC

**Why ABC is more appropriate:**
- All renderers already explicitly inherit from `Renderer`, so we're not benefiting from Protocol's structural subtyping
- `@abstractmethod` provides static type checking (vs runtime `NotImplementedError`)
- The class provides default implementations (`build_generation_prompt`, `build_supervised_example`), which is an ABC pattern

### 3. Improved documentation

- Added mathematical formula showing token assembly (with caveat that it's for the simple case)
- Documented when `build_supervised_example` should be overridden
- Added concrete examples for each RenderedMessage field
- Fixed double-escaped newlines in docstrings

## Test plan
- [x] All 26 renderer tests pass
- [x] No functional changes

### DIFF
--- a/tinker_cookbook/renderers.py
+++ b/tinker_cookbook/renderers.py
@@ -9,6 +9,7 @@ import logging
 import re
 import urllib.request
 from datetime import datetime
+from abc import ABC, abstractmethod
 from enum import StrEnum
 from typing import Literal, NotRequired, Optional, Protocol, TypedDict, cast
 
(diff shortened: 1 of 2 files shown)


tyler-griggs (approved):
Looks great!

PR #221, opened 2025-12-24
Add tool specification helpers and comprehensive tool calling support

joschu:
## Human-Written Overview

This PR sets up a model-agnostic way of defining a bunch of tools, such that each renderer knows how to format the tool specs in the system message. We also implement proper support for parsing the tool calls in a bunch of models. 

I also refactored test_renderers.py so that the functions that test equivalence with the HF templates are parametrized, so that we can loop over the models and conversations. This led to finding some issues, e.g. with the gpt-oss renderer not including the analysis channel (CoT) in SFT. The renderers and testing infra still needs some work.

I wrote an end-to-end test test_tool_calling_e2e.py that shows the model a basic tool and checks that it emits a properly formatted tool call. 

Future work, and issues uncovered:

- GPT OSS tool calling support
- Check correspondence between token sequences from build_supervised_example and build_generation_prompt. I think there's currently a mismatch for the Qwen3 (hybrid) renderer. To avoid train-deploy mismatch, these should be consistent.
- Check that renderers satisfy sequence extension property (https://tinker-docs.thinkingmachines.dev/rl/sequence-extension) when expected

## Summary

Comprehensive tool calling support for all model families with HF template compatibility.

### New Features
- **`ToolSpec` TypedDict** - Type-safe tool definitions following OpenAI function calling format
- **`create_conversation_prefix_with_tools()`** - Standard API on all renderers for adding tools to conversations
- **`UnparsedToolCall`** - Captures tool calls that fail to parse (e.g., invalid JSON) for debugging

### Tool Calling Support by Model
- **Qwen3** - `<tool_call>` format with JSON body
- **Llama3** - `<function=name>args</function>` format with ipython role for tool responses
- **DeepSeek V3** - Official HF template format with special tokens (`<｜tool▁calls▁begin｜>` etc.)
- **Kimi K2** - `<|tool_calls_section_begin|>` format

### Fixes
- **DeepSeek tool format** - Fixed to match official HF template (removed code blocks, uses `## Tools` header with `### tool_name` subheaders)
- **System message handling** - Fixed for DeepSeek (supports `role="system"`), OpenAI/GPT-OSS (always includes analysis channel for train-test consistency)
- **Empty token chunk error** - Fixed tinker "Chunk N has empty tokens list" error for system messages

### Tests
- Parametrized HF compatibility tests across all models and conversation types
- Tool call parsing tests for all model families (valid JSON, invalid JSON, mixed)
- E2E test script (`scripts/test_tool_calling_e2e.py`) for production model testing

## Test plan

- [ ] `uv run pytest tinker_cookbook/tests/test_tool_calling.py -v`
- [ ] `uv run pytest tinker_cookbook/tests/test_renderers.py -v`
- [ ] Verify DeepSeek tool format matches official docs


tinker_cookbook/tests/test_tool_calling.py
```diff
@@ -74,7 +74,7 @@ def test_qwen3_parse_single_tool_call(model_name: str, renderer_name: str):
     tokenizer = get_tokenizer(model_name)
     renderer = get_renderer(renderer_name, tokenizer)
 
-    # Simulate model response with tool call
+    # Simulate model response with tool call (uses "arguments" per Qwen standard)
```
joschu: unnecessary change to comment

tinker_cookbook/tests/test_tool_calling.py
```diff
@@ -106,7 +106,7 @@ def test_qwen3_parse_multiple_tool_calls(model_name: str, renderer_name: str):
     tokenizer = get_tokenizer(model_name)
     renderer = get_renderer(renderer_name, tokenizer)
 
-    # Simulate model response with multiple tool calls
+    # Simulate model response with multiple tool calls (uses "arguments" per Qwen standard)
```
joschu: unnecessary change to comment

tinker_cookbook/renderers.py
```diff
@@ -223,7 +252,11 @@ def ensure_text(content: Content) -> str:
 
 
 def _tool_call_payload(tool_call: ToolCall) -> dict[str, object]:
-    """Minimal JSON payload for embedding in <tool_call> blocks."""
+    """Minimal JSON payload for embedding in <tool_call> blocks.
+
+    Uses "arguments" key per Qwen tool calling format:
```
joschu: this comment about "arguments" only pertains to the diff in this PR, is unnecessary

tinker_cookbook/renderers.py
```diff
@@ -553,8 +607,88 @@ def _end_message_token(self) -> int:
...
+            try:
+                # Validate JSON
+                json.loads(args_str)
+                tool_calls.append(
+                    ToolCall(
+                        function=ToolCall.FunctionBody(name=func_name, arguments=args_str),
+                    )
+                )
+            except json.JSONDecodeError:
+                continue
```
joschu: Seems like this'll just silently fail and omit the tool call if it's not parsed correctly. Need to come up with a coherent policy for what to do when various kinds of tool errors occur.

bledden:
I noticed the Llama3 rendering was missing from this branch (the PR description mentions it but the code didn't have it yet). I opened #223 to add:

- Tool responses render with `ipython` role and `{"output": ...}` wrapping
- Assistant messages with `tool_calls` render as `<function=name>{args}</function>`

This is separate from the `unparsed_tool_calls` error handling work - that's about parsing failures, whereas this is about rendering support for multi-turn tool conversations.

Also noticed the type-check is failing because `RoleColonRenderer` and `GptOssRenderer` are missing `create_system_prefix_with_tools` - happy to help with that if needed.

tinker_cookbook/tests/test_renderers.py
```diff
@@ -17,161 +17,354 @@
...
-        # Thinking field should not be rendered in this case as it is not the last message.
-        convo[1]["thinking"] = "The user is sharing a greeting. We should respond politely."
+        # Add thinking field to first assistant message for testing
+        for msg in convo:
+            if msg["role"] == "assistant" and "thinking" not in msg:
+                msg["thinking"] = "The user is sharing a greeting. We should respond politely."
+                break
         aug_convo = convo
     elif model_name.startswith("moonshotai"):
+        # Kimi K2 HF template adds default system message; we include it explicitly
```
joschu: I think this comment is unnecessary now?

tinker_cookbook/tests/test_renderers.py
```diff
@@ -17,161 +17,354 @@
...
+# =============================================================================
+
+
+def _prepare_conversation_for_model(
+    model_name: str, convo: list[Message], is_generation: bool
+) -> tuple[list[Message], list[Message]]:
+    """Prepare conversation for both cookbook renderer and HF template comparison.
+
+    Unfortunately, aug_convo and hf_convo sometimes need to differ because HF templates
+    auto-add certain content that our renderers expect to be present in the input:
```
joschu, line 170: Also point out that only certain Message fields are even used by the HF renderer. E.g. it doesn't know how to use the tool call fields. (Is that right?)

tinker_cookbook/tests/test_renderers.py
```diff
@@ -17,161 +17,354 @@
...
+        f"[{conv_desc}] Cookbook tokens: {cookbook_tokens}\n"
         f"Cookbook string: {tokenizer.decode(cookbook_tokens)}\n"
         f"HF tokens: {hf_tokens}\n"
         f"HF string: {tokenizer.decode(hf_tokens)}"
     )
 
 
+# =============================================================================
+# Tool Use Rendering Tests (no HF comparison - HF templates don't support tool calls)
+# =============================================================================
```
joschu, line 348: this is pointless now, right? we don't have any tool call conversations

tinker_cookbook/renderers.py
```diff
@@ -76,62 +76,26 @@ class FunctionBody(pydantic.BaseModel):
     """The function body containing tool name and arguments."""
 
 
-class ToolOk(StrictBase):
```
joschu, line 79: note that ToolOk / ToolError weren't being used before (whoops)

tinker_cookbook/renderers.py
```diff
@@ -290,6 +287,27 @@ def render_message(self, idx: int, message: Message, is_last: bool = False) -> R
     def parse_response(self, response: list[int]) -> tuple[Message, bool]:
         raise NotImplementedError
 
+    def create_system_prefix_with_tools(
```
joschu: maybe create_conversation_prefix_with_tools?

tinker_cookbook/tests/test_renderers.py
```diff
@@ -17,161 +17,402 @@
...
+    # Deep copy to avoid mutating the original
+    convo = copy.deepcopy(convo)
+
+    # Apply model-specific modifications first, then create hf_convo
     if model_name.startswith("meta"):
         today = date.today().strftime("%d %b %Y")
         system_msg: Message = {
             "role": "system",
-            "content": f"Cutting Knowledge Date: December 2023\nToday Date: {today}\n\n",
+            # No trailing newlines - HF template uses | trim which strips them
```
joschu: rm this comment, relevant to diff but not the code

tinker_cookbook/renderers.py
```diff
@@ -1062,32 +1276,173 @@ def _end_message_token(self) -> int:
...
+        self, tools: list[ToolSpec], system_prompt: str = ""
+    ) -> list[Message]:
+        """Create system/user message with DeepSeek V3 tool specifications.
+
+        DeepSeek V3 uses JSON function definitions in the system prompt.
+        Tool calls use special tokens with JSON in code blocks.
+
+        Note: DeepSeek V3 doesn't support system role natively. Returns user role
+        unless system_role_as_user is True (in which case system messages are
+        automatically converted to user role by render_message).
```
joschu: I think Deepseek does accept having an initial system message (which doesn't have the system header). Can you double check the chat template for this?

bledden:
Heads up - PR #220 renames RenderedMessage fields (prefix -> header, content -> output, suffix -> stop_overlap). Once that merges, this PR will need a rebase. PRs #223, #224, and #226 (tool rendering additions) are stacked on this (#221) and will also need updating upon an approval and merge of #220 (which I have mostly ready, depending on updates you make here on #221).

joschu:
@bledden -- Agreed, I'm going to rewrite #220 to work on top of this.
Re: your other PRs, I'm pretty sure I have everything except GPT-OSS working here, so let's focus on #224 after this merges. That said, I'm not 100% sure the deepseek implementation gets all the nitty gritty details correct, e.g. whether or not to hide thinking traces in multi-turn trajectories, so if you could look into some of those details, that'd be great

bledden:
Looked into the thinking trace handling question and I found a few things to fix, opened #229 to address them:

1. **Thinking in history** - wasn't stripping it from completed turns. I added `strip_thinking_from_history` param (default=True) to match what we do for Qwen3. For RL where extension property matters, set to False.

2. **Post-tool formatting** - assistant messages after tool output had the wrong format (role token + `</think>` prefix where there shouldn't be any).

3. **When to preserve thinking** - it's not as simple as "always strip from history". Messages with `tool_calls` should keep their thinking (it's the reasoning behind the call, same turn). Post-tool responses too (reasoning about the tool result).

Tests are passing and the approach follows the same pattern as Qwen3. Let me know if you'd prefer different defaults.

tinker_cookbook/tests/test_renderers.py
```diff
@@ -38,140 +47,409 @@ def _load_tokenizer(model_name: str) -> Any:
...
     elif model_name.startswith("moonshotai"):
         aug_convo = convo
     else:
         raise ValueError(f"Unknown model name: {model_name}")
 
+    return aug_convo, convo
+
+
+# Test matrix: models x conversations for generation tests
+_GENERATION_TEST_PARAMS = [
```
tyler-griggs, line 243: nice!

bledden:
One thing I noticed while working on #229: the current `build_generation_prompt` in #221 correctly skips the assistant prefix when the *last* message is a tool (for generation). But for *historical* assistant messages that follow tool responses in multi-turn conversations, they still get the `<|Assistant|>` role token and `</think>` prefix.

Based on the HF template behavior, these post-tool assistant messages in history should:
- Skip the role token (content flows directly after `<|tool_output_end|>`)
- Skip the `</think>` prefix
- Preserve any thinking traces

#229 adds a `follows_tool` parameter to `render_message` to handle this. Wanted to flag in case you'd prefer to fold this into #221 directly rather than having it as a follow-up PR, if you agree with that behavior.

joschu:
@bledden Your approach sounds good to me. I'm going to try to get this and #220 in and then direct attention to your PRs for deepseek and gpt-oss. I also started on #230 to help us verify the hybrid model behavior.

tyler-griggs (approved):

PR #223, opened 2025-12-24
Add Llama3 tool call and tool response rendering

bledden:
While reviewing [this PR](https://github.com/thinking-machines-lab/tinker-cookbook/pull/221), I noticed the Llama3Renderer has parsing support but the corresponding rendering is missing. The PR description mentions "Llama3 tool response rendering (ipython role)" but looking at the code, `render_message` doesn't actually handle it yet.

This adds:
- Tool responses render with `ipython` role and `{"output": ...}` wrapping
- Assistant messages with `tool_calls` render as `<function=name>{args}</function>`

I also added tests for both.

This is separate from the error handling discussion - that's about parsing failures, whereas this is about rendering support for multi-turn tool conversations.

Reference: https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_1/

### DIFF
--- a/tinker_cookbook/renderers.py
+++ b/tinker_cookbook/renderers.py
@@ -565,6 +565,25 @@ class RoleColonRenderer(Renderer):
         assert isinstance(bos_token_str, str)
         return self.tokenizer.encode(bos_token_str, add_special_tokens=False)
 
+    def create_system_prefix_with_tools(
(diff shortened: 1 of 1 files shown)


joschu:
Thanks for implementing this. I ended up implementing llama support in #219.

PR #229, opened 2025-12-26
fix: DeepSeek thinking traces and post-tool formatting

bledden:
## Summary

Addresses Josh's question in #221 about whether DeepSeek gets thinking trace handling right in multi-turn trajectories.

After investigating, I found a few issues with how we handle thinking traces and post-tool formatting:

- **Thinking traces in history**: Like Qwen3, we should strip `<think>...</think>` from completed assistant turns by default. Added `strip_thinking_from_history` param (default=True) to match Qwen3's behavior. For RL training where you need extension property, set it to False.

- **Post-tool assistant formatting**: Messages right after a tool response shouldn't have the `<|Assistant|>` role token or `</think>` prefix - the content flows directly after `<|tool_output_end|>`.

- **Thinking preservation in certain cases**: Not all thinking should be stripped. Messages with `tool_calls` keep their thinking (it's the model's reasoning about *why* it's making that call - part of the same turn). Post-tool responses also preserve thinking since it's reasoning about *how* to interpret the tool result.

The default behavior matches what we already do for Qwen3. Let me know if the team prefers a different default.

## Test plan

- [x] Validated with 8 test scenarios covering all cases
- [x] Existing tests still pass

### DIFF
--- a/tinker_cookbook/tests/test_renderers.py
+++ b/tinker_cookbook/tests/test_renderers.py
@@ -923,9 +923,7 @@ def test_deepseek_post_tool_formatting():
 
             # Content should not start with </think>
             content_chunk = rendered["content"][0]
-            assert isinstance(content_chunk, tinker.EncodedTextChunk), (
-                "Expected EncodedTextChunk"
-            )
+            assert isinstance(content_chunk, tinker.EncodedTextChunk), "Expected EncodedTextChunk"
             content_str = tokenizer.decode(list(content_chunk.tokens))
             assert not content_str.startswith("</think>"), (
                 f"Post-tool assistant should not have </think> prefix: {content_str}"


joschu:
We need to think through the Renderer interface; non-ideal that we have to change the `render_message` arguments (but I see that it's necessary here). This suggests that `render_message` might not belong in the base class.

bledden:
I prototyped an alternative approach on the `experiment/render-hints` branch (#231) that might address your concern about the `render_message` interface while preserving code reuse.

**The core idea:** Separate context computation from rendering by introducing a `compute_render_hints()` method that returns a list of `RenderHints` (a TypedDict with fields like `is_last`, `follows_tool`). The base class computes default hints, and `render_message` receives the pre-computed hints rather than computing context itself.

This keeps the loop logic in `build_generation_prompt` and `build_supervised_example` centralized while giving renderers the context they need.

### Comparison: Removing `render_message` from base vs. `RenderHints` pattern

| Aspect | Remove from base | RenderHints pattern |
|--------|------------------|---------------------|
| Flexibility | Maximum - each renderer controls everything | High - hints enable context-aware rendering |
| Code reuse | ~130 lines duplicated per renderer | No duplication |
| TrainOnWhat logic | Duplicated in 10 renderers | Single implementation |
| New renderer effort | ~150 lines to implement | ~30 lines to implement |
| Maintenance | Bug fixes needed in 10 places | Centralized |

The `TrainOnWhat` weight-assignment logic in `build_supervised_example` (~30 lines) is model-agnostic - it's seems like it is about *which* messages to train on, not *how* to render them. Removing `render_message` from the base class would require duplicating this across all renderers if I am not mistaken.

That said, if there's a strong preference for renderers to own their full rendering pipeline, the your suggestion of removing render_message from the base class provides maximum flexibility at the cost of code duplication/added maintenance. Happy to discuss further or adjust the approach.

joschu:
Thanks, I built on this branch and then merged in #237

PR #238, opened 2025-12-26
Refactor thinking from Message.thinking field to ThinkingPart in content list

joschu:
## Human-Written Overview

This PR moves to making message content more structured. Previously we would typically have content strings with <think> or <function_calls> blocks inside, now we move towards using a list of content chunks. Our previous code was also inconsistent -- there was a thinking attribute on messages, which was only being used by gpt-oss and kimi. Now all the models are handled more uniformly.

This change will break a lot of code that assumes that content is a string, but I think it's for the best, in the long run. When I originally wrote the code, we were mostly dealing with the Qwen3 models and text-only data, and I figured that we should keep it simple until we have more use cases. Now with images and tool calling and a handful more models, the old string-based setup was getting really confusing, e.g., the rendering code had lots of string splitting and substitution.

---

## Summary

- Refactors how thinking/reasoning content is represented in `Message` type
- Moves from separate `thinking` field to `ThinkingPart` in the content list (similar to kosong's approach)
- Enables cleaner thinking stripping without string manipulation
- Adds `get_text_content()` helper for extracting text from messages after `parse_response`

### Core refactor
- Add `ThinkingPart` TypedDict with `type: Literal["thinking"]` and `thinking: str`
- Update `ContentPart = TextPart | ImagePart | ThinkingPart | ToolCallPart | UnparsedToolCallPart`
- Remove `thinking` field from `Message` TypedDict
- Add `ensure_list()` utility to normalize content to list form
- Add `remove_thinking()` utility to filter out ThinkingPart elements
- Add `get_text_content()` utility to extract text from messages (stripping thinking)
- Update `GptOssRenderer` to parse multi-channel format into ThinkingPart/TextPart
- Update `KimiK2Renderer` to use `ensure_list` + `remove_thinking`
- Update `KimiK2Renderer.parse_response` to return `ThinkingPart` in content
- Update `Qwen3Renderer` to support `ThinkingPart` stripping via list filtering
- Update `DeepSeekV3ThinkingRenderer` to support `ThinkingPart` stripping
- Update `_preprocess_message_parts` to filter out ThinkingPart (VL renderers)

## Breaking Changes

1. **`Message.thinking` field removed**: Code that accesses `message["thinking"]` or `message.get("thinking")` will need to use the new utility functions or check the content list directly.

2. **`parse_response` may return list content**: For thinking models (Qwen3, DeepSeek, GptOss, KimiK2), when thinking/reasoning is present, `message["content"]` is now a list containing `ThinkingPart` and `TextPart` instead of a string.

3. **Code using `ensure_text` after `parse_response` may break**: If the model produces thinking content, `ensure_text(message["content"])` will fail. Use `get_text_content(message)` instead.

## Migration Guide

**Extracting text content after `parse_response`:**
```python
# Before (breaks with thinking models)
message, _ = renderer.parse_response(tokens)
content = ensure_text(message["content"])  # Fails if content is a list!

# After (works with all models)
from tinker_cookbook.renderers import get_text_content

message, _ = renderer.parse_response(tokens)
content = get_text_content(message)  # Extracts text, strips thinking
```

**Reading thinking from message:**
```python
from tinker_cookbook.renderers import ensure_list

parts = ensure_list(message["content"])
thinking = "".join(p["thinking"] for p in parts if p["type"] == "thinking")
text = "".join(p["text"] for p in parts if p["type"] == "text")
```

**Creating message with thinking:**
```python
from tinker_cookbook.renderers import ThinkingPart, TextPart

message = {
    "role": "assistant",
    "content": [
        ThinkingPart(type="thinking", thinking="Let me think..."),
        TextPart(type="text", text="Hello"),
    ],
}
```

## Test plan

- [x] All 68 renderer tests pass
- [x] Type checking passes (pyright 0 errors)
- [x] Pre-commit checks pass
- [ ] Manual testing with thinking models


tinker_cookbook/renderers.py
```diff
@@ -241,6 +279,184 @@ def ensure_text(content: Content) -> str:
...
+
+    return ToolCall(
+        function=ToolCall.FunctionBody(name=name, arguments=json.dumps(arguments)),
+        id=tool_id,
+    )
+
+
+def parse_content_blocks(
+    content: str,
+) -> tuple[list[ContentPart], list[UnparsedToolCall]]:
```
joschu: This is an ugly return type. Ideally we return one list and somehow work the unparsed tool calls into that list.

tinker_cookbook/renderers.py
```diff
@@ -241,6 +279,184 @@ def ensure_text(content: Content) -> str:
...
+    Returns:
+        A list of ContentPart elements. If content is a string, returns a single
+        TextPart wrapping that string.
+    """
+    if isinstance(content, str):
+        return [TextPart(type="text", text=content)]
+    return content
+
+
+def remove_thinking(content: Content) -> list[ContentPart]:
```
joschu: can we make the argument list[ContentPart] as well? seems slightly cleaner

tinker_cookbook/renderers.py
```diff
@@ -241,6 +279,184 @@ def ensure_text(content: Content) -> str:
...
+    if not isinstance(name, str):
+        return UnparsedToolCall(raw_text=raw_text, error="Missing or invalid 'name' field")
+    if not isinstance(arguments, dict):
+        return UnparsedToolCall(raw_text=raw_text, error="Missing or invalid 'arguments' field")
+
+    if tool_id is not None and not isinstance(tool_id, str):
+        tool_id = None
+
+    return ToolCall(
+        function=ToolCall.FunctionBody(name=name, arguments=json.dumps(arguments)),
```
joschu, line 338: this is ugly, but I see that it's because we store arguments as raw json
please add a TODO about revisiting this choice

tinker_cookbook/renderers.py
```diff
@@ -324,11 +540,17 @@ def __init__(self, tokenizer: Tokenizer):
...
-        return (
-            message["content"]
-            if isinstance(message["content"], list)
-            else [TextPart(type="text", text=message["content"])]
-        )
+        """Convert message content to list form, filtering out ThinkingPart and ToolCallPart.
+
+        ThinkingPart and ToolCallPart are filtered because most renderers
+        (especially VL renderers) don't support them in the message parts.
+        """
```
joschu: This is a strange method to have on the base class. why is it here? Is it just for the VL renderer? If so, why not move it there?

tinker_cookbook/renderers.py
```diff
@@ -890,22 +1103,45 @@ def _wrap_qwen_tool_response(self, content: str) -> str:
...
+                # For the last message, serialize to <think>content</think>text format
+                thinking_content = "".join(p["thinking"] for p in parts if p["type"] == "thinking")
+                text_content = "".join(p["text"] for p in parts if p["type"] == "text")
+                if thinking_content:
+                    output_content = f"<think>{thinking_content}</think>{text_content}"
+                else:
+                    output_content = text_content
+        else:
+            # Content is string (possibly with embedded <think> tags) - use string manipulation
+            output_content = ensure_text(content)
```
joschu: what if it has tool calls but not thinking? doesn't this throw an error?

tinker_cookbook/renderers.py
```diff
@@ -1383,25 +1574,34 @@ def render_message(self, message: Message, ctx: RenderContext) -> RenderedMessag
...
+            isinstance(content, list)
+            and any(p["type"] == "thinking" for p in content)
+        )
+
+        # Only use ensure_text for non-ThinkingPart content
+        # (ThinkingPart content is handled separately in the assistant branch)
+        if not has_thinking_part:
+            content_str = ensure_text(content)
+        else:
+            content_str = ""  # Will be set in assistant branch
```
joschu: this seems fragile. can you refactor, or set this to some value that'll fail more loudly, if code gets changed later and we forget to set it?

tinker_cookbook/renderers.py
```diff
@@ -1423,15 +1646,15 @@ def render_message(self, message: Message, ctx: RenderContext) -> RenderedMessag
                 role_token = self._get_special_token("Assistant")
                 header_tokens = [role_token]
 
-                # Handle thinking traces in history
+                # Legacy string-based stripping for content without ThinkingPart
```
joschu: We don't need to handle stripping of <think> blocks in the string content. Let's remove all the code that does this from both Qwen and Deepseek to simplify.

tinker_cookbook/tests/test_renderers.py
```diff
@@ -935,3 +956,565 @@ def test_deepseek_post_tool_formatting():
...
                 f"Post-tool assistant should not have </think> prefix: {output_str}"
             )
+
+
+# =============================================================================
+# ThinkingPart Tests
+# =============================================================================
+
+
+def test_ensure_list_string_input():
```
joschu: trivial, remove

tinker_cookbook/tests/test_renderers.py
```diff
@@ -935,3 +956,565 @@ def test_deepseek_post_tool_formatting():
...
+    """Test that ensure_list normalizes string content to list form."""
+    content = "Hello world"
+    result = ensure_list(content)
+    assert len(result) == 1
+    assert result[0]["type"] == "text"
+    assert result[0]["text"] == "Hello world"
+
+
+def test_ensure_list_list_input():
```
joschu: trivial, remove

tinker_cookbook/tests/test_renderers.py
```diff
@@ -935,3 +956,565 @@ def test_deepseek_post_tool_formatting():
...
+    """Test that ensure_list returns list content unchanged."""
+    content = [
+        TextPart(type="text", text="Hello"),
+        ThinkingPart(type="thinking", thinking="Let me think..."),
+    ]
+    result = ensure_list(content)
+    assert result == content
+
+
+def test_remove_thinking():
```
joschu: trivial, remove

tinker_cookbook/tests/test_renderers.py
```diff
@@ -935,3 +956,565 @@ def test_deepseek_post_tool_formatting():
...
+        ThinkingPart(type="thinking", thinking="Let me think..."),
+        TextPart(type="text", text="The answer is 42."),
+    ]
+    result = remove_thinking(content)
+    assert len(result) == 1
+    assert result[0]["type"] == "text"
+    assert result[0]["text"] == "The answer is 42."
+
+
+def test_gpt_oss_with_thinking_part_in_content():
```
joschu: remove, should be covered by HF equivalence

tinker_cookbook/tests/test_renderers.py
```diff
@@ -935,3 +956,565 @@ def test_deepseek_post_tool_formatting():
...
+    output_tokens = [tok for chunk in rendered.output for tok in chunk.tokens]
+    output_str = tokenizer.decode(output_tokens)
+
+    # Should contain thinking in analysis channel
+    assert "analysis" in output_str
+    assert "Step 1: Think about it." in output_str
+    assert "The answer is 42." in output_str
+
+
+def test_kimi_k2_with_thinking_part_in_content():
```
joschu: remove, should be covered by HF equivalence

joschu: Please add documentation that we don't need to test things about build_generation_prompt that'll clearly be verified by the HF equivalence tests. Remove tests like this. OTOH HF doesn't do parsing so we need those tests. Please add comments to the top of this file to codify this guideline.

tinker_cookbook/tests/test_renderers.py
```diff
@@ -935,3 +956,565 @@ def test_deepseek_post_tool_formatting():
...
+    text_parts = [p for p in parts if p["type"] == "text"]
+
+    assert len(thinking_parts) == 1
+    assert thinking_parts[0]["thinking"] == "Let me think step by step."
+
+    assert len(text_parts) == 1
+    assert text_parts[0]["text"] == "The final answer is 7."
+
+
+def test_qwen3_stripping_with_thinking_part():
```
joschu: this should be tested by HF equivalence -- remove

tinker_cookbook/tests/test_renderers.py
```diff
@@ -935,3 +956,565 @@ def test_deepseek_post_tool_formatting():
...
+
+    output_tokens = [tok for chunk in rendered.output for tok in chunk.tokens]
+    output_str = tokenizer.decode(output_tokens)
+
+    # Should serialize thinking to <think>...</think> format
+    assert "<think>Internal reasoning...</think>" in output_str
+    assert "The visible response." in output_str
+
+
+def test_deepseek_stripping_with_thinking_part():
```
joschu: remove

tinker_cookbook/tests/test_renderers.py
```diff
@@ -935,3 +956,565 @@ def test_deepseek_post_tool_formatting():
...
+    # Invalid tool call is included as UnparsedToolCallPart, text is still captured
+    assert len(parts) == 2
+    assert parts[0]["type"] == "unparsed_tool_call"
+    assert "Invalid JSON" in parts[0]["error"]
+    assert parts[1] == TextPart(type="text", text="text after")
+
+
+# =============================================================================
+# Qwen3 parse_response Tests
+# =============================================================================
```
joschu, line 1061: Please shorten this code by removing trivial assertions and tests. Just keep tests that are checking some code that involves tricky logic.

tinker_cookbook/tests/test_renderers.py
```diff
@@ -935,3 +956,565 @@ def test_deepseek_post_tool_formatting():
...
+    assert content[0]["type"] == "tool_call"
+
+    # Backward-compatible field
+    assert len(message["tool_calls"]) == 1
+
+
+# =============================================================================
+# DeepSeek parse_response Tests
+# =============================================================================
+
```
joschu, line 1183: Shorten, following guidelines above from qwen.

tinker_cookbook/tests/test_renderers.py
```diff
@@ -935,3 +956,565 @@ def test_deepseek_post_tool_formatting():
...
+    assert content[0] == ThinkingPart(type="thinking", thinking="step 1")
+    assert content[1] == TextPart(type="text", text="partial")
+    assert content[2] == ThinkingPart(type="thinking", thinking="step 2")
+    assert content[3] == TextPart(type="text", text="final")
+
+
+# =============================================================================
+# GptOss parse_response Tests
+# =============================================================================
+
```
joschu, line 1248: shorten if possible

tinker_cookbook/renderers.py
```diff
@@ -132,8 +132,70 @@ class ImagePart(TypedDict):
     image: str | Image.Image
 
 
+class ThinkingPart(TypedDict):
+    """
+    Container for a thinking/reasoning part in a message.
```
joschu: greatly shorten the docstring. just one line docstring, and then put a comment after each field. apply the same guideline to related types.

tinker_cookbook/renderers.py
```diff
@@ -241,6 +302,186 @@ def ensure_text(content: Content) -> str:
     raise ValueError(f"Expected text content, got multimodal content with {len(content)} parts")
 
 
+def ensure_list(content: Content) -> list[ContentPart]:
+    """
+    Normalize content to list form.
```
joschu: shorten docstring

tinker_cookbook/renderers.py
```diff
@@ -241,6 +302,186 @@ def ensure_text(content: Content) -> str:
...
+        TextPart wrapping that string.
+    """
+    if isinstance(content, str):
+        return [TextPart(type="text", text=content)]
+    return content
+
+
+def remove_thinking(parts: list[ContentPart]) -> list[ContentPart]:
+    """
+    Remove ThinkingPart elements from content.
```
joschu: shorten docstring

tinker_cookbook/renderers.py
joschu: this seems unresolved

tinker_cookbook/renderers.py
```diff
@@ -1210,9 +1398,24 @@ def __init__(self, tokenizer: Tokenizer, image_processor: ImageProcessor):
...
+        """
+        content = message["content"]
+        if isinstance(content, str):
+            base_parts: list[ImagePart | TextPart] = [TextPart(type="text", text=content)]
+        else:
+            # Filter out parts that VL renderers don't support
+            base_parts = cast(
+                list[ImagePart | TextPart],
+                [p for p in content if p["type"] in ("text", "image")],
+            )
```
joschu: are you sure tool calling doesn't work for VL renderers?

tinker_cookbook/renderers.py
```diff
@@ -1410,9 +1567,29 @@ def render_message(self, message: Message, ctx: RenderContext) -> RenderedMessag
...
+                    and not ctx.is_last
+                ):
+                    # Remove thinking parts for historical messages
+                    parts = remove_thinking(parts)
+                    output_content = "".join(p["text"] for p in parts if p["type"] == "text")
+                else:
+                    # Render inline: <think>...</think> for thinking, text for text
+                    thinking = "".join(p["thinking"] for p in parts if p["type"] == "thinking")
+                    text = "".join(p["text"] for p in parts if p["type"] == "text")
+                    output_content = f"<think>{thinking}</think>{text}" if thinking else text
```
joschu: we shouldn't ever create text with think tags in it. why are we doing it?

PR #224, opened 2025-12-25
Complete GptOss Harmony format support (tool calling, CoT extraction)

bledden:
## Summary

This PR adds comprehensive GptOss (OpenAI Harmony format) support, building on #238's ThinkingPart refactor.

**Note: This PR depends on #238 being merged first.** The branch is rebased on top of #238.

### Tool Calling Support
- **render_message**: Tool calls rendered in commentary channel with `to=functions.{name}` routing
- **render_message**: Tool responses from `functions.{name} to=assistant` format
- **parse_response**: Extracts tool calls from commentary channel
- **create_system_prefix_with_tools**: TypeScript-style namespace definitions
- **Stop sequences**: Both `<|return|>` (response end) and `<|call|>` (tool call) tokens

### Thinking/CoT Extraction (fixes #239)
- **parse_response**: Extracts analysis channel content into `ThinkingPart`
- Handles full format: `<|channel|>analysis<|message|>{thinking}<|end|>`
- Handles continuation format: `{thinking}<|end|>` (when generation prompt ends with analysis channel header)

### Tests
- 8 new tests for GptOss tool calling and thinking extraction
- All 92 tests pass (excluding gated Llama model tests)
- Pyright passes with 0 errors

## Example

```python
from tinker_cookbook.renderers import GptOssRenderer, get_renderer
from tinker_cookbook.tokenizer_utils import get_tokenizer

tokenizer = get_tokenizer("openai/gpt-oss-20b")
renderer = get_renderer("gpt_oss_medium_reasoning", tokenizer)

# Parse response with thinking and tool call
response = "<|channel|>analysis<|message|>Let me check the weather.<|end|><|start|>assistant<|channel|>commentary to=functions.get_weather<|constrain|>json<|message|>{\"city\": \"Tokyo\"}<|call|>"
tokens = tokenizer.encode(response, add_special_tokens=False)
message, success = renderer.parse_response(tokens)

# message["content"] is a list with ThinkingPart and optionally TextPart
# message["tool_calls"] contains the parsed tool call
```

## Test Plan
- [x] All 92 existing tests pass
- [x] Pyright type checking passes
- [x] New GptOss tests added for tool calling and thinking extraction
- [ ] Manual testing with Tinker API (validated thinking extraction works)

## Related Issues
- Fixes #239 (gpt-oss renderer doesn't expose raw CoT/reasoning as a message)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #249, opened 2025-12-27
Improve GptOss tool calling and add Kimi K2 tool_call_id warning

joschu:
## Summary

- Refactor GptOss system prompt handling to prepend as a message instead of using `_bos_tokens`
- Add warning to Kimi K2 renderer when `tool_call_id` is missing from tool result messages
- Document `tool_call_id` and `name` fields in Message type for tool result messages
- Fix pyright errors in gpt_oss.py

### GptOss Renderer
- System prompt is now prepended as a message via `build_generation_prompt`/`build_supervised_example` instead of using `_bos_tokens` (which should only contain actual BOS tokens)
- Internal `_INTERNAL_SYSTEM_ROLE` distinguishes OpenAI's system prompt from user-provided system messages
- User "system" messages map to "developer" role with "# Instructions" wrapper (matching HF template)

### Kimi K2 Renderer  
- Added warning when `tool_call_id` is missing from tool result messages (similar to GptOss warning for missing `name`)

### Message Type Documentation
- Documented `tool_call_id` field: For correlating results by ID (used by Kimi K2)
- Documented `name` field: For identifying function by name (used by GptOss)

## Related

Supersedes #224 - incorporates the tool calling support but addresses review feedback:
- Keeps renderer stateless (no `_has_function_tools` or `_tool_call_id_to_name` instance variables)
- Uses `None` for tool call IDs since Harmony format doesn't include them
- Proper system prompt handling

## Test plan

- [x] `pytest tinker_cookbook/tests/test_renderers.py` - all pass
- [x] `pytest tinker_cookbook/tests/test_renderer_parsing.py` - all pass
- [x] pyright passes
- [x] pre-commit passes


tinker_cookbook/renderers/gpt_oss.py
```diff
@@ -1,40 +1,193 @@
 """
-GptOssRenderer - OpenAI's open source model format.
+GptOssRenderer - OpenAI's open source model format (Harmony).
 
 Format like this (no newlines between messages, last message should end with <|return|> but
 be replaced by <|end|> when continuing the convo):
     <|start|>system<|message|>You are ChatGPT...<|end|><|start|>user<|message|>How much is 1+1?<|end|><|start|>assistant<|channel|>final<|message|>2<|end|><|start|>
+
+Harmony channels:
```
joschu: please say a bit more about channels -- one per assistant message? what is the channel intended to control?

joschu: this docstring is redundant to the class docstring. just keep one of them, probably the class one, and make the other really minimal

tinker_cookbook/renderers/gpt_oss.py
```diff
@@ -103,93 +286,319 @@ def render_message(self, message: Message, ctx: RenderContext) -> RenderedMessag
...
+        # GptOss has no BOS token. System prompt is prepended as a message.
+        return []
+
+    def build_generation_prompt(
+        self, messages: list[Message], role: Role = "assistant", prefill: str | None = None
+    ) -> tinker.ModelInput:
+        """Build generation prompt, prepending system message if configured."""
+        system_msg = self._get_system_message()
+        if system_msg:
+            messages = [system_msg] + list(messages)
```
joschu, line 392: this might cause issues if the user tries to provide their own system message. can we detect this and either not prepend or emit a warning that something's probably wrong?

PR #267, opened 2025-12-30
Fix DeepSeekV3ThinkingRenderer to trigger thinking mode

### LINKED ISSUE #264: deepseek renderer DeepSeekV3ThinkingRenderer doesn't sample with thinking
After #238 DeepSeekV3DisableThinkingRenderer implementation adds `</think>` to header [as expected](https://huggingface.co/deepseek-ai/DeepSeek-V3.1#non-thinking), but somehow confusingly DeepSeekV3ThinkingRenderer does *not* [add `<think>`](https://huggingface.co/deepseek-ai/DeepSeek-V3.1#thinking). 

That makes it difficult for a user of this API to sample with reasoning from DeepSeek as it requires knowing that one has to add this token to BOTH request and response sides  (otherwise it's not going to be parsed into ThinkingPart) manually before sampling.

Using OpenAI-compatible API it's not even possible to do the right thing, as adding a suffix to the prompt results in `<｜begin▁of▁sentence｜><｜User｜>prompt<think><｜Assistant｜>` (although in practice that seems to trigger reasoning, but, similar to gpt-oss in #239, it doesn't get parsed into "reasoning_content" field in the assistant message)

joschu:
## Summary

- Fixes #264: `DeepSeekV3ThinkingRenderer` now correctly adds `<think>` to generation prompts to trigger DeepSeek's reasoning mode
- Matches HuggingFace template behavior with `thinking=True`
- Refactored renderer to use a private base class `_DeepSeekV3BaseRenderer` for cleaner inheritance

## Changes

**Renderer (`tinker_cookbook/renderers/deepseek_v3.py`):**
- `DeepSeekV3ThinkingRenderer.build_generation_prompt()` adds `<think>` via prefill
- `DeepSeekV3ThinkingRenderer.render_message()` adds `</think>` to supervised assistant headers (when `strip_thinking_from_history=True`)
- No generation prompt added after tool responses (matches HF behavior)

**Tests (`tinker_cookbook/tests/test_renderers.py`):**
- Extended HF equivalence tests to cover thinking mode with `thinking=True`
- Added skip condition for consistency tests (HF thinking mode has different headers for supervised vs generation by design)

## Test plan

- [x] All DeepSeek renderer tests pass (26 passed, 9 skipped)
- [x] Full renderer test suite passes (102 passed, 25 skipped)
- [x] HF equivalence tests verify thinking renderer matches `thinking=True` behavior

### DIFF
--- a/tinker_cookbook/renderers/deepseek_v3.py
+++ b/tinker_cookbook/renderers/deepseek_v3.py
@@ -22,7 +22,6 @@ from tinker_cookbook.renderers.base import (
     ensure_text,
     parse_response_for_stop_token,
     parse_think_blocks,
-    remove_thinking,
 )
 from tinker_cookbook.tokenizer_utils import Tokenizer
 
@@ -105,9 +104,7 @@ class _DeepSeekV3BaseRenderer(Renderer):
 
             # Determine if we should strip thinking content from this message
             should_strip_thinking = (
-                self.strip_thinking_from_history
-                and not has_tool_calls
-                and not ctx.is_last
+                self.strip_thinking_from_history and not has_tool_calls and not ctx.is_last
             )
 
             if isinstance(content, list):
--- a/tinker_cookbook/renderers/deepseek_v3.py
+++ b/tinker_cookbook/renderers/deepseek_v3.py
@@ -27,14 +27,12 @@ from tinker_cookbook.renderers.base import (
 from tinker_cookbook.tokenizer_utils import Tokenizer
 
 
-class DeepSeekV3ThinkingRenderer(Renderer):
+class _DeepSeekV3BaseRenderer(Renderer):
     """
-    Renderer for DeepSeek V3 models in THINKING mode.
+    Base renderer for DeepSeek V3 models with common rendering logic.
 
-    Format:
-        <|begin_of_sentence|><|User|>question<|Assistant|><think>reasoning</think>answer<|end_of_sentence|>
-
-    For non-thinking mode, use DeepSeekV3DisableThinkingRenderer instead.
+    This is a private base class. Use DeepSeekV3ThinkingRenderer or
+    DeepSeekV3DisableThinkingRenderer instead.
 
     System messages at position 0 are rendered without role tokens (matching HF template).
     System messages at later positions require system_role_as_user=True to convert to user role.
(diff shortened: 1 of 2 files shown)


tinker_cookbook/renderers/deepseek_v3.py
```diff
@@ -363,7 +361,73 @@ def create_conversation_prefix_with_tools(
...
+    def render_message(self, message: Message, ctx: RenderContext) -> RenderedMessage:
+        """Render message with </think> header for assistant messages.
+
+        When strip_thinking_from_history=True (default), ALL assistant messages
+        (not following tool) get </think> added to their header. This matches
+        HF template behavior with thinking=True.
+        """
+        rendered = super().render_message(message, ctx)
+
+        # Add </think> to header for assistant messages with content (not following tool)
```
joschu: are you implying that messages that follow the tool don't have content? why not?

tinker_cookbook/renderers/deepseek_v3.py
```diff
@@ -363,7 +361,73 @@ def create_conversation_prefix_with_tools(
...
+        # when strip_thinking_from_history is True. This matches HF thinking=True behavior.
+        # Skip empty assistant messages (used for generation prompt position).
+        follows_tool = ctx.prev_message is not None and ctx.prev_message["role"] == "tool"
+        has_content = bool(message.get("content"))
```
joschu: not sure why this is here

tinker_cookbook/renderers/deepseek_v3.py
```diff
@@ -363,7 +361,73 @@ def create_conversation_prefix_with_tools(
...
+        should_add_think_close = (
+            message["role"] == "assistant"
+            and not follows_tool
+            and self.strip_thinking_from_history
+            and has_content
+        )
```
joschu, line 402: if we strip thinking from history, don't we want something like <think></think> not just the closing tag? I don't think we should be injecting strings like this -- this processing should involve iterating over the list and looking for ThinkPart tags. When stripping tags, we should replace the interior by "" or newlines, depending on what the specification says.

tinker_cookbook/renderers/deepseek_v3.py
```diff
@@ -363,7 +361,73 @@ def create_conversation_prefix_with_tools(
...
+        if should_add_think_close:
```
joschu, line 404: this seems hacky; should be looking at the parts and adding think tags when we render the parts. The only think tag that we insert should be the one in the generation prompt that's basically a prefill.

tinker_cookbook/renderers/deepseek_v3.py
```diff
@@ -363,7 +361,73 @@ def create_conversation_prefix_with_tools(
...
+        # HF only adds generation prompt (with <think>) when last message is user, not tool
+        if messages and messages[-1]["role"] == "tool":
+            # After tool response, don't add <think> - just use base behavior
+            return super().build_generation_prompt(messages, role, prefill)
+
+        # Use prefill to add <think> at the end of the generation prompt
+        think_prefill = "<think>"
+        if prefill:
+            think_prefill = think_prefill + prefill
+        return super().build_generation_prompt(messages, role, think_prefill)
```
joschu, line 429: this can be simplified by just saying something like super().build_generation_prompt(messages, role, maybe_think_prefill + prefill)

tinker_cookbook/renderers/deepseek_v3.py
```diff
@@ -105,28 +103,38 @@ def render_message(self, message: Message, ctx: RenderContext) -> RenderedMessag
...
                         rendered_parts.append(p["text"])
                     # ToolCallPart handled via message's tool_calls field
                 output_content = "".join(rendered_parts)
             else:
-                # String content - pass through as-is.
-                # Note: strip_thinking_from_history only works with list-based content.
-                # For stripping to work on historical messages, use structured content
-                # with ThinkingPart separated from text (as returned by parse_response).
-                output_content = content
+                # String content
```
joschu: don't bother stripping from string content (write a comment noting this)

tinker_cookbook/tests/test_renderers.py
```diff
@@ -920,6 +971,10 @@ def has_tools(msgs):
         pytest.skip(f"{renderer_name} strips thinking content, breaking roundtrip consistency")
     if has_tool_content and renderer_name in _RENDERERS_WITHOUT_TOOL_SUPPORT:
         pytest.skip(f"{renderer_name} doesn't support tool calling")
+    if renderer_name in _RENDERERS_WITH_DIFFERENT_SUPERVISED_GEN_HEADERS:
```
bzz, line 974: I guess this also excludes it from testing consistency between `build_generation_prompt` and `parse_response` which is now the issue in https://github.com/thinking-machines-lab/tinker-cookbook/issues/264#issuecomment-3703794561

PR #269, opened 2025-12-31
rollouts.py accepts logs from environment.step()

josancamon19:
allows for easier debugging/learning if you want to look at the rollouts through different training steps (specially useful in multi-turn examples), includes an implementation for guess_number/env.py

```python
step_result = StepResult(
    next_observation=self._obs,
    next_stop_condition=self.stop_condition,
    episode_done=episode_done,
    reward=reward,
    logs={
        "guess": action_content,
        "feedback": user_turn["content"],
        "target": self.gold_answer,
    },
)
```

CleanShot 2025-12-30 at 17 12 28


joschu:
This is a good addition! Made a couple minor adjustments, will merge after tests pass.

PR #276, opened 2026-01-01
ifBench rlvr instruction following ai2 

josancamon19:
https://github.com/allenai/IFBench implementation following the recipe standards. Thought it'd be interesting to add to the repo.

### DIFF
--- a/tinker_cookbook/recipes/if_rl/env.py
+++ b/tinker_cookbook/recipes/if_rl/env.py
@@ -96,9 +96,7 @@ class IfBenchEnv(ProblemEnv):
                 "all_loose": float(scores.binary_loose),
                 "num_instructions": len(results),
                 **{
-                    f"instr/{r.instruction_id}": float(
-                        r.loose_pass if use_loose else r.strict_pass
-                    )
+                    f"instr/{r.instruction_id}": float(r.loose_pass if use_loose else r.strict_pass)
                     for r in results
                 },
             },


tinker_cookbook/recipes/if_rl/ifbench/instructions.py
```diff
@@ -0,0 +1,2419 @@
...
+            "Paris",
+            "Vienna",
+            "Bratislava",
+            "Budapest",
+            "Vaduz",
+            "Chisinau",
+            "Bern",
+            "Ljubljana",
+            "Zagreb",
+        ]
```
joschu: looks like this is missing a return value

joschu:
Could you share a bit more in the README about what the model learns over the course of training? E.g.

- how the completions change on some test prompts
- how much the reward goes up

I'd assume that the Qwen instruct model is already decent at generic instruction following, so I'm wondering what this dataset teaches.

tinker_cookbook/recipes/if_rl/ifbench/instructions.py
josancamon19: not sure what do you mean, the return value is below, https://github.com/allenai/IFBench/blob/c6767a19bd82ac0536cab950f2f8f6bcc6fabe7c/instructions.py#L1690

joschu: d'oh :)

josancamon19:
this is mostly a hillclimb on ifbench, trained qwen3-4B for 334 steps to saturation, from 28% to 63% on ifbench, ifEval increased by ~1%, was expecting more (constraints are very much OOD from each other).

to clarify, this is not the setup they used in the paper, 
- [this](https://huggingface.co/datasets/allenai/IF_multi_constraints_upto5) is the actual dataset they use for rlvr, it would just be a much larger endeavor, the results reported are on 1500-2000 steps, about 6x more steps, (and many more samples, rollouts_count=16). which should increase 8% on ifEval, and 20% on ifBench.
- the 2nd setup was with a base model (same data) including the chat template they used (I think this if trained on ifBench, should yield significant gains on ifEval pretty fast), also max_tokens is set here to 10k (4x).
- another option would be to make a split on ifBench 0.9/0.1, and keep track of the results in the held out set.

let me know if you prefer any of those vs this setup.

with some further hp tuning and train setup / reward type ablations, surely it can yield better results. Had better results before with clipping and other setings.

plot below is the reward same training run, just the code is not syncing wandb logs yet but starting from scratch (should prob be fixed).

CleanShot 2026-01-02 at 22 28 41
CleanShot 2026-01-02 at 23 02 59

</br>

---

below are some examples of the improvements in ifeval for Qwen3-4B-Instruct

**Note:** as expected on the paper, one consequence of this is the model starts paying more attention to the constraints than instructions.

https://github.com/user-attachments/assets/63514771-343b-40dc-8115-e6e8ef306e46

josancamon19:
here are the jsonl results for both evals (base + best checkpoint) https://drive.google.com/drive/folders/19b4-ugTS2b19YKIhKddPcixVym3f4WZ3?usp=sharing

josancamon19:
minor follow up here @joschu, we can also close this PR if it doesn't add much

joschu:
@josancamon19 Thanks! Sorry for the delay. Wanted this for an experiment, and remembered it was unmerged.

PR #278, opened 2026-01-03
Support modal for sandboxed code execution

tyler-griggs:
## Summary

  - Add Modal as an alternative sandbox backend for code execution (alongside the existing SandboxFusion)
  - Refactor sandboxed execution code into the `sandbox/` directory, add Modal support there
  - Update the `code_rl` recipe to optionally use either SandboxFusion or Modal

## Modal
The `ModalSandbox` class is a pretty thin wrapper around the Modal API to help users initialize sandboxes and submit code execution requests to them. One sandbox is not enough to handle the load efficiently during RL, so I added a `ModalSandboxPool` that round robins requests across a pool of sandboxes.

Many coding datasets (LCB, BigCodeBench, etc.) utilize multiple files, so I added a `run_in_workdir` which handles adding files to the sandbox and cleaning them up. 

## SandboxFusion refactor
Pulled the SandboxFusion-specific logic out of `code_grading.py` into the `execution/` directory. Almost all of the code in `sandboxfusion.py` is copied over from `code_grading.py`.

### SandboxFusion
- [x] Run `code_rl` recipe with sandbox fusion to ensure no regression from before refactor

### Modal

- [x] Run `code_rl` recipe and compare to sandbox fusion results.

### DIFF
--- a/tinker_cookbook/sandbox/__init__.py
+++ b/tinker_cookbook/sandbox/__init__.py
@@ -20,11 +20,3 @@ __all__ = [
     "SandboxBackend",
     "SandboxFusionClient",
 ]
-
-
-# ModalSandbox and ModalSandboxPool are lazily imported to avoid requiring modal as a dependency
-def get_modal_sandbox_pool(**kwargs):
-    """Factory to create a ModalSandboxPool without importing modal at module load."""
-    from tinker_cookbook.sandbox.modal_sandbox import ModalSandboxPool
-
-    return ModalSandboxPool(**kwargs)


tinker_cookbook/sandbox/modal_sandbox.py
```diff
@@ -0,0 +1,181 @@
...
+            files: Files to write {filename: content}
+            command: Command and arguments (e.g., ["python", "run.py"])
+            timeout: Execution timeout in seconds
+
+        Returns:
+            Tuple of (exit_code, stdout, stderr)
+        """
+        sandbox = await self._queue.get()
+        try:
+            return await asyncio.to_thread(sandbox.run_in_workdir, files, command, timeout)
```
saltzm, line 179: You could avoid the need for `to_thread` by using the async APIs for Sandbox functions - will comment in `run_in_workdir` on what that'd look like

tinker_cookbook/sandbox/modal_sandbox.py
```diff
@@ -0,0 +1,181 @@
...
+            files: Files to write {filename: content}
+            command: Command and arguments (e.g., ["python", "run.py"])
+            timeout: Execution timeout in seconds
+
+        Returns:
+            Tuple of (exit_code, stdout, stderr)
+        """
+        workdir = f"/workspace/{uuid.uuid4().hex[:12]}"
+        try:
+            self._sandbox.exec("mkdir", "-p", workdir).wait()
```
saltzm, line 115: Any of the Sandbox functions that do blocking IO have an async variant, accessed using `.aio` like so:
```
process = await self._sandbox.exec.aio("mkdir", "-p", workdir)
process.wait()
```

If you use these everywhere you can make this function `async` and avoid the need for `to_thread`.

For a good example of an integration that does this, check out the [harbor repo](https://github.com/laude-institute/harbor/blob/main/src/harbor/environments/modal.py). They have examples of async uses of `exec` and `Sandbox.open`

saltzm: In addition, for robustness, you may want to check the return code returned from `wait()` and make sure it's `0`

tyler-griggs: Thanks, will update this!

rolyatmax:
Hey @tyler-griggs! Another Modal Sandboxes engineer here. This is awesome!

Two thoughts that might be useful:

1. By default, Modal Sandboxes are terminated after 5 minutes. Depending on how long you expect to use the sandbox pool, it may be a good idea to increase the sandbox's lifetime by configuring it with a longer [timeout](https://modal.com/docs/guide/sandboxes#timeouts). (e.g. `modal.Sandbox.create(..., timeout=600)`)

2. If you decide to use longer Sandbox timeouts, you may want to avoid running unused sandboxes after you're done with the pool either by explicitly [terminating](https://modal.com/docs/reference/modal.Sandbox#terminate) each sandbox when you're done with the pool or by configuring your sandboxes with an [idle timeout](https://modal.com/docs/guide/sandboxes#idle-timeouts). (e.g. `modal.Sandbox.create(..., timeout=1800, idle_timeout=180)`)

tyler-griggs:
Oops I missed this, thank you for the input! I'll open a PR to add configurable timeouts.

PR #279, opened 2026-01-03
Fix Kimi K2 and Qwen3 tool declaration rendering to match HuggingFace

TonyLianLong:
Fix Kimi K2 and Qwen3 tool declaration rendering to match HuggingFace

This commit fixes tool declaration formatting in Kimi K2 and Qwen3 renderers
to produce identical output to HuggingFace's chat templates:

**Kimi K2 changes:**
- Use `sort_keys=True` in tool JSON serialization (matches HF custom template)
- Improve _ensure_system_message to handle tool_declare messages correctly
- Fix import ordering

**Qwen3 changes:**
- Fix system prompt and tools text concatenation with proper separator

**Testing:**
- Add comprehensive test suite for both renderers (`test_kimi_k2_tool_declaration.py`, `test_qwen3_tool_declaration.py`)
- Tests verify token-level and string-level matching with HF
- Update test_renderers.py to extract tools from tool_declare messages

### DIFF
--- a/tinker_cookbook/tests/test_kimi_k2_tool_declaration.py
+++ b/tinker_cookbook/tests/test_kimi_k2_tool_declaration.py
@@ -6,6 +6,7 @@ import pytest
 from transformers import AutoTokenizer
 
 from tinker_cookbook.renderers import get_renderer
+from tinker_cookbook.renderers.base import ToolSpec, Message, ensure_text
 from tinker_cookbook.tokenizer_utils import get_tokenizer
 
 
@@ -64,7 +65,7 @@ def test_tool_declaration_no_duplicate_system():
     tokenizer = get_tokenizer("moonshotai/Kimi-K2-Thinking")
     renderer = get_renderer("kimi_k2", tokenizer)
 
-    tools = [
+    tools: list[ToolSpec] = [
         {"name": "test", "description": "Test", "parameters": {"type": "object", "properties": {}}}
     ]
     prefix = renderer.create_conversation_prefix_with_tools(tools, "")
@@ -87,7 +88,7 @@ def test_tool_json_keys_are_sorted():
     tokenizer = get_tokenizer("moonshotai/Kimi-K2-Thinking")
     renderer = get_renderer("kimi_k2", tokenizer)
 
-    tools = [
+    tools: list[ToolSpec] = [
         {
             "name": "get_weather",
             "description": "Get weather",
@@ -103,7 +104,7 @@ def test_tool_json_keys_are_sorted():
     ]
 
     messages = renderer.create_conversation_prefix_with_tools(tools, "")
-    tool_declare_content = messages[0]["content"]
+    tool_declare_content = ensure_text(messages[0]["content"])
 
     # Parse the JSON to check key ordering
     tools_parsed = json.loads(tool_declare_content)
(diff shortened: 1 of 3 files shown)


tinker_cookbook/tests/test_renderers.py
```diff
@@ -550,12 +561,22 @@ def test_supervised_example_against_hf_chat_templates(
...
+    # Extract tools from tool_declare messages and filter them out when converting to OpenAI format
+    tools_for_hf = None
+    hf_convo = []
+    for m in convo:
+        if m["role"] == "tool_declare":
+            # Parse the JSON content to extract tools for HF
+            tools_for_hf = json.loads(m["content"])
+        else:
+            openai_msg = cookbook_renderer.to_openai_message(m)
+            hf_convo.append(openai_msg)
```
joschu, line 577: this isn't being used now, right? can you add a conversation example (in the list of conversations that we test) that does include the tool declaration?

tinker_cookbook/tests/test_kimi_k2_tool_declaration.py
```diff
@@ -0,0 +1,260 @@
+"""Tests for Kimi K2 tool declaration rendering."""
```
joschu, line 1: This is good for checking specific details, but it'd be nice (and more concise) to just have an equality check that compares the tinker-cookbook Renderer.build_generation_prompt to apply_chat_template.

joschu (approved):

tinker_cookbook/tests/test_renderers.py
TonyLianLong: tools are declared by passing in `tools` kwarg in `apply_chat_template` (added in line 579). This is because while our renderer implements tool declaration as a message, hf implements it as an additional kwarg. The for loop extracts the tool declare message and passes the declaration to `apply_chat_template`, and discards the message from `hf_convo` (since hf doesn't take the declaration as a message).

tinker_cookbook/tests/test_kimi_k2_tool_declaration.py
TonyLianLong: Yeah we could probably remove a few tests (e.g., the json order one), since it would be also cause the string equality tests to fail, but I kind of want to keep this file, since it captures issues such as double system prompts exposed by bugs in `create_conversation_prefix_with_tools`, which the original `test_renderers.py` didn't catch. I guess we could prune some test in a followup PR :) .

Note that a single equality test is probably not enough. The duplicate system prompts would not be caught by the equality test since the duplicate was introduced to messages (which are used both by our renderer and the hf `apply_chat_template`).

tinker_cookbook/tests/test_renderers.py
joschu: yes I understand that the logic here is correct, I just was wondering if it's being exercised. I think to exercise it we'd need _SUPERVISED_CONVERSATIONS / _GENERATION_CONVERSATIONS to contain a message with tool declaration

TonyLianLong: I see! This is in fact a bit complicated: `tool_declare` is only added by Kimi K2 renderer:

https://github.com/thinking-machines-lab/tinker-cookbook/blob/81eeeb3cd08e3c62181a93a4ac7e387f9613b067/tinker_cookbook/renderers/kimi_k2.py#L415

Adding `tool_declare` might break the test for other models. On the other hand, we already have tests that use `tool_declare` message for Kimi K2:

https://github.com/thinking-machines-lab/tinker-cookbook/blob/2c39599c18cec2cc8e144b83102e4f4a458e1ae3/tinker_cookbook/tests/test_kimi_k2_tool_declaration.py#L134

This test compares the tokens from our renderer and the tokens from hf `apply_chat_template`. I guess it should be fine for now? I could make another PR after figuring out a plan.

PR #284, opened 2026-01-05
Fix DeepSeekV3ThinkingRenderer.parse_response to handle prefilled <think>

### LINKED ISSUE #283: deepseek renderer DeepSeekV3ThinkingRenderer doesn't parse reasoning
Model response does not get the `<thinking>` prefix added to the header of the first message (the way prefill [gets it as suffix](https://github.com/thinking-machines-lab/tinker-cookbook/pull/267/changes#diff-3661508706e847a9c7589e599f40160ac12bfa6e36b42ab7a5b5b1705e301630R404)) so [.parse_response()](https://github.com/thinking-machines-lab/tinker-cookbook/pull/267/changes#diff-3661508706e847a9c7589e599f40160ac12bfa6e36b42ab7a5b5b1705e301630L267) doesn't include ThinkingParts

Way to reproduce reported in https://github.com/thinking-machines-lab/tinker-cookbook/issues/264#issuecomment-3703801300

bledden:
Fixes #283.

## Summary

When `build_generation_prompt` adds `<think>` as a prefill to trigger thinking mode, the model's response tokens don't include this prefix. The `parse_response` method was failing to extract thinking content because it expected `<think>` to be in the response.

This fix prepends `<think>` tokens before parsing so `parse_think_blocks` can correctly identify and extract thinking content into `ThinkingPart`.

## Changes

- Added `parse_response` override in `DeepSeekV3ThinkingRenderer` that prepends the `<think>` tokens before delegating to the base class
- Added test case that specifically covers the bug scenario (response without `<think>` prefix)

## Test plan

- [x] Added `test_deepseek_parse_response_without_think_prefix` test
- [ ] Existing DeepSeek tests should continue to pass

### DIFF
--- a/tinker_cookbook/renderers/deepseek_v3.py
+++ b/tinker_cookbook/renderers/deepseek_v3.py
@@ -428,6 +428,38 @@ class DeepSeekV3ThinkingRenderer(_DeepSeekV3BaseRenderer):
         think_prefill = "<think>" + (prefill or "")
         return super().build_generation_prompt(messages, role, think_prefill)
 
+    def parse_response(self, response: list[int]) -> tuple[Message, bool]:
+        """Parse response, prepending <think> if needed to account for prefill.
+
+        Since build_generation_prompt adds <think> as a prefill, the model's response
+        tokens don't include this prefix. We need to prepend it before parsing so that
+        parse_think_blocks can correctly identify and extract the thinking content.
+
+        Without this, the response would look like:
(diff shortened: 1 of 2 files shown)


bzz:
Nice! I've also tried to add a more generic round-trip test in #285

bledden:
Closing in favor of #285 which has a simpler string-level fix and more comprehensive test coverage across all thinking renderers.

PR #285, opened 2026-01-05
Fix DeepSeekV3ThinkingRenderer to parse thinking traces

### LINKED ISSUE #283: deepseek renderer DeepSeekV3ThinkingRenderer doesn't parse reasoning

bzz:
Closes #283. It includes a new renderer test for all thinking models that checks correspondence `build_supervised_example` and `parse_response` though a round-trip like

```py
continuation_tokens = .render_message([expected_message])[prefill_len:]
.build_supervised_example([user_message, .parse_response(continuation_tokens)]) \
   == .build_generation_prompt([user_message]) + continuation_tokens
```

test plan:
 - pytest tinker_cookbook/tests/test_renderer_parsing.py

### DIFF
--- a/tinker_cookbook/renderers/deepseek_v3.py
+++ b/tinker_cookbook/renderers/deepseek_v3.py
@@ -433,16 +433,17 @@ class DeepSeekV3ThinkingRenderer(_DeepSeekV3BaseRenderer):
 
         When sampling with build_generation_prompt, the <think> tag is part of the
         prefill and not included in the sampled tokens. The response will be
-        "reasoning</think>answer" - we prepend <think> so parse_think_blocks works.
+        "reasoning</think>answer" so we prepend <think> if necessary.
         """
-        content = self.tokenizer.decode(response)
-        # prefill case that satisfies both tests:
-        # - test_deepseek_parse_response_no_thinking_returns_string()
-        # - test_deepseek_parse_response_multiple_think_blocks()
-        if "</think>" in content:
-            content = "<think>" + content
-            response = self.tokenizer.encode(content, add_special_tokens=False)
-        return super().parse_response(response)  # pyright: ignore[reportCallIssue]
+        think_prefix_token: int = self.tokenizer.convert_tokens_to_ids("<think>")  # type: ignore[assignment]
+        think_suffix_token: int = self.tokenizer.convert_tokens_to_ids("</think>")  # type: ignore[assignment]
+
+        # Only prepend <think> if the response doesn't already start with it and contains </think>
+        starts_with_think = len(response) > 0 and response[0] == think_prefix_token
+        if not starts_with_think and think_suffix_token in response:
+            response = [think_prefix_token] + response
+
+        return super().parse_response(response)
 
 
 class DeepSeekV3DisableThinkingRenderer(_DeepSeekV3BaseRenderer):


joschu:
Good catch, thanks. But the current fix affects both the non-thinking and thinking renderer. How about just making it so the thinking renderer always prefixes with `<think>`, since we're always prefilling with that?

joschu:
Also please note the pyright error.

bzz:
> current fix affects both the non-thinking and thinking renderer.

true, but also the condition "</think>" in content is never true for `DisableThinking` responses so it should not affect the runtime behavior (I've added a case to double-check that).

> How about just making it so the thinking renderer always prefixes with <think>, since we're always prefilling with that?

~sounds good~ thinking more about it - that would break the round-trip test of `.parse_response` called on full supervised response (logic highlighted in PR description). 
To keep that property and localize the fix in DeepSeekV3ThinkingRenderer we effectively will need something like the [fix from #284](https://github.com/thinking-machines-lab/tinker-cookbook/pull/284/changes#diff-3661508706e847a9c7589e599f40160ac12bfa6e36b42ab7a5b5b1705e301630R431).

(the other way of moving string-level fix away from base renderer I can see ATM will have a price decoding/encoding the message in `DeepSeekV3ThinkingRenderer.parse_response()`)

Let me know which one you prefer - I'm happy to roll back the fix and merge only the tests, or cherry-pick @bledden fix from #284 in here!

TonyLianLong:
Thanks for proposing the fix! Some questions from my end:

> thinking more about it - that would break the round-trip test of .parse_response called on full supervised response (logic highlighted in PR description).

Would you mind explaining a bit why always adding `<think>` would break this test?

Specifically, what would go wrong if we do this in `parse_response`:
```python
        if "DisableThinking" not in self.__class__.__name__:
            content = "<think>" + content
```


Also about string-based parsing: since we already use regex here

https://github.com/thinking-machines-lab/tinker-cookbook/blob/c6b38cead98012872d0b029e4514f860bbc44fca/tinker_cookbook/renderers/base.py#L447

it might be ok to manipulate in the string level instead of token level.

bzz:
> > thinking more about it - that would break the round-trip test of .parse_response called on full supervised response
> 
> Would you mind explaining a bit why always adding `<think>` would break this test?

Not at all, thanks for asking! Apparently, when writing that I got confused by considering to validate a stronger assumption of parsing results of `.build_supervised_example()` 
```py
    supervised, weights = renderer.build_supervised_example(full_convo)
    prompt, response = _split_by_weights(supervised.to_ints(), weights)
    assert response == continuation_tokens
```

that 
* never made it into this test 
* seems to be [handled elsewhere](https://github.com/thinking-machines-lab/tinker-cookbook/pull/267/changes/BASE..5e7ed5a69cd7ac603c601ce9ec073d2897f5b155#diff-2de31a766f638ca1f0376b57c416cccbc0eaeb14eff61d1311e4e64fb463ddebR921)
* would be expected to fail with the current change anyways  🤦 

Adding `<think>` unconditionally would break a couple of tests e.g. [test_renderer_parsing.py::test_deepseek_parse_response_no_thinking_returns_string](https://github.com/thinking-machines-lab/tinker-cookbook/blob/75e21aaa6f7dbe1ced330965768837373b9a9c9c/tinker_cookbook/tests/test_renderer_parsing.py#L253) and `test_deepseek_parse_response_multiple_think_blocks()`.

I've moved the logic to `DeepSeekV3ThinkingRenderer.parse_response()` for a price of extra encoding/decoding , please let me know if that works.

TonyLianLong:
> I've moved the logic to DeepSeekV3ThinkingRenderer.parse_response() for a price of extra encoding/decoding , please let me know if that works.

Since `parse_response` is token level, and we know `<think>` and `</think>` are both one token. Can we directly scan for `</think>` token and add `<think>` to the front without decoding/encoding the full string?

@joschu what do you think?

bledden:
@TonyLianLong something like https://github.com/thinking-machines-lab/tinker-cookbook/pull/284 by chance?

Specifically:
```
def parse_response(self, response: list[int]) -> tuple[Message, bool]:
    think_prefix_tokens = self.tokenizer.encode("<think>", add_special_tokens=False)
    response_list = list(response)

    # Only prepend <think> if the response doesn't already start with it
    starts_with_think = (
        len(response_list) >= len(think_prefix_tokens)
        and response_list[: len(think_prefix_tokens)] == think_prefix_tokens
    )
    if not starts_with_think:
        response_list = think_prefix_tokens + response_list

    return super().parse_response(response_list)
```

TonyLianLong:
Yes this is what I meant. I believe `tokenizer.convert_tokens_to_ids` would work so we don't need to call `tokenizer.encode`, as `<think>` is a single token.

bzz:
Thanks for a prompt reviews, folks! 🙇 

We'll still have to check for closing tag as well if we are to stick with the [test_renderer_parsing.py::test_deepseek_parse_response_no_thinking_returns_string()](https://github.com/thinking-machines-lab/tinker-cookbook/blob/75e21aaa6f7dbe1ced330965768837373b9a9c9c/tinker_cookbook/tests/test_renderer_parsing.py#L253)

TonyLianLong:
Makes sense to check for the closing tag! The current implementation seems to pass all the tests and looks good to me.

I think this should fix everything we found with the deepseek renderer? Any missing pieces that are still not fixed?

PR #286, opened 2026-01-05
fix: populate metrics from OptimStepResponse

### LINKED ISSUE #275: OptimStepResponse.metrics: should it be populated?
We were hoping to monitor gradient norms during SFT while debugging an issue.  OptimStepResponse.metrics exists in the api and docs, but it's unclear what it should contain. In our runs optim_step().result().metrics seems to always be empty.  Should it be populated? Are grad norms exposed anywhere?

nathon-lee:
fix(supervised): populate metrics from OptimStepResponse

Previously, the result of `optim_step_future` was awaited but ignored in `finish_batch`, causing optimization metrics (like gradient norms) to be dropped.

This commit modifies `tinker_cookbook/supervised/train.py` to:
- Capture the `OptimStepResponse` result from the future.
- Merge its `metrics` field (if present) into the current step's metrics dictionary.

Fixes #275

### DIFF
--- a/tinker_cookbook/rl/train.py
+++ b/tinker_cookbook/rl/train.py
@@ -197,6 +197,7 @@ async def train_step(
 
     adam_params = tinker.AdamParams(learning_rate=learning_rate, beta1=0.9, beta2=0.95, eps=1e-8)
     training_logprobs_D: list[torch.Tensor] = []
+    optim_result: tinker.OptimStepResponse | None = None
 
     # Enqueue first batch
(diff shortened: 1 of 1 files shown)


tyler-griggs:
Thanks @nathon-lee! I made a slight simplification since the `OptimStepResponse` is a pydantic `BaseModel`, we don't need to check whether it has the `metrics` field. Then I added the same metrics extraction to other training loops that were previously dropping them.

PR #295, opened 2026-01-08
Fix: allow streaming dataset batch skipping

Pran-Ker:
HF's shuffle is deterministic, so a skip goes through the same data as sequential.

Here is a quick test which I ran to make sure the batch skipping works correctly:

```python

from datasets import load_dataset
from tinker_cookbook.supervised.data import StreamingSupervisedDatasetFromHFDataset, conversation_to_datum
from tinker_cookbook.renderers import get_renderer, TrainOnWhat
from tinker_cookbook.tokenizer_utils import get_tokenizer

tokenizer = get_tokenizer("meta-llama/Llama-3.1-8B-Instruct")
renderer = get_renderer("llama3", tokenizer)

def map_fn(row: dict):
    text = row.get('text', '')[:100]
    messages = [{"role": "user", "content": text[:30]}, {"role": "assistant", "content": text}]
    return conversation_to_datum(messages, renderer, max_length=100, train_on_what=TrainOnWhat.ALL_ASSISTANT_MESSAGES)

# Create dataset
hf_dataset = load_dataset("wikitext", "wikitext-2-raw-v1", split="train", streaming=True)
dataset = StreamingSupervisedDatasetFromHFDataset(hf_dataset, batch_size=5, length=20, map_fn=map_fn)

# Test forward iteration
dataset.get_batch(0)
dataset.get_batch(1)
print("✓ Forward iteration works")

# Test forward skipping
dataset.get_batch(5)
print("✓ Forward skipping works")

# Test backward seek rejection
try:
    dataset.get_batch(3)
    print("✗ Backward seek should have failed")
except ValueError:
    print("✓ Backward seek properly rejected")

print("\n✓ ALL TESTS PASSED")

```

And the result:

CleanShot 2026-01-08 at 15 43 44@2x

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 **/__pycache__
 .env
 .env.*
+uv.lock
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,7 +61,7 @@ trackio = [
     "trackio<1.0.0",
 ]
 verifiers = [
-    "verifiers>=0.1.8.post0",
+    "verifiers>=0.1.9,<0.1.10",
     "openai",
 ]
 all = [
--- a/tinker_cookbook/distillation/train_on_policy.py
+++ b/tinker_cookbook/distillation/train_on_policy.py
@@ -142,8 +142,10 @@ class Config:
     kl_penalty_coef: float = 1.0
     kl_discount_factor: float = 0.0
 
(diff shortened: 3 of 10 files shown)


tyler-griggs:
Thanks @Pran-Ker! Tested it, looks good.

PR #300, opened 2026-01-14
Modal Sandboxes - maintain warm pool

ekzhang:
Changes:

- Maintain the pool over time, removing old sandboxes and adding fresh ones after a configurable duration (lower than the actual timeout) — this also means we don't reuse sandboxes
- Configure a container timeout for the sandboxes themselves, not just the exec (default is 600 seconds = 10 minutes, previously not configured)
- Add a `terminate()` method to ModalSandboxPool so people can kill the pool when they're done.

### DIFF
--- a/tinker_cookbook/sandbox/README.md
+++ b/tinker_cookbook/sandbox/README.md
@@ -47,7 +47,7 @@ Example usage:
 from tinker_cookbook.sandbox.modal_sandbox import ModalSandbox, ModalSandboxPool
 
 # Single sandbox
-sandbox = ModalSandbox()
+sandbox = await ModalSandbox.create()
 exit_code, stdout, stderr = await sandbox.run_in_workdir(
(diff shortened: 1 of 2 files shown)


tinker_cookbook/sandbox/README.md
```diff
@@ -47,7 +47,7 @@ Example usage:
 from tinker_cookbook.sandbox.modal_sandbox import ModalSandbox, ModalSandboxPool
 
 # Single sandbox
-sandbox = ModalSandbox()
+sandbox = await ModalSandbox.create()
```
ekzhang, line 50: Made this async since I didn't want to accidentally block the event loop while we're creating new `ModalSandbox` instances in this warm pool maintenance setup

ekzhang:
CI is failing due to unrelated type-check now (edit: fixed in #301)

tyler-griggs (approved):

PR #302, opened 2026-01-15
Adhere to modal sandbox limits (total instances, creation rates)

tyler-griggs:
### PR Summary

Update warm pool handling to adhere to sandbox count limits and creation rate limits.

I hit a couple hiccups that stemmed from launching RL jobs with >1k simultaneous tool calls:

1. Unbounded sandbox counts: previously if the warm pool was empty, a request would still create a new sandbox. This quickly exceeded modal account sandbox limits.  
1. Async sandbox termination and creation: after fixing the above, sandbox creation and termination were running asynchronously of each other, which allowed for the total number of active sandboxes to exceed the max pool size.
  2. Rate limit violations - Creating all sandboxes in the pool in parallel exceeded Modal's rate limit, causing "Sandbox creation rate limit exceeded" errors. 

###  New behavior:
  - _maintain_pool_step() now handles termination, waits, then creates new sandboxes. The max pool size is never exceeded.
  - MODAL_CREATION_RATE_LIMIT env var (default: 4) controls max creations per second

### DIFF
--- a/tinker_cookbook/sandbox/modal_sandbox.py
+++ b/tinker_cookbook/sandbox/modal_sandbox.py
@@ -137,8 +137,9 @@ class ModalSandbox:
         return exit_code, stdout, stderr
 
     async def terminate(self) -> None:
-        """Terminate the Modal sandbox."""
+        """Terminate the Modal sandbox and wait for it to fully shut down."""
         await self._sandbox.terminate.aio()
+        await self._sandbox.wait.aio(raise_on_termination=False)
 
 
 class ModalSandboxPool:


tinker_cookbook/sandbox/modal_sandbox.py
```diff
@@ -145,74 +147,67 @@ class ModalSandboxPool:
...
-            raise BaseExceptionGroup(f"Errors creating {len(failures)} Modal sandboxes", failures)
+        """Single iteration of pool maintenance: terminate used sandboxes, create new ones."""
+        # Batch terminate used sandboxes
+        if self._to_terminate:
+            to_terminate, self._to_terminate = self._to_terminate, []
+            await asyncio.gather(*(sb.terminate() for sb in to_terminate))
+
+        # Create new sandboxes in parallel (respecting rate limit)
+        total = self._warm_pool.qsize() + self._active_count
+        need = min(self._creation_rate_limit, self._pool_size - total)
```
ekzhang, line 201: Makes sense!

tinker_cookbook/sandbox/modal_sandbox.py
```diff
@@ -145,74 +147,67 @@ class ModalSandboxPool:
...
-                failures.append(sb)
-            else:
-                self._warm_pool.append(sb)
-        if failures:
-            raise BaseExceptionGroup(f"Errors creating {len(failures)} Modal sandboxes", failures)
+        """Single iteration of pool maintenance: terminate used sandboxes, create new ones."""
+        # Batch terminate used sandboxes
+        if self._to_terminate:
+            to_terminate, self._to_terminate = self._to_terminate, []
+            await asyncio.gather(*(sb.terminate() for sb in to_terminate))
```
ekzhang, line 197: This might be a gotcha of the Modal API but this only sends a signal, if you want to actually wait for the sandbox to finish I think it's terminate and then wait

```python
modal_sandbox_obj: modal.Sandbox  # say you have this

await modal_sandbox_obj.terminate.aio()
await modal_sandbox_obj.wait.aio()
```

ekzhang (approved):
Nice!

tinker_cookbook/sandbox/modal_sandbox.py
tyler-griggs: Ah got it, fixed. thanks!

PR #305, opened 2026-01-18
Add support for resuming W&B runs by ID

josancamon19:
This ended up being a bit trickier than I expected—especially from a UX standpoint.

The current implementation can successfully resume a run, but it should only be used when you want to continue training longer with the *same* configuration.

If you pass a new config, it won’t update in W&B (and it really should be a new run). Also, if you resume from a checkpoint that isn’t the final checkpoint, the graphs can become messy and inconsistent.

This won’t happen by default, but giving users an easy failure mode like this makes me uneasy, and it makes me question whether we should expose it as an option at all.

Any thoughts from @joschu or @tyler-griggs would be really helpful.


josancamon19:
Not finished yet, there’s still some cleanup and a few edge cases to handle, but I’d love to hear your thoughts before I get to that.

tinker_cookbook/utils/ml_log.py
```diff
@@ -219,22 +220,74 @@ def __init__(
...
+            logger.info(f"Resumed WandB run {resume_run_id}, step offset: {self.step_offset}")
+        else:
+            self.run = wandb.init(
+                project=project,
+                config=dump_config(config) if config else None,
+                dir=str(log_dir) if log_dir else None,
+                name=wandb_name,
+            )
+
+    def _get_last_step_from_run(self) -> int:
```
tyler-griggs: I think we could just remove step offset handling entirely and simplify this file quite a lot. It is reasonable for the caller to handle step offsets. Namely, the caller is responsible for checkpointing / resuming the training state, which will include the training step. See `sl_loop.py` for an example of this, where the resumed `step` is derived from the `batch_idx` of the checkpointed data loader state.

tyler-griggs:
Thanks for writing this up, @josancamon19. There's always the option of just using the env vars `WANDB_RUN_ID` and `WANDB_RESUME`. Wandb will pick these up automatically, so technically users can currently resume a wandb run by ID. This would simplify things quite a lot! 

We could even add a note (e.g., in recipes/README.md under `Resuming`) like "To resume logging to an existing W&B run, set the following environment variables: `WANDB_RUN_ID=<run_id>` and `WANDB_RESUME=must`."

josancamon19:
yep, makes a lot of sense!

josancamon19:
Hey @tyler-griggs, was adding a minor PR with docs changes, but after testing this, noticed that it doesn't work due to step unsync
```
wandb: WARNING Tried to log to step 19 that is less than the current step 37. Steps must be monotonically increasing, so this data will be ignored. See https://wandb.me/define-metric to log data out of order.
```

added minimal changes and wrapped them inside WandbLogger to fix this + readme update

josancamon19:
ping here, @tyler-griggs

YujiaBao:
Thanks for digging into this @josancamon19 — I think your instinct in the PR description ("makes me question whether we should expose it as an option at all") is actually the right call.

The step monotonicity issue you hit is real, but I think the fix belongs in the training loop's checkpoint/resume logic rather than the logger. The training loop should persist and restore the global step counter on resume — that way all loggers (JSON, W&B, Neptune, Trackio) get correct steps automatically without any special handling.

Patching it in `WandbLogger` with a step offset has a few downsides:
- It only fixes W&B, not the other loggers
- It'll double-count steps if/when the training loop properly restores the step counter from a checkpoint
- The `wandb.Api()` call at init is fragile (stale data after crashes, auth differences, network dependency)

The README addition about `WANDB_RESUME`/`WANDB_RUN_ID` env vars is useful though — that part could be split into its own tiny docs PR.

What do you think about closing this and instead fixing step tracking in the resume path of the training recipes?

PR #319, opened 2026-01-22
Add streaming parsing for the Kimi K2 renderer

nealwu:
Adds streaming parsing support to the Kimi K2 renderer to enable real-time display of model output.

  - **New streaming types** in `base.py`: `StreamingMessageHeader`, `StreamingTextDelta`, `StreamingThinkingDelta`, `MessageDelta`
  - **`Utf8TokenDecoder`**: Utility for handling tokens that split UTF-8 character boundaries
  - **`KimiK2StreamingParser`**: Stateful parser that yields incremental deltas as tokens arrive
  - **`KimiK2Renderer.parse_response_streaming()`**: New method that wraps the streaming parser

### Tests

  - `test_kimi_streaming_simple_text` - basic text
  - `test_kimi_streaming_with_thinking` - <think> blocks
  - `test_kimi_streaming_matches_batch` - consistency with batch parse
  - `test_kimi_streaming_multiple_think_blocks` - interleaved thinking
  - additional tests

### DIFF
--- a/tinker_cookbook/renderers/base.py
+++ b/tinker_cookbook/renderers/base.py
@@ -190,6 +190,10 @@ class StreamingThinkingDelta:
 MessageDelta = Union[StreamingMessageHeader, StreamingTextDelta, StreamingThinkingDelta, "Message"]
 
 
+# Unicode replacement character - indicates incomplete/invalid UTF-8 sequence
+_REPLACEMENT_CHAR = "\ufffd"
+
+
 @dataclass
 class Utf8TokenDecoder:
     """Handles incremental UTF-8 decoding from tokens.
@@ -198,7 +202,14 @@ class Utf8TokenDecoder:
     might be split across 2 tokens). This class buffers tokens until a
     valid UTF-8 string can be decoded.
 
-    Based on the pattern from hash_role2_renderer.py's Utf8TokenDecoder.
+    Detection strategy:
+    1. Try decoding all pending + new tokens
+    2. If result contains trailing U+FFFD (replacement char), it's incomplete
+    3. Scan backwards to find longest prefix without trailing replacement chars
+    4. Emit that prefix, buffer the rest
+
+    This handles tiktoken-style tokenizers that return replacement chars
+    instead of throwing exceptions for incomplete UTF-8.
     """
 
     tokenizer: "Tokenizer"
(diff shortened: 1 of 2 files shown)


tinker_cookbook/renderers/base.py
```diff
@@ -143,6 +143,141 @@ class UnparsedToolCallPart(TypedDict):
...
+
+        Returns:
+            Decoded string if complete UTF-8 sequences are available,
+            None if all tokens were buffered (incomplete sequence).
+        """
+        self._pending_tokens.extend(tokens)
+
+        # Try to decode all pending tokens (common case)
+        try:
+            text = self.tokenizer.decode(self._pending_tokens)
```
opherlieber, line 250: Assuming the tokenizer won't fail on partial UTF and return '�', when will that be fixed here, if at all? 
I think in this case we might want either:
- To always backtrack/keep the last N(e.g. 8) tokens, and if we get more tokens we can validate that the previous delta's didnt change. If they did we would need some sort of backspace/revert delta though.
- Alternatively, assuming it's always '�' for partial unicode (And no other cases have inconsistent detokinization), we can maybe just detect '�' here and and not return a delta for that and keep the tokens pending until the '�' is at least N tokens behind which means it's an actual '�' and not a partial unicode.

tinker_cookbook/renderers/base.py
```diff
@@ -143,6 +143,141 @@ class UnparsedToolCallPart(TypedDict):
...
+        # Scan backwards to find longest decodable prefix.
+        # We only need to try removing a few tokens since UTF-8 sequences are at
+        # most 4 bytes and tokens are typically 1-4 bytes each.
+        for remove in range(
+            1, min(len(self._pending_tokens), self._MAX_TRAILING_TOKENS_TO_TRY) + 1
+        ):
+            prefix = self._pending_tokens[:-remove]
+            if not prefix:
+                break
+            try:
```
opherlieber, line 267: It looks like this tries to handle it, but it assumes exception in case of partial unicode?

tinker_cookbook/renderers/kimi_k2.py
```diff
@@ -79,6 +87,255 @@ def _parse_tool_calls_section(
...
+    """
+    max_check = min(len(text), len(tag) - 1)  # -1 because full tag would be found, not buffered
+    for length in range(max_check, 0, -1):
+        if text.endswith(tag[:length]):
+            return length
+    return 0
+
+
+@dataclass
+class KimiK2StreamingParser:
```
opherlieber, line 131: Feels like a lot if not most of the code here is not kimi specific? Maybe as a follow-up PR we can move all common logic to a StreamingParser base-class and then add only model-specific overrides/properties in the subclasses

tinker_cookbook/renderers/base.py
opherlieber: Oh just saw you wrote this is a known limitation. Do we know if/how it's solved in our internal streaming stack?
This feels like it can be pretty common with emojis streaming.

nealwu: thanks for flagging! it seems like it consistently turns into the replacement char you mentioned above, so I updated the code to handle that logic and hold for more tokens when it detects the replacement char at the end. this should work better

tinker_cookbook/renderers/kimi_k2.py
nealwu: agreed, sounds good 👍

PR #324, opened 2026-01-27
Set 7 day TTL by default on checkpoints

danobi:
To help save users money.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ dependencies = [
     "pillow",
     "rich",
     "termcolor",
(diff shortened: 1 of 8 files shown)


tyler-griggs (approved):
Does `ttl_seconds=0` mean checkpoints never expire? If so, could we document that somewhere? thanks for this!

danobi:
`ttl_seconds=0` means immediately expire. `ttl_seconds=None` means never. It's documented in the API docs: https://tinker-docs.thinkingmachines.ai/api-reference/trainingclient#save_state

PR #333, opened 2026-01-31
fix: pass trust_remote_code=True for all tokenizers

### LINKED ISSUE #327: trust_remote_code=True needed in tokenizer_utils.py for models with custom tokenizers (e.g., Kimi-K2-Thinking)
When running evaluations on models that require custom tokenizer code (like moonshotai/Kimi-K2-Thinking), the evaluation fails because tokenizer_utils.py doesn't pass trust_remote_code=True to AutoTokenizer.from_pretrained().          
                                                                                   
**Steps to Reproduce**                                                               
                                                                                   
python -m tinker_cookbook.eval.run_inspect_evals \                               
      model_name=moonshotai/Kimi-K2-Thinking \                                     
      tasks=my_eval.py \                                                           
      renderer_name=qwen3 \                                                        
      log_dir=logs/eval                                                            
                                                                                   
**Error Message**                                                                    
                                                                                   
ValueError: The repository moonshotai/Kimi-K2-Thinking contains custom code which must be executed to correctly load the model. You can inspect the repository content at https://hf.co/moonshotai/Kimi-K2-Thinking. Please pass the argument `trust_remote_code=True` to allow custom code to be run.
                                                                                   
**Root Cause**                                                                       
                                                                                   
In tinker_cookbook/tokenizer_utils.py, line 31:                                  
                                                                                   
```
@cache                                                                           
def get_tokenizer(model_name: str) -> Tokenizer:                                 
      from transformers.models.auto.tokenization_auto import AutoTokenizer         
                                                                                   
      # Avoid gating of Llama 3 models:                                            
      if model_name.startswith("meta-llama/Llama-3"):                              
          model_name = "thinkingmachineslabinc/meta-llama-3-tokenizer"             
                                                                                   
      return AutoTokenizer.from_pretrained(model_name, use_fast=True)  # Missing trust_remote_code   
```

bledden:
## Summary
- Pass `trust_remote_code=True` unconditionally to `AutoTokenizer.from_pretrained()` instead of model-specific checks
- Models with custom tokenizer implementations (e.g. `moonshotai/Kimi-K2-Thinking`) require this flag to load

This is a safer general approach since the library already trusts model code for training. Adding per-model checks doesn't scale as more models with custom tokenizers are added.

Fixes #327

### DIFF
--- a/tinker_cookbook/tests/test_tokenizer_utils.py
+++ b/tinker_cookbook/tests/test_tokenizer_utils.py
@@ -1,0 +1,70 @@
+"""Tests for tokenizer_utils.get_tokenizer."""
+
+from unittest.mock import MagicMock, patch
+
+
+def _fresh_get_tokenizer():
+    """Import get_tokenizer with a cleared cache to avoid cross-test pollution."""
+    import importlib
+
+    import tinker_cookbook.tokenizer_utils as mod
+
+    importlib.reload(mod)
+    return mod.get_tokenizer
+
+
+@patch("tinker_cookbook.tokenizer_utils.AutoTokenizer", create=True)
+def test_trust_remote_code_always_passed(mock_auto):
+    """trust_remote_code=True should be passed for every model, not just specific ones."""
+    mock_auto = MagicMock()
+    with patch("transformers.models.auto.tokenization_auto.AutoTokenizer", mock_auto):
+        get_tokenizer = _fresh_get_tokenizer()
+        get_tokenizer("some-org/some-model")
+
+    mock_auto.from_pretrained.assert_called_once_with(
+        "some-org/some-model", use_fast=True, trust_remote_code=True
+    )
+
+
+@patch("tinker_cookbook.tokenizer_utils.AutoTokenizer", create=True)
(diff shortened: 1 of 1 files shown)


YujiaBao (changes requested):
Thanks for the PR! We've decided not to make `trust_remote_code=True` the default for all tokenizers — even though we trust model code for training, we'd prefer to keep this opt-in for security reasons.

HuggingFace already supports the `HF_TRUST_REMOTE_CODE` environment variable ([docs](https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables)). Setting `HF_TRUST_REMOTE_CODE=1` makes all `from_pretrained` calls behave as if `trust_remote_code=True` was passed, so users working with new custom-tokenizer models can opt in without needing a code change.

Our current approach of adding per-model `trust_remote_code=True` (as we do for Kimi-K2-Thinking and Kimi-K2.5) is intentional — it keeps the default secure while still supporting known models.

Closing in favor of keeping the current behavior. If you run into issues with a specific model, please open an issue and we can add it to the allowlist.

PR #334, opened 2026-01-31
fix: normalize SFT weights per-example for consistent gradient magnitudes

### LINKED ISSUE #271: Unnormalized token-sum loss is a surprising default for SFT
I think many users will not expect (and be surprised by) the default SFT implementation not normalizing loss by total number of target tokens.

At least, I was, and spent some time debugging training instability that seems to be ultimately due to variable length batches combined with token sum loss.

I have a commit in my fork that adds an option to normalize the token-weighted loss (i.e., mean loss over target tokens), but I'm not sure what you'd want the implementation to look like, and since it's a simple change, it might just be easier for you to reimplement it in the style you prefer than try to coordinate with me. :)

My change also preserves the default behavior, but I think it might actually be good to switch the default for SFT. Even though a change like that would be somewhat disruptive, it might be worth the short term pain to eliminate a surprising default.

bledden:
## Summary
- Normalize per-token weights in `datum_from_model_input_weights` to sum to 1.0 per example
- Produces a token-mean loss instead of a token-sum loss, making gradient magnitudes consistent across batches with variable-length sequences
- Adds `normalize_weights` parameter (default `True`) to opt out if needed via `normalize_weights=False`

## Motivation

The current token-sum loss causes batches with more target tokens to produce proportionally larger gradients. With variable-length sequences (common in chat/instruction datasets), this effectively makes the learning rate fluctuate with per-batch token count. As @joschu noted, this is mostly masked by Adam but can cause instability with high-variance sequence length datasets.

## Breaking change note

This PR defaults `normalize_weights=True`, which silently changes loss behavior for all existing callers of `datum_from_model_input_weights` (SFT, preference, DPO, VLM classifier). The old token-sum behavior is recoverable by passing `normalize_weights=False`. This aligns with the direction discussed in #271, but if a more gradual rollout is preferred — e.g., threading the option through `ChatDatasetBuilderCommonConfig` so users can toggle it per-dataset, or defaulting to `False` for backward compatibility — happy to adjust.

## Test plan
- [x] Verify weights sum to 1.0 per example after normalization
- [x] Verify `normalize_weights=False` recovers old behavior
- [x] Verify all-zero weights are handled safely (no division by zero)
- [x] Verify single non-zero weight edge case
- [x] Verify target tokens are correctly left-shifted
- [x] Verify max_length truncation interacts correctly with normalization
- [ ] Verify existing recipe smoke tests pass

Fixes #271

### DIFF
--- a/tinker_cookbook/supervised/common.py
+++ b/tinker_cookbook/supervised/common.py
@@ -1,5 +1,4 @@
 import logging
-from collections.abc import Callable
 from typing import Literal
 
 import tinker
@@ -7,9 +6,7 @@ import torch
 
 from tinker_cookbook.exceptions import DataValidationError
 
-WeightNormalization = Literal["none", "mean"] | Callable[[torch.Tensor], torch.Tensor]
-
-_logged_normalization_modes: set[str] = set()
+_logged_reduction_modes: set[str] = set()
 
 logger = logging.getLogger(__name__)
 
@@ -108,7 +105,7 @@ def datum_from_model_input_weights(
     model_input: tinker.ModelInput,
     weights: torch.Tensor,
     max_length: int | None = None,
-    normalization: WeightNormalization = "none",
+    reduction: Literal["none", "mean"] = "none",
 ) -> tinker.Datum:
     """Create a training Datum from a ModelInput and per-token weights tensor.
 
@@ -124,12 +121,11 @@ def datum_from_model_input_weights(
             aligned with ``model_input.length``.
         max_length (int | None): Optional maximum sequence length.  If
             provided, the input is truncated from the right to fit.
-        normalization (WeightNormalization): How to normalize per-token loss
+        reduction (Literal["none", "mean"]): How to reduce per-token loss
             weights after slicing.  ``"none"`` preserves raw weights
             (token-sum loss).  ``"mean"`` normalizes weights to sum to 1.0
             per example (token-mean loss), making gradient magnitudes
-            consistent across variable-length sequences.  A callable
-            ``(Tensor) -> Tensor`` applies a custom transformation.
+            consistent across variable-length sequences.
             Defaults to ``"none"``.
 
     Returns:
@@ -143,7 +139,7 @@ def datum_from_model_input_weights(
 
         datum = datum_from_model_input_weights(model_input, weights, max_length=2048)
         datum = datum_from_model_input_weights(
-            model_input, weights, max_length=2048, normalization="mean",
+            model_input, weights, max_length=2048, reduction="mean",
         )
     """
 
@@ -183,21 +179,16 @@ def datum_from_model_input_weights(
     )
     weights = weights[1 : len(target_tokens) + 1]
 
-    # Apply weight normalization
-    if normalization == "mean":
+    # Apply weight reduction
+    if reduction == "mean":
         total = float(weights.sum())
         if total > 0:
             weights = weights / total
-        if "mean" not in _logged_normalization_modes:
-            logger.info("Weight normalization: 'mean' (token-mean loss)")
-            _logged_normalization_modes.add("mean")
-    elif callable(normalization):
-        weights = normalization(weights)
-        if "callable" not in _logged_normalization_modes:
-            logger.info("Weight normalization: custom callable")
-            _logged_normalization_modes.add("callable")
-    elif normalization != "none":
-        raise ValueError(f"Unknown normalization mode: {normalization!r}")
+        if "mean" not in _logged_reduction_modes:
+            logger.info("Weight reduction: 'mean' (token-mean loss)")
+            _logged_reduction_modes.add("mean")
+    elif reduction != "none":
+        raise ValueError(f"Unknown reduction mode: {reduction!r}")
 
     return tinker.Datum(
         model_input=input_model_input,
--- a/tinker_cookbook/supervised/common_test.py
+++ b/tinker_cookbook/supervised/common_test.py
@@ -1,4 +1,4 @@
-"""Tests for supervised/common.py weight normalization in datum_from_model_input_weights."""
+"""Tests for supervised/common.py weight reduction in datum_from_model_input_weights."""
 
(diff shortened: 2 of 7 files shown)


tinker_cookbook/supervised/common.py
```diff
@@ -67,6 +67,7 @@ def datum_from_model_input_weights(
     model_input: tinker.ModelInput,
     weights: torch.Tensor,
     max_length: int | None = None,
+    normalize_weights: bool = True,
```
YujiaBao: **Suggestion: generalize the normalization interface instead of a boolean flag.**

A boolean `normalize_weights` locks this into one normalization strategy (sum-to-one). There are other reasonable approaches — e.g., normalizing by number of non-zero weights (mean over target tokens only), or a user-supplied callable. Consider something like:

```python
normalization: Literal["none", "mean"] | Callable[[torch.Tensor], torch.Tensor] = "none"
```

This is more extensible, avoids a breaking default change, and makes the caller's intent explicit. `"none"` preserves current behavior, `"mean"` gives the sum-to-one normalization from this PR, and a callable covers future needs without further API changes.

Separately, defaulting to anything other than `"none"` would be a silent breaking change — this function is called from SFT, DPO, preference, and VLM classifier paths. I'd recommend defaulting to `"none"` and having SFT recipe code explicitly pass `"mean"`. DPO callers in particular should stay with `"none"` since standard DPO uses sequence log-probability (token-sum), and normalizing chosen/rejected independently introduces implicit length bias.

tinker_cookbook/supervised/common.py
```diff
@@ -121,6 +126,11 @@ def datum_from_model_input_weights(
     )
     weights = weights[1 : len(target_tokens) + 1]
 
+    if normalize_weights:
+        total_weight = float(weights.sum())
+        if total_weight > 0:
```
YujiaBao: **Minor: consider logging when normalization is active.**

Since the normalization mode meaningfully changes loss semantics, it would be helpful to emit a `logger.info` (or `logger.warning` for `"none"`) on first call so users can confirm which mode is in effect. This is especially useful for debugging training instability — the kind of issue that motivated #271 in the first place.

tinker_cookbook/supervised/common.py
YujiaBao: **DPO concern:** The callers in `dpo_datasets.py` normalize chosen and rejected independently. Standard DPO uses sequence log-probability (token-sum), not token-mean. Normalizing each side to mean-per-token introduces implicit length normalization — shorter sequences get their per-token log-probs weighted more heavily. This is a separate design choice from "fix SFT gradient variance" and may not be desirable for preference training. At minimum, DPO callers should explicitly pass `normalize_weights=False`.

tinker_cookbook/supervised/common_test.py
```diff
@@ -0,0 +1,105 @@
+"""Tests for supervised/common.py weight normalization in datum_from_model_input_weights."""
```
YujiaBao: **Nit: test file location.** We recently did a round of housekeeping to consolidate on colocated `*_test.py` files for unit tests (i.e., tests live next to the code they test). This should probably move to `tinker_cookbook/supervised/common_test.py` to match the updated convention.

bledden:
Hi @YujiaBao — thanks for the thorough review! I've pushed a v2 addressing all your feedback, rebased on latest `main`:

**Changes in v2:**
- **Replaced `normalize_weights: bool`** with an extensible `normalization` parameter: `Literal["none", "mean"] | Callable[[Tensor], Tensor]`
  - Defaults to `"none"` at `datum_from_model_input_weights` (no breaking change)
  - `conversation_to_datum` defaults to `"mean"` for SFT paths
- **DPO + preference callers** now explicitly pass `normalization="none"` with comments explaining why (token-sum / length bias)
- **Added `logger.info`** on first use of each normalization mode for training debuggability
- **Moved tests** to colocated `tinker_cookbook/supervised/common_test.py`
- **Added test** for callable normalization and invalid mode `ValueError`
- **Exported `WeightNormalization`** type alias from `supervised/__init__.py`
- **Updated downstream compat test** for new `conversation_to_datum` signature

All unit tests pass (10/10) and pyright is clean (0 errors). I haven't been able to run the integration/lifecycle tests yet since I don't currently have Tinker API credits — I'm planning to cover that by this weekend. Pushing this up now in case anyone on your side can run the lifecycle suite in the meantime.

bledden:
Hi @YujiaBao — rebased on latest `main` (1d7ba97) and **all tests now pass with full API validation**:

**Unit tests (10/10):**
```
tinker_cookbook/supervised/common_test.py::test_normalization_mean_sums_to_one PASSED
tinker_cookbook/supervised/common_test.py::test_normalization_none_preserves_original PASSED
tinker_cookbook/supervised/common_test.py::test_normalization_default_is_none PASSED
tinker_cookbook/supervised/common_test.py::test_all_zero_weights_no_division_by_zero PASSED
tinker_cookbook/supervised/common_test.py::test_single_nonzero_weight_normalizes_to_one PASSED
tinker_cookbook/supervised/common_test.py::test_target_tokens_are_left_shifted PASSED
tinker_cookbook/supervised/common_test.py::test_max_length_truncation_with_mean_normalization PASSED
tinker_cookbook/supervised/common_test.py::test_max_length_truncation_without_normalization PASSED
tinker_cookbook/supervised/common_test.py::test_callable_normalization PASSED
tinker_cookbook/supervised/common_test.py::test_invalid_normalization_raises PASSED
```

**Downstream compat (9/9):** All pass including updated `conversation_to_datum` signature check.

**Integration tests (Tinker API):**
- ✅ `test_recipe_chat_sl` — 2 SFT steps, completed in 113s. Confirmed `Weight normalization: 'mean' (token-mean loss)` log emitted.
- ✅ `test_recipe_dpo` — 2 DPO steps, completed in 64s. Confirmed **no** normalization log (DPO correctly passes `normalization="none"`).

Ready for re-review when you get a chance!

PR #341, opened 2026-02-02
Fix Kimi K2 thinking preservation in renderer

### LINKED ISSUE #340: Kimi K2: renderer mismatches with the template for tool calling
The handling of when `<think></think>` is used mismatches with the [official template](https://huggingface.co/moonshotai/Kimi-K2-Thinking/blob/main/chat_template.jinja). The official template omits thinking trajectory for assistant messages before the `last_non_tool_call_assistant_msg` whereas we omit everything before the last assistant message (i.e., omitting the thinking parts in assistant messages before the tool calling).

Specifically, if we have one user query, and several assistant messages interleaved with tool calls, the thinking part should be **visible** for all assistant messages.

https://github.com/thinking-machines-lab/tinker-cookbook/blob/dee2fe4b773952d565ff1a42e57f53cf6d064745/tinker_cookbook/renderers/kimi_k2.py#L443

For example, if a trajectory is:
```
[tool_declare]
[system]
[user]
[assistant] (thinking part 1) with tool call
[tool response]
[assistant] (thinking part 2) user output
```

Then the official version will output both thinking parts, while we only have thinking part 2.

TonyLianLong:
- Update Kimi K2 renderer to fix two issues:
    1. Inclusive boundary (`>=`) incorrectly preserved the last non-tool-call assistant’s thinking.
    2. Missing-sentinel case (`last_assistant_idx == -1`) incorrectly stripped thinking everywhere.
  - Align logic with HuggingFace template behavior when tool_calls is empty/
    missing.
  - Add tests covering thinking stripping/preservation across tool-call and non-tool-call suffix cases.

Fixes #340

### DIFF
--- a/tinker_cookbook/renderers/base.py
+++ b/tinker_cookbook/renderers/base.py
@@ -937,6 +937,8 @@ class Renderer(ABC):
     ) -> list[tuple[tinker.ModelInput, torch.Tensor]]:
         """
         Build tokens and per-token weights for supervised fine-tuning.
+        This function returns a list of examples in the form of tuples, where each tuple contains a model input and a tensor of weights.
+        This is needed because some renderers do not satisfy the extension property, so we need to return a list of examples instead of a single example.
 
         This default implementation concatenates rendered messages in order, which assumes the renderer satisfies the extension property.
         Override this method if your renderer does not satisfy the extension property.
@@ -945,6 +947,7 @@ class Renderer(ABC):
         if self.has_extension_property:
             return [self.build_supervised_example(messages, train_on_what=train_on_what)]
         else:
+            # TODO: Add a default implementation that calls `build_supervised_example` for each message and merges examples with shared prefixes.
             raise NotImplementedError(
                 "build_supervised_examples has not been implemented for this renderer."
             )


TonyLianLong:
Well it seems like the change conforms to kimi's official renderer but breaks `test_thinking_generation_parse_correspondence`.

Specifically, after the generation of the final answer, all the thinking traces (including the one in the last text message) are omitted.

This is the behavior of Kimi's hf template. However, I'm not sure if this is what we want in tinker cookbook because the rollouts will not be amenable to SFT in this case.

One potential fix (that deviates from hf template) but would preserve all the thinking for the last round of conversation between user and assistant:
```bash
diff --git a/tinker_cookbook/renderers/kimi_k2.py b/tinker_cookbook/renderers/kimi_k2.py
index e1621a3..18abff2 100644
--- a/tinker_cookbook/renderers/kimi_k2.py
+++ b/tinker_cookbook/renderers/kimi_k2.py
@@ -505,7 +505,10 @@ class KimiK2Renderer(Renderer):
 
         # Find last assistant message without tool calls (matches HF template behavior).
         last_assistant_idx = -1
-        for idx in range(len(messages) - 1, -1, -1):
+        # We intentionally skip the last message (which differs from HF template behavior) since for a complete conversation, 
+        # we would want to preserve the thinking of the last round of conversation between the user and the assistant (which could include multiple assistant messages and tool calls). 
+        # This is because the trajectory would then be taken for SFT without losing all the thinking content.
+        for idx in range(len(messages) - 2, -1, -1):
             if messages[idx]["role"] == "assistant" and not messages[idx].get("tool_calls"):
                 last_assistant_idx = idx
                 break
```

May worth an discussion. @joschu

joschu (approved):

TonyLianLong:
After discussions with @joschu , `build_supervised_examples` and `LAST_ASSISTANT_TURN` are added.

tinker_cookbook/renderers/base.py
```diff
@@ -929,6 +930,25 @@ def build_generation_prompt(
...
+
+        This default implementation concatenates rendered messages in order, which assumes the renderer satisfies the extension property.
+        Override this method if your renderer does not satisfy the extension property.
+        """
+
+        if self.has_extension_property:
+            return [self.build_supervised_example(messages, train_on_what=train_on_what)]
+        else:
+            raise NotImplementedError(
+                "build_supervised_examples has not been implemented for this renderer."
```
joschu, line 952: add todo about having default impl here

joschu (approved):

PR #349, opened 2026-02-05
[tinker-cookbook] add a way to register custom renderers & tokenizers

kennyyu:
This adds a way to dynamically register new renderers and tokenizers. This allows users to use renderers + tokenizers not in the cookbook repo and have all of the tinker cookbook `get_renderer()`, `get_tokenizer()` continue to work without any changes.

### DIFF
--- a/tinker_cookbook/tokenizer_utils.py
+++ b/tinker_cookbook/tokenizer_utils.py
@@ -7,6 +7,7 @@ Avoid importing AutoTokenizer and PreTrainedTokenizer until runtime, because the
 
 from __future__ import annotations
 
+from collections.abc import Callable
 from functools import cache
 from typing import TYPE_CHECKING, Any, TypeAlias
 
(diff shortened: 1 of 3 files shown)


tinker_cookbook/tokenizer_utils.py
```diff
@@ -19,9 +20,68 @@
...
+    Returns:
+        True if the tokenizer was unregistered, False if it wasn't registered.
+    """
+    if name in _CUSTOM_TOKENIZER_REGISTRY:
+        del _CUSTOM_TOKENIZER_REGISTRY[name]
+        return True
+    return False
+
 
-@cache
```
tyler-griggs, line 23: To sanity check: if the tokenizer is not custom (ie, it's from HF) then HF handles caching. If it's custom, the expectation is that the user `@cache`'s the custom tokenizer - is that right?

tyler-griggs (approved):
nice!

tinker_cookbook/tokenizer_utils.py
kennyyu: Yep that's right!

PR #358, opened 2026-02-09
[renderer] fix regression in renderer with no user messages

andy-thinkingmachines:
# Description 
In https://github.com/thinking-machines-lab/tinker-cookbook/pull/341 we created a regression to assume that there will always be a User Message when taking the max operation over all messages. This will raise a `ValueError` if there is None. 

We already assume in https://github.com/andy-thinkingmachines/tinker-cookbook/blob/1c6497626c18553bf4022b633a4d97b9f946d487/tinker_cookbook/renderers/base.py#L1003 that if there are no user messages then `last_user_idx=-1` so this just sets this default explicitly. 

# Tests
-[x] Added pytest`test_supervised_example_no_user_messages` in `tinker_cookbook/tests/test_renderers.py` to catch for this regression.

### DIFF
--- a/tinker_cookbook/renderers/base.py
+++ b/tinker_cookbook/renderers/base.py
@@ -983,7 +983,8 @@ class Renderer(ABC):
             )
 
         last_user_idx = max(
-            idx for idx, message in enumerate(messages) if message["role"] == "user"
+            (idx for idx, message in enumerate(messages) if message["role"] == "user"),
+            default=-1,
         )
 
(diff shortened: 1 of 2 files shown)


TonyLianLong:
LGTM! Also wondering in what cases would we have no user messages? Like system message only?

andy-thinkingmachines:
> LGTM! Also wondering in what cases would we have no user messages? Like system message only?

I think for several SFT workloads we typically will only have assistant messages and train without masking.

PR #359, opened 2026-02-10
Kimi2.5 Vision Renderering Support

opherlieber:


tinker_cookbook/image_processing_utils.py
```diff
@@ -28,8 +28,12 @@ def get_image_processor(model_name: str) -> ImageProcessor:
 
     from transformers.models.auto.image_processing_auto import AutoImageProcessor
 
-    processor = AutoImageProcessor.from_pretrained(model_name, use_fast=True)
-    assert processor.is_fast, f"Could not load fast image processor for {model_name}"
+    kwargs: dict[str, Any] = {}
+    if model_name == "moonshotai/Kimi-K2.5":
+        kwargs["trust_remote_code"] = True
+        kwargs["revision"] = "3367c8d1c68584429fab7faf845a32d5195b6ac1"
```
nealwu, line 34: what is this?

tinker_cookbook/renderers/base.py
```diff
@@ -1101,6 +1101,9 @@ def get_number_of_image_patches(
     ) -> int:
         raise NotImplementedError()
 
+    def get_resize_config(self, image_data: dict) -> dict:
```
nealwu: this is a nit but can we make these types `dict[str, Any]`?

tinker_cookbook/image_processing_utils.py
opherlieber: Kimi requires remote-code for the tokenizer (Like in _get_hf_tokenizer), i.e. running the tokenizer code from the HF hub (I did try to upgrade transformers in this PR to try and remove this dependency but it broke a bunch of stuff, we can consider that in a separate PR)

tinker_cookbook/renderers/base.py
opherlieber: Updated

PR #392, opened 2026-03-03
fix: use math.ceil for batch count to not ignore remainder

Jah-yee:
Issue #381: Use math.ceil instead of floor division (//) to not
silently ignore the last few rows when batch_size does not divide evenly.

Happy to make any changes if needed!

### DIFF
--- a/tinker_cookbook/recipes/sl_loop.py
+++ b/tinker_cookbook/recipes/sl_loop.py
@@ -4,6 +4,7 @@ Uses existing modules but with a simple, flat training loop.
 """
 
 import logging
+import math
 import time
 
 import chz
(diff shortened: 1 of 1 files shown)


YujiaBao:
Thanks for the contribution! I think dropping the last incomplete batch is actually intentional here — this is a common pattern in ML training (equivalent to `drop_last=True` in PyTorch's DataLoader). A partial batch would have different effective gradient magnitude than full batches, which can cause a training spike at the end of the epoch.

The existing code already handles the boundary safely with `min()` on line 107, so there's no out-of-bounds risk — those trailing examples are just intentionally skipped.

If you'd like, a nice alternative would be to add a log warning when examples get dropped, so it's not silent:

```python
n_dropped = len(train_dataset) % config.batch_size
if n_dropped:
    logger.info(f"Dropping last {n_dropped} examples (batch_size={config.batch_size})")
```

PR #403, opened 2026-03-09
Fix TinkerMessageCompleter dropping tool_calls

joschu:
## Summary
- `TinkerMessageCompleter.__call__` was constructing the return `Message` with only `role` and `content`, silently dropping `tool_calls` from the parsed response.
- Now preserves `tool_calls` when present in the output of `renderer.parse_response`.

## Test plan
- [ ] Verify tool-calling RL envs / evals that use `TinkerMessageCompleter` now receive tool calls in responses


dphuang2 (approved):

dphuang2:
we really need better testing in the cookbook, thanks!

PR #404, opened 2026-03-09
Add Granite 4.0 renderer

eltonjohnfanboy:
Hi! 
I was doing some trainings using the IBM Granite 4.0 models with Tinker and needed a renderer for the Granite 4.0 chat format. These models use a different template from the existing renderers:
```yaml
<|start_of_role|>system<|end_of_role|>You are a helpful assistant.<|end_of_text|>
 <|start_of_role|>user<|end_of_role|>Hello<|end_of_text|>
 <|start_of_role|>assistant<|end_of_role|><think>reasoning</think>answer<|end_of_text|>
```
So this PR adds two renderers:
- Granite4Renderer — thinking enabled, with tool calling support
- Granite4DisableThinkingRenderer — thinking disabled by prefilling an empty <think>\n</think>\n block

Added also tests for both renderers covering system messages, thinking blocks, tool calling, and roundtrip parsing. 16 tests pass, 5 skip (because tool format intentionally differs from HF's available_tools role, and the disable-thinking variant skips thinking roundtrip tests).

### DIFF
--- a/tinker_cookbook/tests/test_renderers.py
+++ b/tinker_cookbook/tests/test_renderers.py
@@ -24,7 +24,7 @@ Testing guidelines:
 
 import copy
 import json
-from typing import Callable, cast
+from typing import Any, Callable, cast
 
 import pytest
 import tinker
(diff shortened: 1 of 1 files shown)


YujiaBao:
Thanks for the PR! Currently the Tinker cookbook renderers only cover models officially supported by the Tinker API. Since IBM Granite isn't supported at the moment, we'll hold off on adding this to avoid confusion. Happy to revisit if that changes!

eltonjohnfanboy:
Understood, thanks!! :))

Issue #407, opened 2026-03-09
get_recommended_renderer returns kimi_k2 for all "moonshotai" models


pcmoritz:
See https://github.com/thinking-machines-lab/tinker-cookbook/blob/66d1cdc38dc28416a092de6a5e52d634ae90663b/tinker_cookbook/model_info.py#L138

I believe for K2.5 models, we should return the `"kimi_k25"` renderer.

PR #408, opened 2026-03-09
Add docs on how to integrate tinker-cookbook with litellm

pcmoritz:
I learned this from https://medium.com/@yugez/tuning-any-ai-agent-with-tinker-agent-lightning-part-1-1d8c9a397f0e

This example makes it easy to convert agentic harnesses / programs that use litellm into token in / token out format, so they can target the Tinker API and the tokens can be used for training.

If we want to merge this, we should do a few things to clean it up:

- [x] Fix https://github.com/thinking-machines-lab/tinker-cookbook/issues/407 and remove the hard coded kimi_k25 renderer
- [ ] Remove the hard-coded max_tokens
- [ ] Make sure the docs render properly
- [ ] Possibly add a test so this integration won't break in the future

This already runs successfully with a tool heavy agentic harness. It can be improved a little by e.g. translating tinker's `tinker.BadRequestError` into litellm's ContextWindowExceededError.

### DIFF
--- a/docs/compatible-apis/litellm.mdx
+++ b/docs/compatible-apis/litellm.mdx
@@ -101,21 +101,24 @@ class TinkerCookbookLLM(CustomLLM):
         renderer = _get_renderer(base_model)
         sampling_client = _get_sampling_client(base_model)
 
-        # Render messages -> token IDs
-        tinker_msgs = _to_tinker_messages(messages)
+        # Inject tool declarations into the prompt
+        tools = optional_params.get("tools")
+        if tools:
+            tool_specs = [t["function"] for t in tools if "function" in t]
+            match tinker_msgs:
+                case [{"role": "system", "content": system_prompt}, *rest]:
+                    pass
+                case rest:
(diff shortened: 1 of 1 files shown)


dphuang2:
this is good! its also really nice since a lot of people build their agents with LiteLLM and litellm also has nice callbacks for integrations like langfuse

I was thinking about adding a recipe in the cookbook on how to fine tune an agent that you've already built outside of Tinker (either with raw OAI/Anthropic calls or agent frameworks), and this would be a good start.

pcmoritz:
Yes I totally agree! Let me know if you want to merge this PR and if there is anything else that should be fixed! If you want to take it over and merge it some other form, feel free to do that. I'm also attempting to post train an agent on the tokens that are returned and can keep the thread updated with what I learn :)

YujiaBao:
Thanks for the suggestion. We have integrated litellm on this PR https://github.com/thinking-machines-lab/tinker-cookbook/pull/458

PR #409, opened 2026-03-09
Restructure tests and add daily recipe smoke test CI

YujiaBao:
## Summary

Restructure the test directory layout, add automated smoke testing for recipe train scripts, and harden the test infrastructure for long-term maintainability.

### 1. Colocate unit tests with source code

Unit tests now live alongside the modules they test as `*_test.py` files, following monorepo conventions:

```
tinker_cookbook/
├── renderers/
│   ├── renderers_test.py
│   ├── parsing_test.py
│   ├── renderer_pickle_test.py
│   ├── tool_calling_test.py
│   └── ...
├── rl/
│   ├── message_env_test.py
│   ├── multiturn_weight_assignment_test.py
│   ├── builder_pickle_test.py
│   ├── rollout_logging_test.py
│   └── ...
├── supervised/
│   └── resume_test.py
├── recipes/
│   ├── distillation/harbor_multiturn_test.py
│   ├── harbor_rl/harbor_tools_test.py
│   ├── search_tool/chroma_pickle_test.py
│   ├── verifiers_rl/verifiers_pickle_test.py
│   └── ...
├── utils/
│   ├── logtree_test.py
│   ├── trace_test.py
│   └── ...
```

Test-only utilities (`conversation_generator`, `mock_logger`) are inlined into their single consumers — nothing test-related ships in the package (hatch excludes `*_test.py`).

### 2. Smoke tests in `tests/`

End-to-end recipe smoke tests live in `tests/` (flat, no subdirectory):

```
tests/
├── __init__.py
├── conftest.py          # pytest skip hook (skip locally, fail on CI if no API key)
├── helpers.py           # run_recipe() helper with real-time output streaming
├── test_recipe_*.py     # one file per recipe
├── smoke_tests.py       # existing direct API smoke tests
└── ...                  # standalone validation scripts
```

### 3. Auto-discovering recipe smoke test CI

The `smoke-test-recipes.yaml` workflow:
- **Auto-discovers** test files: a `discover` job finds all `test_*.py` in `tests/`, then generates a matrix
- **Each recipe runs as a parallel CI job** with its own status, logs, and retries
- **Real-time output streaming**: recipe stdout is printed line-by-line in CI logs for debuggability
- **Pass criteria**: recipe passes once step 1 begins (confirming step 0 completed — sampling, training, and eval all worked), then gets killed
- **Adding a new recipe smoke test**: just add `tests/test_recipe_foo.py` — no workflow changes needed
- **Trigger**: daily at 6am UTC + manual `workflow_dispatch`

### 4. Hardened test infrastructure

- **Single source of truth**: removed `scripts/smoke_test_recipes.sh` (duplicated recipe list and run logic). `pytest tests/` is now the only entry point for smoke tests
- **CI safety**: missing `TINKER_API_KEY` skips tests locally but fails hard on CI (detects `CI` env var)
- **Non-blocking I/O**: uses `select()` with 5s timeout before `readline()` so recipe timeout is enforced even during quiet hangs
- **Exit code validation**: step-0 success path requires `exit code == 0` — a recipe that crashes after printing step 0 correctly fails the test
- **Cross-platform**: replaced `preexec_fn=os.setsid` with `process_group=0` (Python 3.11+)
- **Smaller smoke test configs**: reduced batch/group sizes and disabled evals (`eval_every=0`) so recipes complete within timeout
- **Fixed broken args**: multiplayer_rl tests were passing invalid `behavior_if_log_dir_exists` arg (not in their CLIConfig)
- **Documented uncovered recipes**: `conftest.py` lists all recipes without smoke tests and why

### 5. Other fixes

- `test_renderer_parsing.py` was importing `get_tokenizer` from `test_renderers.py` instead of the library
- Split upstream `test_pickle_builders.py` into per-module colocated test files
- Split upstream `test_multiturn_weight_assignment.py` — RL core tests vs harbor recipe tests
- Added `workflow_dispatch` trigger to unit test CI
- Updated `CLAUDE.md`, `AGENTS.md`, `pyproject.toml`, CI workflows

### What these smoke tests are for
- **Catching recipe-level regressions**: verifying that each recipe's end-to-end pipeline (config parsing, data prep, training loop, eval) still runs without crashing
- **Daily CI**: fast feedback when recipe code changes break something
- **Single model per recipe**: each recipe runs with one (small) model using minimal configs

### What these smoke tests are NOT for
- **API/SDK correctness**: testing that `forward_backward` works for every supported model belongs in the [tinker SDK repo](https://github.com/thinking-machines-lab/tinker), not here
- **Full model coverage**: we do not test every recipe × model combination (that's a separate weekly integration suite, see below)
- **Training quality**: we only check that training starts, not that loss decreases or outputs are correct

### Follow-up TODOs
1. **Replace pattern-based step detection**: the current pass/fail logic relies on regex matching against log output (`Step 1`, `batch_idx=1`). A follow-up PR will modify the base recipe infrastructure to expose a shared smoke test harness (e.g., exit-after-N-steps mode) so all recipes signal completion through a common mechanism rather than ad-hoc log scraping
2. **Verify training metrics**: currently we only check that training starts, not that it produces correct results. Future work could assert expected loss ranges or metric values per recipe to catch silent regressions (e.g., a renderer bug that produces valid but wrong tokens)
3. **Weekly full-model integration tests** (`tests/integration/`): a separate CI workflow that runs each recipe with its default config across all compatible models — no size reduction, longer timeouts. Purpose: catch model-specific regressions that daily smoke tests (single model, tiny configs) would miss

### Design decisions
- **Dataset downloads on every run**: each smoke test downloads its HF dataset fresh. This is intentional — it verifies the data pipeline (HF API, dataset parsing) works end-to-end. Since we run daily and datasets are small, the overhead is acceptable
- **15-minute default timeout**: generous to accommodate server cold starts and slow sampling. Individual recipes can override via `timeout=` parameter

## Smoke test CI results

**All 11 recipes passing**: [latest run](https://github.com/YujiaBao/tinker-cookbook/actions/runs/23013386437)

## Recipes covered (11 smoke tests)

chat_sl, dpo, vlm_classifier, off_policy_reasoning, shorter, on_policy_distillation, on_policy_multi_teacher, rlhf_pipeline, guess_number, text_arena, twenty_questions

**Skipped** (require external setup): prompt_distillation, harbor_rl, code_rl, search_tool, verifiers_rl, if_rl, math_rl, rubric

## Test plan
- [x] `pytest tinker_cookbook/` — 342 passed, 39 skipped
- [x] `pytest tests/` — [11/11 smoke tests passing](https://github.com/YujiaBao/tinker-cookbook/actions/runs/23013386437)
- [x] `ruff check` — all passed
- [x] `ruff format --check` — all passed
- [x] No remaining references to old `tinker_cookbook/tests` paths
- [x] `TINKER_API_KEY` added as repo secret

### DIFF
--- a/tests/test_recipe_guess_number.py
+++ b/tests/test_recipe_guess_number.py
@@ -7,6 +7,6 @@ def test_guess_number():
         [
             "batch_size=8",
             "group_size=2",
-            "eval_every=999",
+            "eval_every=0",
         ],
     )
--- a/tests/test_recipe_text_arena.py
+++ b/tests/test_recipe_text_arena.py
@@ -7,6 +7,6 @@ def test_text_arena():
         [
             "batch_size=16",
             "num_train_datapoints=128",
-            "eval_every=999",
+            "eval_every=0",
         ],
     )


joschu:
Looks good to me. Codex review found a couple issues

[P1] Smoke test helper treats crashed recipes as success
The helper currently returns success whenever the child process exits after emitting a step-0 pattern, regardless of its exit status. That means a recipe can start step 0, then crash with a non-zero code, and this smoke test still goes green. The success path here should require proc.returncode == 0 or otherwise distinguish intentional termination from recipe failure.


[P2] Per-test timeout is not enforced for quiet hangs
The loop checks elapsed time before calling proc.stdout.readline(), but readline() blocks until a newline arrives. If a recipe stalls without producing more output, control never returns to the timeout check, so SMOKE_TEST_TIMEOUT is effectively bypassed and the workflow only fails at the outer job timeout. This needs nonblocking reads or a timed poll/read loop.

joschu (approved):

YujiaBao:
> Looks good to me. Codex review found a couple issues
> 
> [P1] Smoke test helper treats crashed recipes as success The helper currently returns success whenever the child process exits after emitting a step-0 pattern, regardless of its exit status. That means a recipe can start step 0, then crash with a non-zero code, and this smoke test still goes green. The success path here should require proc.returncode == 0 or otherwise distinguish intentional termination from recipe failure.
> 
> [P2] Per-test timeout is not enforced for quiet hangs The loop checks elapsed time before calling proc.stdout.readline(), but readline() blocks until a newline arrives. If a recipe stalls without producing more output, control never returns to the timeout check, so SMOKE_TEST_TIMEOUT is effectively bypassed and the workflow only fails at the outer job timeout. This needs nonblocking reads or a timed poll/read loop.

@joschu  Thanks! Fixed in the latest commit.

P1: Changed the step-0 success path to require proc.returncode == 0. A recipe that crashes after printing step 0 now correctly fails the test.
P2: Replaced the blocking readline() with select() using a 5s timeout before each read. The timeout check now runs every 5s even if the recipe produces no output, so quiet hangs are caught within 5s of the deadline.

Both fixes are in `tests/helpers.py` — latest CI run is all green: https://github.com/YujiaBao/tinker-cookbook/actions/runs/23013386437

PR #410, opened 2026-03-09
Fix get_recommended_renderer returning kimi_k2 for Kimi-K2.5

### LINKED ISSUE #407: get_recommended_renderer returns kimi_k2 for all "moonshotai" models

YujiaBao:
## Summary
- Branch on `version_str` for moonshotai models so `Kimi-K2.5` returns `["kimi_k25", "kimi_k25_disable_thinking"]` instead of `["kimi_k2"]`
- Adds explicit `raise ValueError` for unknown future moonshot versions

Closes #407

## Test plan
- [x] Verified `moonshotai/Kimi-K2-Thinking` → `["kimi_k2"]`
- [x] Verified `moonshotai/Kimi-K2.5` → `["kimi_k25", "kimi_k25_disable_thinking"]`

### DIFF
--- a/tinker_cookbook/model_info.py
+++ b/tinker_cookbook/model_info.py
@@ -135,7 +135,12 @@ def get_recommended_renderer_names(model_name: str) -> list[str]:
     elif attributes.organization == "openai":
         return ["gpt_oss_no_sysprompt", "gpt_oss_medium_reasoning"]
     elif attributes.organization == "moonshotai":
-        return ["kimi_k2"]
+        if attributes.version_str == "K2.5":
+            return ["kimi_k25", "kimi_k25_disable_thinking"]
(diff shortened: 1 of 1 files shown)


pcmoritz:
Nice, thanks for fixing! I wonder if it makes sense to add the recommended renderer to the `model_info` in some form going forward or keep a separate database of renderers, so it will be less likely that such bugs happen in the future (because people who add the model will be "forced" to think about which renderer should be used).

PR #411, opened 2026-03-09
Add multi-turn on-policy distillation for Harbor environments                                                                                      

ZitongYang:
## Summary                                                                        
  - Add `HarborDistillationEnvGroupBuilder` and dataset classes for multi-turn on-policy distillation with zero reward (KL-only training signal) in sandbox environments
  - Add Harbor multi-turn distillation training recipe (`on_policy_distillation_harbor_multi_turn.py`)
  - Add tests verifying that `trajectory_to_data` correctly assigns mask=1 to agent-generated tokens and mask=0 to environment tokens in multi-turn  tool-use trajectories

  ## Test plan
  - [x] `uv run pytest tinker_cookbook/tests/test_multiturn_weight_assignment.py -v` — 12 tests pass (no API key or network required)

### DIFF
--- a/tinker_cookbook/recipes/distillation/README.md
+++ b/tinker_cookbook/recipes/distillation/README.md
@@ -85,7 +85,7 @@ Multi-turn distillation reuses three layers of infrastructure:
 
 2. **`harbor_rl` recipe** (`tinker_cookbook/recipes/harbor_rl/`) — Applies `tool_use` to Harbor sandbox tasks. `HarborBashTool` wraps a sandbox as a `@tool`-decorated bash command. `HarborEnvGroupBuilder` creates sandboxed environments with task-specific grading via `HarborReward`.
 
-3. **Multi-turn distillation** (`tinker_cookbook/distillation/multiturn_env.py`) — Reuses `HarborEnvGroupBuilder` directly, passing `reward_fn=_zero_reward` (always returns 0.0) to override the default `HarborReward`. The only training signal is KL divergence against the teacher.
+3. **Multi-turn distillation** (`tinker_cookbook/recipes/distillation/harbor_multiturn.py`) — `HarborDistillationDatasetBuilder` subclasses `HarborDatasetBuilder`, passing `reward_fn=zero_reward` (always returns 0.0) to override the default `HarborReward`. The only training signal is KL divergence against the teacher.
 
 Environment-provided tokens (system prompt, user message, tool responses, assistant headers) are masked out during training — only the student's generated tokens contribute to the loss.
 
@@ -118,8 +118,7 @@ python -m tinker_cookbook.recipes.distillation.on_policy_distillation_harbor_mul
     command_timeout=120 \
     save_every=5 \
     eval_every=5 \
-    wandb_project=tinkercookbook-cpt \
-    wandb_name=onpodi-kimiharbormultiturn-20260305-231347
+    wandb_name=cookbook-multiturn-onpodi
 ```
 
 ## Additional details
--- a/tinker_cookbook/recipes/distillation/on_policy_distillation_harbor_multi_turn.py
+++ b/tinker_cookbook/recipes/distillation/on_policy_distillation_harbor_multi_turn.py
@@ -22,8 +22,7 @@ Example usage:
         command_timeout=120 \
         save_every=5 \
         eval_every=5 \
-        wandb_project=tinkercookbook-cpt \
-        wandb_name=onpodi-kimiharbormultiturn-20260305-231347
+        wandb_name=cookbook-multiturn-onpodi
 """
 
 import asyncio
--- a/tinker_cookbook/recipes/harbor_rl/harbor_env.py
+++ b/tinker_cookbook/recipes/harbor_rl/harbor_env.py
@@ -141,7 +141,7 @@ class HarborEnvGroupBuilder(EnvGroupBuilder):
         return [(0.0, {}) for _ in trajectory_group]
 
     def logging_tags(self) -> list[str]:
-        return ["harbor", self.task.task_name]
+        return ["harbor"]
 
 
 class HarborDataset(RLDataset):


YujiaBao:
@ZitongYang can we also update the readme.md file under the distilllation recipe dir?

ZitongYang:
> @ZitongYang can we also update the readme.md file under the distilllation recipe dir?

Sure! updated the readme @tyler-griggs

tinker_cookbook/recipes/distillation/on_policy_distillation_harbor_multi_turn.py
```diff
@@ -0,0 +1,186 @@
...
+Multi-turn on-policy distillation with Harbor sandbox environments.
+
+The student interacts with a harbor sandbox over multiple turns (tool calls),
+and the training signal comes from KL divergence against a teacher model.
+Environment responses are masked out; only student-generated tokens contribute.
+
+Example usage:
+    python -m tinker_cookbook.recipes.distillation.on_policy_distillation_harbor_multi_turn \
+        model_name=moonshotai/Kimi-K2-Thinking \
+        load_checkpoint_path=tinker://dadeaa6a-4df5-52cb-b511-1bb28f769ad4:train:0/weights/final \
```
YujiaBao: it is not clear what this checkpoint path refers to

ZitongYang: my bad, this is for testing. it's a specific checkpoint where i have an expected behavior in mind

tinker_cookbook/distillation/multiturn_env.py
```diff
@@ -0,0 +1,162 @@
...
+        self.env_group_builders = env_group_builders
+        self.batch_size = batch_size
+
+    def get_batch(self, index: int) -> Sequence[EnvGroupBuilder]:
+        start = index * self.batch_size
+        end = start + self.batch_size
+        return self.env_group_builders[start:end]
+
+    def __len__(self) -> int:
+        return len(self.env_group_builders) // self.batch_size
```
tyler-griggs: Not sure if our convention is to drop incomplete batches or not? The `HarborDataset` behavior is to do ceiling division instead of floor.

tyler-griggs:
This looks pretty good to me, thanks! There is a lot of duplication between `HarborDistillationEnvGroupBuilder` and `HarborEnvGroupBuilder`. We could even consider just using `HarborEnvGroupBuilder` but replacing the reward function? Don't have a strong opinion here.

tinker_cookbook/distillation/multiturn_env.py
ZitongYang: outdated after the change below

ZitongYang:
Addressed the deduplication feedback

Changes:
- `harbor_env.py::HarborEnvGroupBuilder` now accepts an optional `reward_fn: RewardFn | None`, defaulted to `HarborReward`
- `HarborDistillationDatasetBuilder` now imports `HarborEnvGroupBuilder` and setting `reward_fn=_zero_reward`

tinker_cookbook/distillation/multiturn_env.py
```diff
@@ -0,0 +1,65 @@
...
+model (computed in the training loop).
+"""
+
+from __future__ import annotations
+
+import logging
+
+import chz
+
+from tinker_cookbook.recipes.harbor_rl.harbor_env import (
```
YujiaBao: This module imports directly from `recipes/harbor_rl/`, which inverts the intended dependency direction (library code depending on recipe code). Since this is Harbor-specific, consider moving it into `recipes/distillation/` or `recipes/harbor_rl/` instead — recipe dirs are the right place for task/benchmark-specific details, while `tinker_cookbook/distillation/` should stay cookbook-agnostic.

tinker_cookbook/distillation/multiturn_env.py
```diff
@@ -0,0 +1,65 @@
...
+logger = logging.getLogger(__name__)
+
+
+async def _zero_reward(history) -> tuple[float, dict[str, float]]:
+    """Reward function that always returns zero. KL penalty is the only signal."""
+    return 0.0, {}
+
+
+@chz.chz
+class HarborDistillationDatasetBuilder(RLDatasetBuilder):
```
YujiaBao: `HarborDistillationDatasetBuilder` is ~90% identical to `HarborDatasetBuilder` in `harbor_env.py` (same fields, same builder loop). Consider subclassing or extracting a shared helper to avoid the duplication — especially since they'll need to stay in sync if `HarborEnvGroupBuilder` gains new parameters.

tinker_cookbook/recipes/harbor_rl/harbor_env.py
```diff
@@ -74,6 +75,7 @@ def __init__(
         grader_timeout: int = 60,
         max_trajectory_tokens: int = 32 * 1024,
         sandbox_factory: SandboxFactory | None = None,
+        reward_fn: RewardFn | None = None,
```
YujiaBao, line 78: Nit: the existing `HarborDatasetBuilder` (line 165) doesn't accept or forward `reward_fn`, so this new parameter is only usable via direct `HarborEnvGroupBuilder` construction. Worth threading it through for consistency.

tinker_cookbook/distillation/multiturn_env.py
```diff
@@ -0,0 +1,65 @@
...
+    HarborDataset,
+    HarborEnvGroupBuilder,
+    HarborTask,
+    SandboxFactory,
+)
+from tinker_cookbook.rl.types import RLDataset, RLDatasetBuilder
+
+logger = logging.getLogger(__name__)
+
+
```
YujiaBao: `history` is untyped here. Should be `history: list[Message]` to match the `RewardFn` signature.

tinker_cookbook/tests/test_multiturn_weight_assignment.py
```diff
@@ -0,0 +1,331 @@
...
+    )
+    traj = asyncio.run(do_single_rollout(_make_stub_policy(), env))
+    data = trajectory_to_data(traj, traj_advantage=1.0)
+    return traj, data
+
+
+class TestEndToEndToolUseRollout:
+    def test_trajectory_has_two_transitions(self):
+        traj, _ = _run_e2e_rollout()
+        assert len(traj.transitions) == 2
```
YujiaBao, line 267: Minor: these tests use `asyncio.run()` directly rather than `@pytest.mark.asyncio` + `async def`. Works fine, but the rest of the test suite uses pytest-asyncio — might be worth staying consistent.

tinker_cookbook/distillation/multiturn_env.py
ZitongYang: Yeah, in hindsight it seems like the `distillation` should all be in `recipes/distillation/`. For now moving `multiturn_env.py` to `recipes/distillation/`.

tinker_cookbook/tests/test_multiturn_weight_assignment.py
ZitongYang: Hi Yujia, I actually didn't find the usage of `@pytest.mark.asyncio` in the cookbook https://github.com/search?q=repo%3Athinking-machines-lab%2Ftinker-cookbook%20pytest.mark.asyncio&type=code ? I have perhaps missed something here, can you point me to an example?

ZitongYang:
Thanks @YujiaBao 

Addressed all the comments except the test suite one.

  - harbor_env.py: Added reward_fn and max_trajectory_tokens fields to HarborDatasetBuilder. Extracted _make_env_group_builders() helper that threads all fields (including reward_fn) through to HarborEnvGroupBuilder. Refactored __call__ to use the helper for both train and eval datasets.
  - harbor_multiturn.py: HarborDistillationDatasetBuilder is now a thin subclass of HarborDatasetBuilder — only overrides __call__ to skip the eval dataset. Renamed _zero_reward → zero_reward (public API).
  - on_policy_distillation_harbor_multi_turn.py: Passes reward_fn=zero_reward explicitly at construction.
  - test_multiturn_weight_assignment.py: Updated for the rename.

tinker_cookbook/recipes/distillation/README.md
```diff
@@ -73,6 +73,55 @@ python -m tinker_cookbook.recipes.distillation.on_policy_distillation \
...
+    command_timeout=120 \
+    save_every=5 \
+    eval_every=5 \
+    wandb_project=tinkercookbook-cpt \
+    wandb_name=onpodi-kimiharbormultiturn-20260305-231347
+```
+
 ## Additional details
 
 ### Reward calculation
```
YujiaBao, line 126: Nit: this `wandb_name` looks like it's from a specific dev run. Could you replace it with a generic placeholder (e.g. `my-harbor-distillation-run`) or just remove the line so the auto-generated name is used?

tinker_cookbook/recipes/distillation/README.md
```diff
@@ -73,6 +73,55 @@ python -m tinker_cookbook.recipes.distillation.on_policy_distillation \
...
+
+### Architecture
+
+Multi-turn distillation reuses three layers of infrastructure:
+
+1. **`tool_use` library** (`tinker_cookbook/tool_use/`) — Generic agent-tool interaction. `@tool` decorator defines tools, `build_agent_tool_env()` creates token-level RL environments from tools + renderer + reward function. `AgentToolMessageEnv` manages the message-level episode loop (append assistant message → execute tool calls → check termination).
+
+2. **`harbor_rl` recipe** (`tinker_cookbook/recipes/harbor_rl/`) — Applies `tool_use` to Harbor sandbox tasks. `HarborBashTool` wraps a sandbox as a `@tool`-decorated bash command. `HarborEnvGroupBuilder` creates sandboxed environments with task-specific grading via `HarborReward`.
+
+3. **Multi-turn distillation** (`tinker_cookbook/distillation/multiturn_env.py`) — Reuses `HarborEnvGroupBuilder` directly, passing `reward_fn=_zero_reward` (always returns 0.0) to override the default `HarborReward`. The only training signal is KL divergence against the teacher.
```
YujiaBao: Stale path: this references `tinker_cookbook/distillation/multiturn_env.py` but the file was moved to `tinker_cookbook/recipes/distillation/harbor_multiturn.py`. Also `_zero_reward` → `zero_reward`.

tinker_cookbook/recipes/distillation/on_policy_distillation_harbor_multi_turn.py
```diff
@@ -0,0 +1,189 @@
...
+        max_trajectory_tokens=24576 \
+        temperature=1.0 \
+        kl_penalty_coef=1.0 \
+        sandbox_timeout=600 \
+        command_timeout=120 \
+        save_every=5 \
+        eval_every=5 \
+        wandb_project=tinkercookbook-cpt \
+        wandb_name=onpodi-kimiharbormultiturn-20260305-231347
+"""
```
YujiaBao, line 26: Same as the README — this hardcoded `wandb_name` should be replaced with a placeholder or removed from the example.

tinker_cookbook/recipes/distillation/harbor_multiturn.py
```diff
@@ -0,0 +1,35 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+async def zero_reward(history: list[Message]) -> tuple[float, dict[str, float]]:
+    """Reward function that always returns zero. KL penalty is the only signal."""
+    return 0.0, {}
+
+
+@chz.chz
```
YujiaBao, line 26: `HarborDistillationDatasetBuilder` only differs from the parent in skipping the eval dataset (returns `None` instead of building one). Would it be simpler to add an `include_eval: bool = True` flag to `HarborDatasetBuilder` and avoid the subclass entirely? That keeps things flatter and other recipes could reuse the flag too.

tinker_cookbook/tests/test_multiturn_weight_assignment.py
```diff
@@ -0,0 +1,331 @@
+"""Tests for multi-turn weight assignment in trajectory_to_data.
```
YujiaBao, line 1: 4 of the 5 test classes here (`TestMultiTurnPrefixTrajectory`, `TestSingleTurnTrajectory`, `TestPrefixBreak`, `TestEndToEndToolUseRollout`) are testing the generic `trajectory_to_data` masking logic in `rl/data_processing.py` — they don't involve Harbor or distillation at all. Only `TestZeroReward` is recipe-specific. Consider splitting the generic tests into something like `tests/test_trajectory_to_data.py` so they're discoverable as infrastructure tests and don't get buried under a recipe-specific filename.

tinker_cookbook/recipes/distillation/harbor_multiturn.py
ZitongYang: I think it's conceptually cleaner to have a dedicated `HarborDistillationDatasetBuilder`. It makes downstream code more informative what task we are doing. The duplication is already minimal.

tinker_cookbook/tests/test_multiturn_weight_assignment.py
ZitongYang: oh this is actually the intention. Because I only care about the weight assignment, which is most relevant caveat

YujiaBao (approved):

PR #413, opened 2026-03-09
Move recommended_renderers into ModelAttributes

YujiaBao:
## Summary
- Adds `recommended_renderers: list[str]` field to `ModelAttributes`, so adding a new model **requires** specifying its renderers upfront
- Eliminates the separate dispatch logic in `get_recommended_renderer_names` — it now just returns `attributes.recommended_renderers`
- Prevents the class of bug in #407 where renderer dispatch silently falls through to the wrong default

Follow-up to #410 based on review feedback.

## Test plan
- [x] Verified all model/renderer mappings match the previous behavior
- [x] `pytest tinker_cookbook/tests/test_renderers.py` — 139 passed

### DIFF
--- a/tinker_cookbook/model_info.py
+++ b/tinker_cookbook/model_info.py
@@ -5,18 +5,20 @@ This module associates model names with metadata, which helps  training code cho
 from dataclasses import dataclass
 from functools import cache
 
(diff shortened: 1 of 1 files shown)


joschu:
good idea

PR #415, opened 2026-03-09
Log dropped examples in SL recipe

YujiaBao:
## Summary
- Adds a log message when the last few examples are dropped because `batch_size` doesn't divide the dataset evenly
- Addresses feedback from #392 and #381 — dropping the last batch is intentional (to keep batch size uniform), but it shouldn't be silent

## Test plan
- [x] Verified log message appears when `len(dataset) % batch_size != 0`: `Dropping last 28 examples to keep batch size uniform at 128`
- [x] Unit tests pass (`test_renderers.py`, `test_utils.py`)

### DIFF
--- a/tinker_cookbook/recipes/sl_loop.py
+++ b/tinker_cookbook/recipes/sl_loop.py
@@ -55,6 +55,8 @@ def main(config: Config):
     assert isinstance(dataset, datasets.DatasetDict)
     train_dataset = dataset["train"]
 
+    # Drop the last incomplete batch (like PyTorch's drop_last=True) — a partial
+    # batch has different effective gradient magnitude, which can cause a training spike.
     n_train_batches = len(train_dataset) // config.batch_size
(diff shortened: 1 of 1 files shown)


tinker_cookbook/recipes/sl_loop.py
```diff
@@ -56,6 +56,11 @@ def main(config: Config):
     train_dataset = dataset["train"]
 
     n_train_batches = len(train_dataset) // config.batch_size
+    n_dropped = len(train_dataset) % config.batch_size
```
dphuang2, line 61: can you also explain your [reasoning](https://github.com/thinking-machines-lab/tinker-cookbook/pull/381#issuecomment-4027083071) inline as a comment

YujiaBao: Done in 3dd8ddc!

PR #419, opened 2026-03-10
Add SDPO recipe with SciKnowEval support

YujiaBao:
## Self-Distilled Policy Optimization (SDPO)

Implements the SDPO algorithm from [Huebotter et al., 2026](https://arxiv.org/abs/2601.20802) ([official implementation](https://github.com/lasgroup/SDPO)) as a tinker cookbook recipe.

### What is SDPO?

Standard RL algorithms like GRPO assign a **single scalar reward** per sequence — every token in a correct solution gets the same credit, and every token in an incorrect one gets the same blame. SDPO provides **dense, token-level credit assignment** by using the model's own successful solutions as a teaching signal.

The key insight (Proposition 2.1): the SDPO gradient is a policy gradient where each token gets its own advantage, computed as the log-ratio of a solution-conditioned teacher to the student:

```
advantage_t = log π_teacher(y_t | prompt, solution, y_<t) − log π_student(y_t | prompt, y_<t)
```

Tokens where the teacher is more confident than the student get positive advantage (reinforced), and vice versa. This maps directly to tinker's `importance_sampling` loss — the same pattern used by tinker's online distillation — so we avoid `forward_backward_custom` entirely.

### Algorithm flow

For each batch, the training loop repeats the following steps:

**Step 1 — Rollout.** Sample `group_size` (default 8) responses per problem using the current policy.

**Step 2 — Identify successes.** Check which responses solved the problem correctly (e.g. correct `\boxed{}` answer for math, correct MCQ letter for SciKnowEval).

**Step 3 — Build teacher prompt.** Condition the reference model on information the student doesn't have. This can include:
- A **successful solution** from another rollout in the group (primary signal for math/MCQ).
- **Environment feedback** from the trajectory's own execution, e.g. compiler errors or failing test cases (useful for code tasks; enable with `include_environment_feedback=True`).
- **Both** — the paper (Table 6) shows these are complementary: solution alone 42.6%, feedback alone 39.9%, both together 48.3%.

**Step 4 — Compute teacher logprobs.** Use the frozen reference model to score each trajectory's response tokens under this conditioned teacher prompt.

**Step 5 — Build advantages.** For each response token, set `advantage_t = teacher_lp_t − student_lp_t`. Tokens where the teacher is more confident get positive advantages (reinforced); tokens where the student already matches get near-zero advantages.

**Step 6 — Train.** Call `forward_backward(datums, loss_fn="importance_sampling")` followed by `optim_step()`. The importance weight automatically corrects for any off-policy drift. Refresh the sampling client with the updated weights.

### Teacher regularization

We use a **frozen reference model** (θ_ref) as the teacher base. Table 4 in the paper shows this achieves 48.8 accuracy vs 36.1 for the unregularized variant, while being simpler than EMA (49.3) or trust-region (50.6).

### Results on MATH

Using Qwen3-8B on Hendrycks MATH (MATH-500 eval), SDPO improves from 43.0% to 63.0% after 180 steps:

| Step | `test/env/all/correct` | `sdpo/success_fraction` | `sdpo/mean_group_success_rate` |
|------|------------------------|-------------------------|-------------------------------|
| 0    | 0.430                  | 0.547                   | 0.408                         |
| 180  | 0.630                  | 0.781                   | 0.645                         |

### Code structure

```
tinker_cookbook/sdpo/                  # Reusable core
├── data.py                           # build_sdpo_datum (advantages = teacher_lp - student_lp)
├── teacher.py                        # Teacher prompt construction + logprob computation
└── train.py                          # Config + main() training loop

tinker_cookbook/recipes/sdpo/          # Recipe CLI
├── train.py                          # Thin CLI wrapper
├── sciknoweval_env.py                # Paper's SciKnowEval MCQ environment
└── README.md
```

### Environments

- **SciKnowEval** — the paper's primary benchmark (`hicai-zju/SciKnowEval`), MCQ science questions with domains: chemistry, physics, biology, material
- **MATH** — Hendrycks MATH (12k train / MATH-500 test)
- **GSM8K**, **Polaris**, **DeepMath** — reused from `math_rl` recipe

### Differences from the paper

- **Token-level vs full-logit distillation**: This is the main gap. The paper computes JSD (α=0.5) over the **full vocabulary distribution** at each position (with top-k=100 approximation), which tells the student "redistribute your probability mass across all tokens to match the teacher." Our implementation only computes the advantage on the **sampled token** (`teacher_lp(y_t) - student_lp(y_t)`) via `compute_logprobs`, which misses the teacher's signal about all the other tokens in the vocabulary. Closing this gap would require tinker's forward pass to return top-k logits (not just the logprob of the target token), which would allow us to compute a proper KL over the approximate vocabulary distribution.
- **LoRA vs full fine-tuning**: The paper uses full fine-tuning; this recipe uses LoRA (rank 32) for efficiency.

### Usage

```bash
# SciKnowEval (paper's benchmark)
uv run python -m tinker_cookbook.recipes.sdpo.train \
    model_name="Qwen/Qwen3-8B" env=sciknoweval \
    sciknoweval_domain=chemistry group_size=8 \
    groups_per_batch=32 learning_rate=1e-5 max_tokens=8192

# MATH
uv run python -m tinker_cookbook.recipes.sdpo.train \
    model_name="Qwen/Qwen3-8B" env=math \
    group_size=16 groups_per_batch=64 \
    learning_rate=2e-5 max_tokens=512

# Compare with GRPO baseline (same settings)
uv run python -m tinker_cookbook.recipes.math_rl.train \
    env=math model_name="Qwen/Qwen3-8B" \
    group_size=16 groups_per_batch=64 \
    learning_rate=2e-5 max_tokens=512
```

Metric names match the GRPO recipe (`env/all/correct`, `env/all/reward/total`, `test/env/all/correct`) so curves overlay directly in W&B.

## Test plan

- [x] 17 unit tests passing (`uv run pytest tinker_cookbook/tests/test_sdpo.py`)
- [x] Existing tests unaffected (139 passed)
- [x] End-to-end run on MATH (43% → 63% after 180 steps)
- [ ] End-to-end run on SciKnowEval
- [ ] Side-by-side SDPO vs GRPO comparison with matched settings

### DIFF
--- a/tinker_cookbook/recipes/sdpo/run_grpo_baseline.py
+++ b/tinker_cookbook/recipes/sdpo/run_grpo_baseline.py
@@ -1,0 +1,104 @@
+"""GRPO baseline on SciKnowEval for comparison with SDPO.
+
+Usage:
+    python -m tinker_cookbook.recipes.sdpo.run_grpo_baseline \
+        model_name=Qwen/Qwen3-8B \
+        sciknoweval_domain=chemistry \
+        groups_per_batch=32 \
+        group_size=8 \
+        learning_rate=1e-5 \
+        max_tokens=8192
+"""
+
+import asyncio
+from datetime import datetime
+
+import chz
+
+from tinker_cookbook import checkpoint_utils, cli_utils
+from tinker_cookbook.recipes.sdpo.sciknoweval_env import (
+    SciKnowEvalDatasetBuilder,
+    SciKnowEvalDomain,
+)
+from tinker_cookbook.rl.train import Config, main
+
+
+@chz.chz
+class CLIConfig:
+    model_name: str = "Qwen/Qwen3-8B"
+    lora_rank: int = 32
+    renderer_name: str | None = None
+    load_checkpoint_path: str | None = None
+
+    sciknoweval_domain: SciKnowEvalDomain = "chemistry"
+    seed: int = 0
+
+    group_size: int = 8
+    groups_per_batch: int = 32
+    learning_rate: float = 1e-5
+    max_tokens: int = 8192
+    temperature: float = 1.0
+
+    log_path: str | None = None
+    wandb_project: str | None = None
+    wandb_name: str | None = None
+
+    eval_every: int = 5
+    save_every: int = 10
+
+    base_url: str | None = None
+    behavior_if_log_dir_exists: cli_utils.LogdirBehavior = "ask"
+
+
+async def cli_main(cli_config: CLIConfig) -> None:
+    renderer_name = await checkpoint_utils.resolve_renderer_name_from_checkpoint_or_default_async(
+        model_name=cli_config.model_name,
+        explicit_renderer_name=cli_config.renderer_name,
+        load_checkpoint_path=cli_config.load_checkpoint_path,
+        base_url=cli_config.base_url,
+    )
+
+    model_slug = cli_config.model_name.replace("/", "-")
+    run_name = (
+        f"grpo-sciknoweval-{cli_config.sciknoweval_domain}-{model_slug}-"
+        f"{cli_config.learning_rate}lr-{cli_config.group_size}group-"
+        f"{cli_config.groups_per_batch}batch-seed{cli_config.seed}-"
+        f"{datetime.now().strftime('%Y-%m-%d-%H-%M')}"
+    )
+    log_path = cli_config.log_path or f"/tmp/tinker-examples/sdpo/{run_name}"
+    wandb_name = cli_config.wandb_name or run_name
+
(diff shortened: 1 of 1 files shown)


listar2000:
Hi @YujiaBao -- thanks for the recipe and been trying to reproduce SDPO on Tinker as well. However, it seems that with the newest Tinker SDK version (0.16.1), the `forward_backward_custom` API only supports loss_fn_inputs keys `{'target_tokens', 'weights'}` -- which means that those `is_advantages`/`is_old_logprobs` fields in your `sdpo_combined_loss` no longer works? 

Mind sharing what are some workarounds in this case? Thx in advance.

PR #420, opened 2026-03-11
Add Profiler with @profiled decorator and prof.phase() nesting

YujiaBao:
## Motivation

Training loops in tinker-cookbook involve many async operations (sampling, forward/backward, optimizer steps, evals, checkpointing). Understanding where wall-clock time is spent — and how much of that is local compute vs I/O wait — currently requires ~30 manual `with timed("key", metrics)` call sites that thread a mutable `metrics` dict through every function.

This PR introduces a `Profiler` class with a `@profiled` decorator and `prof.phase()` nesting that automatically records both wall-clock and CPU time, with clear separation between profiling data and training metrics.

### `Profiler` + `prof.measure()` + `prof.phase()` + `@profiled`

```python
from tinker_cookbook.utils.profiling import Profiler, profiled

prof = Profiler()

@profiled("rollout", agg=["mean", "max"])
async def do_rollout(...): ...

@profiled("train_step")
async def train_step(...): ...

for i_batch in range(n_batches):
    metrics = {"progress/batch": i_batch}

    with prof.measure() as profile:
        with prof.phase("sampling"):
            await do_rollout(...)
        with prof.phase("train"):
            await train_step(...)

    metrics.update(profile)
    ml_logger.log_metrics(metrics, step=i_batch)
```

Key design decisions:

- **Profiling data is separate from training metrics.** `prof.measure()` yields a dict that only contains `time/` and `cpu/` keys. The user's `metrics` dict is theirs — they merge explicitly with `metrics.update(profile)`.
- **`/` for hierarchy, `:` for aggregation.** Phase nesting uses `/` (`time/sampling/rollout`), aggregation stats use `:` (`time/sampling/rollout:mean`). No ambiguity.
- **`prof.phase()` for hierarchical timing.** Groups related operations under a named prefix and records the phase's own wall-clock duration, answering "where did my iteration time go?"

### Wall-clock + CPU time

Every `@profiled` function records both:
- `time/{key}` — wall-clock duration (how long the user waited)
- `cpu/{key}` — CPU time (actual compute, excludes `await` / sleep / network I/O)

The gap reveals how much time is spent waiting for the Tinker API vs doing local work:
```
time/rollout = 5.2s    # total elapsed
cpu/rollout  = 0.3s    # local compute only
# → 4.9s was I/O wait (Tinker API roundtrip + GPU execution)
```

### Phase nesting

`prof.phase("name")` groups nested `@profiled` calls under a prefix and records the phase's own duration:

```python
with prof.measure() as profile:
    with prof.phase("eval"):
        await run_evaluations_parallel(...)
    with prof.phase("sampling"):
        await gather_with_progress(...)       # 160 concurrent rollouts
    with prof.phase("train"):
        await do_train_step(...)
```

Produces:

```
time/total                     = 1654s   # full iteration
time/eval                      = 424s    # eval phase wall-clock
time/eval/run_evals            = 424s    # from @profiled
time/sampling                  = 500s    # actual wall-clock of gather
time/sampling/rollout:mean     = 307s    # per-group average
time/sampling/rollout:max      = 481s    # straggler detection
time/train                     = 722s    # train phase wall-clock
time/train/prepare_minibatch   = 18s
time/train/train_step          = 722s
```

Phases can nest further (`prof.phase("outer")` → `prof.phase("inner")` → `time/outer/inner/...`). No-op when called outside a `prof.measure()` window.

### Aggregation for repeated calls

- **Called once** → `time/rollout = 5.2`, `cpu/rollout = 0.3`
- **Called N times** → `time/rollout:total`, `time/rollout:count`, `cpu/rollout:total`, `cpu/rollout:count`, plus user-requested aggs

Built-in aggregations: `"mean"`, `"max"`, `"min"`. Custom lambdas supported:

```python
@profiled("rollout", agg=[("p95", lambda durs: sorted(durs)[int(len(durs) * 0.95)])])
async def do_rollout(...): ...
```

Aggregations are applied to both `time/` and `cpu/` independently.

### Graceful degradation

When `@profiled` fires outside a `prof.measure()` window (e.g., in tests), it logs at DEBUG level but does not record. No error, no data loss. `prof.phase()` is also a no-op outside `prof.measure()`.

### Validation

Invalid `agg` specs are caught at **decoration time**, not at scope exit.

## Integration into RL training loop

This PR integrates the profiler into `do_sync_training()`:

- `@profiled` decorators on `do_group_rollout_and_filter_constant_reward` (with `agg=["mean", "max"]`), `run_evaluations_parallel`, `prepare_minibatch`, and `train_step`
- `prof.phase()` wrapping the three main phases: eval, sampling, train
- Replaces the manual `t_start = time.time()` / `metrics["time/total"]` pattern

## Design rationale: relationship to existing timing systems

The cookbook already has two timing mechanisms. This PR adds a third that fills a distinct gap:

| System | Purpose | Output |
|---|---|---|
| `@scope` (trace.py) | Visual timeline of a single iteration | Chrome Trace → Perfetto |
| `timed()` (misc_utils.py) | Simple wall-clock for dynamic keys | `metrics[f"time/{key}"]` |
| **`@profiled` + `prof.phase()` (this PR)** | **Per-iteration aggregated training metrics** | **dict → wandb** |

These systems are **decoupled by design**. They serve different audiences and lifecycles:

- **`@profiled`** → researcher watching wandb during training. Aggregates per iteration, resets.
- **`@scope`** → engineer debugging a slow iteration in Perfetto. Records every span.
- **`timed()`** → simple inline timing for dynamic keys (`timed(f"train/substep_{i}", metrics)`).

Stacking `@profiled @scope` on the same function gives both wandb metrics and Perfetto traces. `timed()` coexists for dynamic keys that `@profiled` can't handle.

The profiler is designed with **zero external dependencies** (stdlib only: `time`, `contextvars`, `threading`), making it easy to integrate with external tracing systems via adapters.

## Future directions

- **`prof.timed("key")` context manager** — a drop-in replacement for `timed()` that records into the active measurement window instead of requiring a separate metrics dict. Would allow gradual migration of existing `timed()` call sites.
- **Integration with `do_async_training()` and `do_sync_training_with_stream_minibatch()`** — this PR only integrates into `do_sync_training()`. The other two training modes would benefit from the same phase-based profiling.
- **Adapter hooks for external tracing** — an optional callback mechanism so that `@profiled` calls can also emit spans to external tracing systems (e.g., OpenTelemetry) without coupling the profiler to any specific backend.
- **Memory profiling** — tracking peak memory usage per phase alongside wall-clock and CPU time.
- **Profiling dashboard utilities** — helpers to visualize profiling data from `metrics.jsonl` as flame charts or phase breakdowns, complementing the Perfetto timeline from `@scope`.

## What stays the same

- **`timed()` context manager** — not deprecated, still the right tool for dynamic keys like `timed(f"train/substep_{i}_mb_{j}", metrics)`.
- **`@scope` decorator** — untouched. Perfetto tracing and profiling remain decoupled.
- **`ml_log.Logger` / `MultiplexLogger`** — no changes.
- Existing code works as-is; adoption is incremental.

## Concurrency safety

| Scenario | Behavior |
|---|---|
| Single async task | Safe — cooperative scheduling |
| `asyncio.gather()` bare coroutines | Safe — shared context |
| `asyncio.create_task()` children | Safe — inherit same window. Parent must await before window exits. |
| `threading.Thread` / `ThreadPoolExecutor` | Isolated — child threads don't inherit ContextVar, decorator no-ops |
| `asyncio.to_thread()` | Safe — per-scope `threading.Lock` protects concurrent writes |
| Multi-process | Not applicable |

## New / modified files

- `tinker_cookbook/utils/profiling.py` — implementation (`Profiler`, `profiled`, `phase`)
- `tinker_cookbook/utils/profiling_test.py` — 56 tests across 9 test classes
- `tinker_cookbook/rl/train.py` — integration into `do_sync_training()`
- `docs/profiling-design.md` — design rationale

## Test plan

- [x] 56 unit tests passing (`uv run pytest tinker_cookbook/utils/profiling_test.py`)
- [x] Integrated into RL training and confirmed profiling metrics appear in wandb

### DIFF
--- a/tinker_cookbook/rl/train.py
+++ b/tinker_cookbook/rl/train.py
@@ -56,6 +56,7 @@ from tinker_cookbook.rl.types import (
 from tinker_cookbook.tokenizer_utils import Tokenizer
 from tinker_cookbook.utils import logtree, ml_log
 from tinker_cookbook.utils.misc_utils import safezip, split_list, timed
+from tinker_cookbook.utils.profiling import Profiler, profiled
 from tinker_cookbook.utils.trace import scope, trace_init, update_scope_context
 
 logger = logging.getLogger(__name__)
@@ -248,6 +249,7 @@ def _training_logprobs_from_fwd_bwd(
     return [output["logprobs"].to_torch() for output in fwd_bwd_result.loss_fn_outputs]
 
 
+@profiled("train_step")
 @scope
 async def train_step(
     data_D: List[tinker.Datum],
@@ -458,6 +460,7 @@ async def run_single_evaluation(
         return eval_metrics
 
 
+@profiled("run_evals")
 @scope
 async def run_evaluations_parallel(
     evaluators: list[SamplingClientEvaluator],
@@ -887,6 +890,7 @@ async def do_async_training(
     )
 
 
+@profiled("rollout", agg=["mean", "max"])
 @scope
 async def save_checkpoint_and_get_sampling_client(
     training_client: tinker.TrainingClient,
@@ -912,6 +916,7 @@ async def save_checkpoint_and_get_sampling_client(
             return await training_client.save_weights_and_get_sampling_client_async(), metrics
 
 
+@profiled("prepare_minibatch")
 @scope
 async def prepare_minibatch(
     env_group_builders_P: Sequence[EnvGroupBuilder],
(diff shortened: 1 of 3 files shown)


YujiaBao:
I looked at `codetiming`, OpenTelemetry, `prometheus_client`, and `pyinstrument` — none cover the combination of requirements here:

1. **Async-safe contextvars scoping** — only OTel does this well
2. **Hierarchical phase nesting → flat `/`-separated keys** — OTel has hierarchy but outputs spans, not flat dicts
3. **Per-iteration aggregation + reset to a plain dict** (for wandb) — `codetiming` has aggregation but no async safety or hierarchy
4. **Graceful no-op outside a measurement window** — no library models this concept

The closest candidate is OpenTelemetry, but it's designed for export to tracing backends, not in-process dict output. You'd need a custom in-memory exporter, a custom aggregator, and a decorator wrapper — more glue code than the ~450-line stdlib-only implementation here, plus a heavy dependency tree. `codetiming` has nice aggregation (mean/max/min/stdev) but no async support, no contextvars, and no hierarchy.

Given the implementation is stdlib-only (`time`, `contextvars`, `threading`) and well-tested (56 tests), hand-rolling seemed like the simpler and more maintainable choice.

joschu (commented):
As a minor ergonomics thing --

Can we cut down from two objects to 1?

```
prof = Profiler()
...
with prof.measure() as profile:
	...
```

How about

```
with create_profile() as profile:
	...
```

joschu (approved):

YujiaBao:
Replaced by https://github.com/thinking-machines-lab/tinker-cookbook/pull/453

PR #423, opened 2026-03-11
Add pickle support to ChromaTool and VerifiersEnvGroupBuilder

YujiaBao:
## Summary

Follow-up to #422 (Renderer pickle support). Fixes the remaining two non-pickleable `EnvGroupBuilder` implementations by excluding unpickleable objects from pickle state and lazily reconstructing them.

- **ChromaTool:** Exclude async clients (`_chroma_client`, `_gemini_client`) via `__getstate__`, lazily reconnect in `_ensure_clients()` on first `search()` call using stored connection params
- **VerifiersEnvGroupBuilder:** Exclude `vf.Environment` via `__getstate__`, restore from existing `get_vf_env()` context variable on unpickle

## Design

Both fixes use `__getstate__`/`__setstate__` — drop non-pickleable objects on serialization, recover them on deserialization. No changes to constructor signatures, Dataset classes, or DatasetBuilder classes.

**ChromaTool:**
- `_chroma_client` and `_gemini_client` are typed as `Optional` (honest about runtime state after unpickle)
- `_ensure_clients()` returns a `tuple[AsyncClientAPI, genai.Client]` — callers get guaranteed non-optional references
- `_get_embeddings_with_retry` and `_query_chroma_with_retry` take clients as explicit parameters (passed from `_ensure_clients()`) instead of reading nullable `self` fields — type-safe by construction
- `build()` stores `chroma_host` / `chroma_port` on the instance for reconnection after deserialization
- `_ensure_clients()` is called outside `_CONNECTION_SEMAPHORE` so client creation doesn't block behind queued queries

**VerifiersEnvGroupBuilder:**
- `__setstate__` resolves `vf_env` from the `_vf_env_ctx` context variable (already maintained by the verifiers integration), then assigns all fields explicitly
- `self.vf_env` is only assigned after the None check, so pyright correctly infers it as `vf.Environment` (non-optional)
- Raises `RuntimeError` with actionable guidance if `vf_env` cannot be recovered

## Known limitation: VerifiersEnvGroupBuilder cross-process

In cross-process scenarios (ProcessPoolExecutor, Ray), unpickling raises `RuntimeError` because the `_vf_env_ctx` context variable won't be set in the worker process.

This is not a regression — the verifiers integration is fundamentally single-process today. The `custom_do_group_rollout` in `train.py` is a closure that captures shared state (`TinkerAsyncOpenAIClient`, `Renderer`) and is monkey-patched onto the training loop — none of that survives cross-process serialization. Full cross-process support would require the worker to call `set_vf_env(vf.load_environment(...))` before unpickling builders.

Our `__getstate__`/`__setstate__` makes the builder satisfy the pickleable `EnvGroupBuilder` contract and works correctly in same-process scenarios (async, threading).

## Test plan

- [x] `uv run pytest tinker_cookbook/tests/test_pickle_builders.py -v` — 24 passed, 2 skipped (chromadb/verifiers not in base env)
- [x] `uv run pytest tinker_cookbook/tests/test_renderers.py` — existing renderer tests pass
- [x] `uv run pytest tinker_cookbook/tests/test_harbor_tools.py` — existing harbor tests pass
- [x] Pyright — zero new errors (only pre-existing `reportMissingImports` from optional deps)

### DIFF
--- a/tinker_cookbook/recipes/search_tool/tools.py
+++ b/tinker_cookbook/recipes/search_tool/tools.py
@@ -66,10 +66,10 @@ class RetrievalConfig:
 class ChromaTool:
     """Search tool using ChromaDB + Gemini embeddings.
 
-    Pickle support: Non-pickleable async clients (``_chroma_client``, ``_gemini_client``)
-    are excluded from pickle state and lazily reconnected on first use after deserialization.
-    Connection params (``_chroma_host``, ``_chroma_port``) must be set via ``build()`` for
-    this to work.
+    Pickle support: async clients are not pickleable (network connections).
+    ``__getstate__`` excludes them; ``_ensure_clients()`` lazily reconnects
+    before first use after deserialization. Requires ``build()`` so that
+    connection params (host, port) are available for reconnection.
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


kennyyu (approved):
Thank you!

PR #425, opened 2026-03-12
Add pluggable rollout executor via concurrent.futures.Executor

YujiaBao:
## Summary

Allow offloading group rollouts to separate processes, Ray actors, or custom cluster dispatchers by passing a standard `concurrent.futures.Executor`. Uses a `ContextVar` for transparent injection — all call sites (training, evaluation, distillation) benefit automatically with zero code changes.

## Design

**Standard `concurrent.futures.Executor` interface** — no custom abstractions. `ProcessPoolExecutor` works out of the box, and any `Executor` implementation (Ray, Dask, custom dispatchers) is supported.

**Transparent injection via `ContextVar`** — set once via `set_rollout_executor()`, applies to all rollout call sites automatically. No need to thread the executor through Config, training loops, or evaluators. No "which component supports it?" confusion.

**Zero overhead when unused** — when no executor is set (default), the existing async codepath runs unchanged with no indirection.

**How it works:**
- `do_group_rollout_and_filter_constant_reward` checks the `_rollout_executor` ContextVar
- When set: bundles inputs into a pickleable `_RolloutTask`, dispatches via `loop.run_in_executor(executor, _run_rollout_sync, task)`
- When not set: calls the existing async implementation directly
- Workers run `asyncio.run()` in a fresh event loop; `SamplingClient` reconstructs via shadow holders; tokenizers load from `@cache`

## Usage examples

```python
from concurrent.futures import ProcessPoolExecutor

# Default — asyncio in-process, zero change from today:
await main(cfg)

# Multi-process — standard library, no wrappers needed:
await main(cfg, rollout_executor=ProcessPoolExecutor(max_workers=4))
```

**Ray compatibility** — wrap Ray actors as a `concurrent.futures.Executor`:

```python
from concurrent.futures import Executor, Future
import ray

class RayExecutor(Executor):
    def __init__(self, num_actors=4):
        self._actors = [RolloutActor.remote() for _ in range(num_actors)]
        self._next = 0

    def submit(self, fn, /, *args, **kwargs):
        actor = self._actors[self._next % len(self._actors)]
        self._next += 1
        future = Future()
        ref = actor.run.remote(fn, *args)
        threading.Thread(target=lambda: future.set_result(ray.get(ref)), daemon=True).start()
        return future

# Use it:
await main(cfg, rollout_executor=RayExecutor(num_actors=8))
```

**Custom cluster dispatchers** — any system that can run `fn(*args)` on remote workers wraps as an `Executor` the same way. Pass it to `main()` and rollouts run on the cluster with zero training code changes.

## What's covered

The executor applies automatically to all rollout dispatch sites:
- RL training (sync, stream minibatch, async modes)
- Distillation rollouts
- Evaluation rollouts (`RLTestSetEvaluator`)
- Any future code that calls `do_group_rollout_and_filter_constant_reward`

## Changes

- **`tinker_cookbook/rl/train.py`:** Add `_rollout_executor` ContextVar + helpers, `_RolloutTask` dataclass, `_run_rollout_sync()` entry point. Split `do_group_rollout_and_filter_constant_reward` into dispatch wrapper + `_impl`. Add `rollout_executor` param to `main()`.
- **`tinker_cookbook/rl/metric_util.py`:** `RLTestSetEvaluator.__call__` routes through `do_group_rollout_and_filter_constant_reward` when executor is set, so evaluation rollouts are also offloaded.
- **`tinker_cookbook/rl/builder_pickle_test.py`:** Tests for `_RolloutTask` pickle roundtrip and ContextVar set/get.

## Test plan

- [x] `uv run pytest tinker_cookbook/rl/builder_pickle_test.py -v` — 7 tests pass (including 2 new)
- [x] `uv run pytest tinker_cookbook/renderers/renderer_pickle_test.py` — 15 existing renderer tests pass
- [x] Pyright — zero new errors

### DIFF
--- a/tinker_cookbook/rl/train.py
+++ b/tinker_cookbook/rl/train.py
@@ -43,6 +43,7 @@ from tinker_cookbook.rl.rollout_logging import (
     write_rollout_summaries_jsonl_from_groups,
 )
 from tinker_cookbook.rl.rollouts import (
+    do_group_rollout,  # noqa: F401 — re-exported for verifiers monkey-patching
     do_group_rollout_and_filter_constant_reward,
     set_rollout_executor,
 )


tinker_cookbook/rl/rollouts.py
```diff
@@ -110,3 +117,119 @@ async def do_group_rollout(
...
+    )
+
+
+@scope
+async def do_group_rollout_and_filter_constant_reward(
+    sampling_client: tinker.SamplingClient,
+    env_group_builder: EnvGroupBuilder,
+    max_tokens: int,
+    temperature: float,
+    do_remove_constant_reward_groups: bool,
```
kennyyu, line 185: I suspect other folks may want to make the filtering logic customizable and use different filtering logic (e.g. pass a `Callable[[... group of rollouts], bool]`) -- something to think about for the future!

tinker_cookbook/rl/rollouts.py
```diff
@@ -110,3 +117,119 @@ async def do_group_rollout(
...
+    enable_logging: bool = True,
+) -> TrajectoryGroup | None:
+    policy = TinkerTokenCompleter(sampling_client, max_tokens=max_tokens, temperature=temperature)
+
+    with logtree.optional_enable_logging(enable_logging):
```
kennyyu, line 228: How does logtree work across different processes? If multiple processes are on the same machine, will logtree work correctly? (e.g. they all separately append to the same file)

If this is across multiple machines (e.g. with ray), logtree will probably not work as-is, or we'll need additional tooling to grab the logtree logs across the different nodes (something for the future)

tinker_cookbook/rl/rollouts.py
YujiaBao: Good idea\! The current filtering logic (`all_same(rewards)`) is simple enough that it's inline, but a pluggable filter callable would be a clean extension if more filtering strategies emerge. Noted for the future.

tinker_cookbook/rl/rollouts.py
YujiaBao: Good question. logtree state is per-process and not shared — worker processes don't write to the main process's logtree files. This is safe (no file contention) but means structured logs from inside worker rollouts are lost.

In practice this is fine: the training loop already controls logging via `enable_logging` (only `num_groups_to_log` groups get logged), and the actual training metrics (rewards, episode lengths, etc.) are all in the returned `TrajectoryGroup` dataclass — they survive the pickle transfer back to the main process and get logged there.

For Ray across multiple machines, you're right that we'd need additional tooling to collect logtree logs from remote nodes. That's a good future enhancement if there's demand.

PR #427, opened 2026-03-12
Add max_steps to training loops and simplify smoke test harness

YujiaBao:
## Summary

Add `max_steps: int | None = None` to all training loop Configs, wire it through every recipe CLIConfig, and simplify the smoke test harness to use it instead of pattern-based step detection.

### 1. Training loop changes (4 files)

| Loop | File | Implementation |
|---|---|---|
| Supervised | `supervised/train.py` | Cap nested epoch/batch loop with early break |
| RL | `rl/train.py` | Cap `end_batch` in `main()` — applies to all 3 execution modes (sync, async, stream_minibatch) |
| DPO | `preference/train_dpo.py` | Cap nested epoch/batch loop with early break |
| Distillation | `distillation/train_on_policy.py` | Add `max_steps` alongside deprecated `max_step` alias with deprecation warning |

`max_steps=None` (default) preserves existing behavior — no breaking change.

### 2. Recipe CLIConfig wiring (21 files)

Every recipe CLIConfig now has `max_steps: int | None = None` and forwards it to the training Config. This allows `max_steps` to be passed as a CLI argument: `uv run python -m tinker_cookbook.recipes.chat_sl.train max_steps=2`

### 3. Simplified smoke test harness

**Before**: `run_recipe()` launched the recipe, scanned stdout for regex patterns (`Step 1`, `Sampling batch 1`), killed the process with SIGTERM, used `select()` for non-blocking I/O.

**After**: `run_recipe()` appends `max_steps=2` to CLI args, waits for natural exit, checks `returncode == 0`.

Removed:
- Regex step detection patterns
- `os.killpg()` / `signal.SIGTERM` kill logic
- `process_group=0`
- `eval_every=0` overrides in smoke tests

Kept:
- Real-time output streaming (for CI debuggability)
- `select()` with 5s timeout for non-blocking reads
- Timeout enforcement (for cold start protection)

### Backward compatibility

- `max_steps=None` is the default everywhere — zero behavior change for existing users
- Distillation's `max_step` still works but emits a `DeprecationWarning`; passing both `max_step` and `max_steps` raises `ValueError`

## Smoke test CI results

**All 11 recipes passing**: [latest run](https://github.com/YujiaBao/tinker-cookbook/actions/runs/23021803009)

## Test plan
- [x] `pytest tinker_cookbook/` — 342 passed, 39 skipped
- [x] `ruff check` — all passed
- [x] `pyright` — 38 pre-existing errors, 0 new
- [x] [All 11 smoke tests green on CI](https://github.com/YujiaBao/tinker-cookbook/actions/runs/23021803009)

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -2,7 +2,6 @@
 
 import os
 import select
-import signal
 import subprocess
 import time
 
@@ -50,11 +49,7 @@ def run_recipe(
         while True:
             elapsed = time.monotonic() - start_time
             if elapsed >= timeout:
-                # Kill on timeout
-                try:
-                    os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
-                except (OSError, ProcessLookupError):
-                    proc.terminate()
+                proc.terminate()
                 proc.wait(timeout=10)
                 last_lines = "\n".join(output_lines[-30:])
                 pytest.fail(


tinker_cookbook/distillation/train_on_policy.py
```diff
@@ -164,6 +164,8 @@ class Config:
     load_checkpoint_path: str | None = None
 
     # Maximum number of training steps. If None, train on the full dataset.
+    max_steps: int | None = None
+    # Deprecated alias for max_steps. Use max_steps instead.
     max_step: int | None = None
```
dphuang2, line 169: can we just delete this instead of keeping it here, can do in another PR if you want

dphuang2 (approved):

tinker_cookbook/distillation/train_on_policy.py
YujiaBao: I would prefer to keep it as an alias for now so that we maintain backward compatible.

PR #428, opened 2026-03-12
Improve logtree json files and improve documentation

joschu:
## Summary
- add structured `data` payloads to logtree JSON so conversation content can be consumed without parsing raw HTML
- omit raw HTML string children from JSON when structured data is present, and emit DataFrame tables as structured nodes
- simplify RL rollout logging scopes and add documentation for `*_logtree.json` outputs
- add canonical `message_to_jsonable()` serialization for logtree conversation exports, preserving message metadata like `tool_calls`
- keep `logtree.log_formatter()` backward-compatible with legacy formatters that do not define `to_data()`

## Follow-up fixes
- add regression coverage for structured conversation export and legacy formatter compatibility
- fix pyright typing around optional `to_data()` lookup
- format the follow-up type-check fix to satisfy pre-commit

## Verification
- `uv run pytest tinker_cookbook/utils/logtree_test.py -q`
- `uv run --all-extras pyright tinker_cookbook`
- `pre-commit run --all-files`


YujiaBao:
Heads up: main renamed `tinker_cookbook/tests/test_logtree.py` → `tinker_cookbook/utils/logtree_test.py` in #409. A `git merge main` should resolve it cleanly — git's rename detection carries the new tests to the right file.

YujiaBao (approved):

PR #429, opened 2026-03-12
Use numeric step name for final checkpoints

YujiaBao:
## Summary

- Final checkpoints now save under their numeric step name (e.g. `000030`) **in addition to** `"final"`, so the step count is visible directly in tinker:// paths
- Addresses feedback that `"final"` loses information about how many steps produced the checkpoint, making it harder to reason about runs and resume training
- Also fixes DPO final checkpoint TTL bug (`config.ttl_seconds` → `None`) and adds explicit `ttl_seconds=None` to distillation final checkpoint

### Changed files (6 training loops)
- `tinker_cookbook/rl/train.py`
- `tinker_cookbook/supervised/train.py`
- `tinker_cookbook/preference/train_dpo.py`
- `tinker_cookbook/recipes/sl_loop.py`
- `tinker_cookbook/recipes/rl_loop.py`
- `tinker_cookbook/distillation/train_on_policy.py`

## Test plan
- [x] `uv run pytest tinker_cookbook/` — 345 passed, 39 skipped

### DIFF
--- a/tinker_cookbook/distillation/train_on_policy.py
+++ b/tinker_cookbook/distillation/train_on_policy.py
@@ -471,7 +471,7 @@ async def main(
             loop_state={"batch": num_batches},
             ttl_seconds=None,
         )
-        # Also save with the "final" tag
+        # Also save under the name "final"
         _ = await checkpoint_utils.save_checkpoint_async(
             training_client=training_client,
             name="final",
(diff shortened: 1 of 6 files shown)


dphuang2 (commented):
should we route the name generation into a helper function that has named parameters? like `generate_checkpoint_name(epoch=None,step=None,batch=None)`?

Or do you think just a number is fine and people can use `user_metadata` for that sort of bookkeeping?

PR #436, opened 2026-03-13
Log command invocation in logs.log

joschu:
## Summary
- log the current command line invocation each time file logging is configured
- make logs.log append mode explicit and close replaced handlers during reconfiguration
- add a regression test that verifies repeated logger setup appends and records both invocations

## Testing
- pytest -q tinker_cookbook/tests/test_ml_log.py

### DIFF
--- a/tinker_cookbook/tests/test_ml_log.py
+++ b/tinker_cookbook/tests/test_ml_log.py
@@ -1,0 +1,48 @@
+import logging
+import runpy
+import shlex
+import sys
+from pathlib import Path
+from unittest.mock import patch
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[2]))
+
+configure_logging_module = runpy.run_path(
+    str(Path(__file__).resolve().parents[1] / "utils" / "ml_log.py")
+)["configure_logging_module"]
+
+
+def _flush_root_handlers() -> None:
+    for handler in logging.getLogger().handlers:
+        handler.flush()
+
+
+def test_configure_logging_module_logs_invocation_and_appends(tmp_path):
+    log_path = tmp_path / "logs.log"
+
+    argv_first = ["python", "train.py", "--log-path", str(tmp_path), "--run-name", "first run"]
+    with patch.object(sys, "argv", argv_first):
+        root_logger = configure_logging_module(str(log_path))
+        root_logger.info("first message")
+        _flush_root_handlers()
+
+    first_contents = log_path.read_text()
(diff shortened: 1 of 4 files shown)


tinker_cookbook/utils/ml_log_test.py
```diff
@@ -0,0 +1,48 @@
+import logging
```
YujiaBao, line 1: Nit: the repo recently moved to colocated `<module>_test.py` files (see `logtree_test.py`, `trace_test.py` nearby) — this would fit better as `tinker_cookbook/utils/ml_log_test.py`.

tinker_cookbook/tests/test_ml_log.py
```diff
@@ -0,0 +1,48 @@
...
+import shlex
+import sys
+from pathlib import Path
+from unittest.mock import patch
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[2]))
+
+configure_logging_module = runpy.run_path(
+    str(Path(__file__).resolve().parents[1] / "utils" / "ml_log.py")
+)["configure_logging_module"]
```
YujiaBao: The `runpy.run_path` + `sys.path.insert` approach is unnecessary — `configure_logging_module` is a regular importable function. A direct import is simpler and avoids re-executing the module as a script:

```python
from tinker_cookbook.utils.ml_log import configure_logging_module
```

tinker_cookbook/utils/ml_log.py
```diff
@@ -506,14 +515,17 @@ def format(self, record: logging.LogRecord) -> str:
...
     # Force override like basicConfig(..., force=True)
     root = logging.getLogger()
     root.setLevel(level)
-    root.handlers.clear()
+    for handler in root.handlers[:]:
+        root.removeHandler(handler)
+        handler.close()
     root.addHandler(console_handler)
     root.addHandler(file_handler)
+    logger.info("Command line invocation: %s", _get_command_line_invocation())
```
YujiaBao: Minor: `logger` here is the module-level logger (`logging.getLogger(__name__)`), but the function configures and returns `root`. Using `root.info(...)` would be more self-consistent:

```python
root.info("Command line invocation: %s", _get_command_line_invocation())
```

YujiaBao (changes requested):
left a few comments on the test file (location, import style) and a minor consistency note on `logger` vs `root`.

YujiaBao (approved):

PR #439, opened 2026-03-13
Add tag-based versioning (hatch-vcs) and nightly builds

YujiaBao:
## Summary
- Switch from hard-coded `version = "0.1.0"` in pyproject.toml to **hatch-vcs**, which derives the version automatically from git tags
- `tinker_cookbook.__version__` reads from build-generated `_version.py` (with `importlib.metadata` fallback, and a safe default for raw clones)
- Update `publish-pypi` workflow: triggers on semver tag push (`v1.2.3`), verifies built version matches the tag, runs a smoke test before publishing
- Add **nightly** workflow: triggers after scheduled smoke tests pass, builds the package, and publishes a rolling GitHub pre-release

## What changed

| File | Change |
|------|--------|
| `pyproject.toml` | `version = "0.1.0"` → `dynamic = ["version"]`, added `hatch-vcs` with `tag-pattern` |
| `tinker_cookbook/__init__.py` | Import from `_version.py` with layered fallbacks |
| `.github/workflows/publish-pypi.yaml` | Tag-triggered with version verification and smoke test |
| `.github/workflows/nightly.yaml` | New — build + rolling GitHub Release, gated on smoke tests |
| `.gitignore` | Added `tinker_cookbook/_version.py` (generated file) |

## Release flow (going forward)
```bash
# To release a new version:
git tag v0.2.0
git push origin v0.2.0
# → publish-pypi workflow triggers → verifies version → publishes to PyPI
```

## Nightly builds
The nightly workflow triggers automatically after the daily `smoke-test-recipes` run (6am UTC) — only when smoke tests pass and were triggered by schedule (not manual dispatch). Can also be triggered manually via `workflow_dispatch`.

Latest nightly is always available at:
```
https://github.com/thinking-machines-lab/tinker-cookbook/releases/tag/nightly
```

Install directly:
```bash
pip install 'tinker_cookbook @ https://github.com/thinking-machines-lab/tinker-cookbook/releases/download/nightly/tinker_cookbook-<version>-py3-none-any.whl'
```

Build artifacts are also retained for 7 days in the Actions tab.

## Safety measures
- Tag pattern restricted to `v[0-9]+.[0-9]+.[0-9]+` (no accidental triggers from arbitrary `v*` tags)
- `hatch-vcs` configured with `tag-pattern` so the `nightly` tag doesn't interfere with version calculation
- Built version is verified against the git tag before publishing to PyPI
- Import smoke test runs before publishing
- Manual dispatch requires an explicit tag input
- Nightly only triggers after scheduled (not manual) smoke test success

## Post-merge steps
- [x] ~~Tag the commit that was published as v0.1.0 to PyPI~~ (done: `git tag v0.1.0 0747ba0`)
- No additional secrets needed — nightly uses `GITHUB_TOKEN` (automatic)

## Test plan
- [x] `uv build` produces correct versioned artifacts (verified locally)
- [x] `import tinker_cookbook; print(tinker_cookbook.__version__)` works
- [x] `tag-pattern` correctly ignores `nightly` tag (verified locally)
- [ ] CI passes
- [ ] After merge: verify nightly workflow triggers after smoke tests

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -58,16 +58,16 @@ jobs:
             --prerelease \
             --title "Nightly Build (${{ steps.version.outputs.version }})" \
             --notes "$(cat <<EOF
-          Automated nightly build from \`main\` at $(date -u '+%Y-%m-%d %H:%M UTC').
+Automated nightly build from \`main\` at $(date -u '+%Y-%m-%d %H:%M UTC').
 
-          **Version:** \`${{ steps.version.outputs.version }}\`
-          **Commit:** [\`${GITHUB_SHA::8}\`](https://github.com/${{ github.repository }}/commit/${GITHUB_SHA})
+**Version:** \`${{ steps.version.outputs.version }}\`
+**Commit:** [\`${GITHUB_SHA::8}\`](https://github.com/${{ github.repository }}/commit/${GITHUB_SHA})
 
-          ### Install
-          \`\`\`bash
-          pip install 'tinker_cookbook @ https://github.com/${{ github.repository }}/releases/download/nightly/tinker_cookbook-${{ steps.version.outputs.version }}-py3-none-any.whl'
-          \`\`\`
-          EOF
+### Install
+\`\`\`bash
+pip install 'tinker_cookbook @ https://github.com/${{ github.repository }}/releases/download/nightly/tinker_cookbook-${{ steps.version.outputs.version }}-py3-none-any.whl'
+\`\`\`
+EOF
           )"
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}


tinker_cookbook/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Tinker Cookbook: post-training algorithms using the Tinker API."""
+
+__version__ = "0.1.0"
```
dphuang2: doesn't this need to be automated instead of hard-coded? Also, can we use git tags to do versioning? (e.g. push new git tag and gh workflow triggers release?)

tinker_cookbook/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Tinker Cookbook: post-training algorithms using the Tinker API."""
+
+from importlib.metadata import version
+
+__version__ = version("tinker_cookbook")
```
dphuang2: doesn't this come from pyproject.toml, which is also hard-coded?

PR #451, opened 2026-03-13
Consolidate streaming & response normalization into base Renderer

YujiaBao:
## Summary
- Move shared streaming infrastructure (`parse_response_streaming`, `_normalize_response_tokens`, `_parse_response_for_streaming`, `_end_message_token`) into the base `Renderer` class, gated by a `supports_streaming` class attribute
- Eliminate duplicated streaming boilerplate from KimiK2, KimiK25, and Qwen3 renderers
- Add streaming support to Qwen3.5 and DeepSeek V3 (thinking and non-thinking)
- Refactor Qwen3.5 `parse_response` to extract `_normalize_response_tokens` and `_postprocess_parsed_message` hooks, enabling streaming via base class inheritance
- Add KimiK25 to the cross-renderer roundtrip test parametrization

### Base class hooks

Three new private hooks on `Renderer`, all with safe defaults:

| Hook | Default | Purpose |
|------|---------|---------|
| `_normalize_response_tokens(response)` | Identity | Restore prefilled tokens (e.g. `<think>`) that the sampler strips |
| `_parse_response_for_streaming(response)` | Delegates to `parse_response` | Parse without short-circuiting on missing stop token |
| `_end_message_token` | `NotImplementedError` | EOS token ID for the streaming parser |

The concrete `parse_response_streaming` on the base class is gated by `supports_streaming = True` (default `False`), so existing renderers (Llama3, GptOss, RoleColon) are unaffected and still raise `NotImplementedError`.

### Normalization flow

Renderers that prefill tokens in `build_generation_prompt` (KimiK25 prefills `<think>`, Qwen3.5 prefills `<think>\n`, DeepSeek prefills `<think>`) need to restore those tokens before parsing. The normalization call site is:

- **`parse_response`**: each renderer calls `self._normalize_response_tokens(response)` at the top, so subclass overrides are dispatched via polymorphism (e.g. `KimiK2.parse_response` → `KimiK25._normalize_response_tokens`)
- **`parse_response_streaming`**: the base implementation normalizes once before creating the parser. `_parse_response_for_streaming` does NOT normalize again, avoiding double work. All `_normalize_response_tokens` overrides are idempotent as a safety net.

### Per-renderer changes

| Renderer | `supports_streaming` | Overrides removed | New capability |
|----------|---------------------|-------------------|----------------|
| KimiK2 | `True` | `parse_response_streaming` | — |
| KimiK25 | (inherited) | `parse_response`, `parse_response_streaming` | — |
| Qwen3 | `True` | `_normalize_response_tokens` (identity), `parse_response_streaming` | — |
| Qwen3.5 | (inherited) | — | Streaming support |
| DeepSeek V3 | `True` | `parse_response` (ThinkingRenderer) | Streaming support |

### Backward compatibility

- All new methods are `_`-prefixed (private). Adding them is purely additive.
- `parse_response` behavior is unchanged — the identity `_normalize_response_tokens` is a no-op for renderers that don't override it.
- `parse_response_streaming` is gated by `supports_streaming` — renderers that don't opt in still raise `NotImplementedError`.
- Removed overrides are behaviorally equivalent (verified by A/B comparison against main).

## Test plan
- [x] All 338 renderer tests pass (331 original + 7 new)
- [x] Existing KimiK2/K25 streaming tests pass with inherited base implementation
- [x] Existing Qwen3 streaming tests pass with inherited base implementation
- [x] New Qwen3.5 prefill normalization tests (batch + streaming + batch/streaming equivalence + no-op when already present)
- [x] New DeepSeek streaming batch/streaming equivalence tests (thinking, non-thinking, truncated)
- [x] Cross-renderer `test_thinking_generation_parse_correspondence` covers KimiK25
- [x] A/B comparison against main: `parse_response`, `parse_response_streaming`, `build_generation_prompt`, `build_supervised_example` all produce identical outputs for 6 renderers × 5–7 test inputs (74/76 pass; 2 expected diffs from #454's KimiK25 normalization fix)

### DIFF
--- a/tinker_cookbook/renderers/kimi_k25.py
+++ b/tinker_cookbook/renderers/kimi_k25.py
@@ -8,6 +8,7 @@ from tinker_cookbook.image_processing_utils import ImageProcessor
 from tinker_cookbook.renderers.base import (
     ContentPart,
     ImageProcessorProtocol,
+    Message,
     Role,
     ToolSpec,
     image_to_chunk,
--- a/tinker_cookbook/renderers/qwen3_test.py
+++ b/tinker_cookbook/renderers/qwen3_test.py
@@ -24,7 +24,6 @@ from tinker_cookbook.renderers.qwen3_5 import Qwen3_5Renderer
 from tinker_cookbook.tokenizer_utils import get_tokenizer
 
 
-
 def _is_message(obj) -> TypeGuard[Message]:
     """Check if object is a Message dict (TypedDict doesn't support isinstance)."""
     return isinstance(obj, dict) and "role" in obj and "content" in obj


YujiaBao:
Example traces on wandb (you can also toggle over the steps)
image
image

PR #453, opened 2026-03-14
Extend @scope with per-iteration wandb metrics and Gantt charts

YujiaBao:
## Motivation

PR #420 introduced a hand-rolled `Profiler` with `@profiled` decorator for per-iteration timing metrics. Reviewer feedback asked whether we should use an existing library. After researching alternatives (`codetiming`, OpenTelemetry, `pyinstrument`, `prometheus_client`), we found no library covers the combination of async-safe contextvars scoping, per-iteration aggregation to a flat dict, and graceful no-op semantics.

But there's a deeper issue: the cookbook already has `@scope` (trace.py) for Perfetto traces, and `@profiled` would create a parallel system annotating the same functions with the same phase structure. Instead of two systems, this PR **extends `@scope`** to also produce per-iteration timing metrics, file-based span export, and Gantt chart visualization — one annotation, multiple outputs.

### Setup

```python
from tinker_cookbook.utils.trace import scope, scope_span, trace_init, trace_iteration

trace_init(output_file="trace_events.jsonl")  # Perfetto trace + SDK auto-instrumentation
```

`trace_init` auto-instruments Tinker SDK client classes (`TrainingClient`, `SamplingClient`) with `@scope` via class-level patching. All instances — including those from `create_sampling_client()` — are automatically traced.

### Training loop

```python
for i_batch in range(n_batches):
    metrics = {"progress/batch": i_batch, ...}

    with trace_iteration(step=i_batch) as window:
        await run_evals(...)
        await gather_rollouts(...)
        await train_step(...)
        metrics.update(train_step_metrics)

    # Merge timing metrics and log everything in a single call
    metrics.update(window.get_timing_metrics())
    window.write_spans_jsonl(log_path / "timing_spans.jsonl", step=i_batch)
    ml_logger.log_metrics(metrics, step=i_batch)
```

### What gets logged

`trace_iteration` yields an `IterationWindow` that aggregates all `@scope` and `scope_span` spans:

- **Single call** → `time/run_evaluations_parallel = 4.2`
- **Repeated calls** (e.g., 160 concurrent `sample_async`) → `time/sample_async:total`, `time/sample_async:count`, `time/sample_async:mean`, `time/sample_async:max`
- **Total** → `time/total` (wall-clock of the entire iteration)

Span names are flat (function or span name), not hierarchical. If `train_step` calls `forward_backward_async`, both appear as separate top-level keys with inclusive timing.

### Semantic rollout timing with `scope_span`

`scope_span` is a new async context manager for inline named spans — it records to both the Perfetto trace and the IterationWindow without needing to wrap calls in new functions:

```python
async def do_single_rollout(policy, env):
    async with scope_span("env_initial_observation"):
        ob, stop_condition = await env.initial_observation()
    while True:
        async with scope_span("policy_sample"):
            ac_with_logprobs = await policy(ob, stop_condition)
        async with scope_span("env_step"):
            step_result = await env.step(ac_with_logprobs.tokens)
        ...
```

This gives the timeline semantic phases rather than just code structure — you can now see time spent in policy sampling vs environment/grader work vs training.

### File-based output

All timing data is available as files — no wandb dependency needed:

- **`metrics.jsonl`** — `time/*` metrics flow through the standard `ml_logger` (one line per step with both training and timing metrics)
- **`timing_spans.jsonl`** — Raw span records appended per iteration: `{"step": N, "spans": [{"name": ..., "duration": ..., "wall_start": ..., "wall_end": ...}, ...]}`
- **`timing_gantt_NNNNNN.html`** — Optional standalone Plotly Gantt chart saved every N steps (controlled by `Config.span_chart_every`)

When `wandb_project` is configured, timing metrics are automatically included — they flow through `ml_logger.log_metrics()` alongside training metrics, which dispatches to all configured backends (JSON, console, wandb, Neptune, etc.):

```python
# Timing metrics are merged before the single log call
metrics.update(window.get_timing_metrics())
ml_logger.log_metrics(metrics, step=i_batch)  # → metrics.jsonl + wandb + console
```

### Why extend `@scope` instead of a separate system

| Before | After |
|---|---|
| `@scope` → Perfetto only | `@scope` → Perfetto + iteration timing metrics |
| `timed()` → manual metrics dict | Gradually replaced by `@scope` + `trace_iteration` |
| `@profiled` (PR #420) → parallel system | Not needed — subsumed by extended `@scope` |

Single annotation, multiple outputs.

### Caller-controlled logging

`trace_iteration` yields the `IterationWindow` — the caller decides what to do with it. This avoids coupling `trace.py` to any specific logger or backend, and ensures timing + training metrics appear in a single `log_metrics()` call (one line per step in `metrics.jsonl`).

### SDK client auto-instrumentation

`trace_init()` patches SDK client classes at the class level. All instances are automatically traced. The `@scope` decorator no-ops when `_trace_collector is None` and no `IterationWindow` is active, so there's zero overhead when tracing is disabled. Double-wrapping is prevented via a `_scope_instrumented` marker.

### Cross-process future-proofing

`SpanRecord` stores dual timestamps:
- `perf_counter` (high-res, process-local) for accurate duration math
- `time.time()` (cross-process comparable) for Gantt chart positioning

This enables future cross-process span collection (ProcessPoolExecutor, Ray) via `IterationWindow.merge_spans()` without clock alignment issues.

### Concurrency safety

- `IterationWindow.record_span()` is thread-safe (`threading.Lock`)
- `_iteration_window` ContextVar is inherited by child async tasks (`asyncio.gather`, `create_task`)
- Nested `trace_iteration` is supported — inner window is independent from outer
- Exception inside `trace_iteration` / `scope_span` still records partial timing (finally block)

## What stays the same

- `@scope` decorator — unchanged API, still produces Perfetto traces
- `timed()` — not deprecated, still the right tool for dynamic keys like `timed(f"train/substep_{i}_mb_{j}", metrics)`

## New / modified files

- `tinker_cookbook/utils/trace.py` — `IterationWindow`, `trace_iteration()`, `scope_span()`, `scope_span_sync()`, `save_gantt_chart_html()`, `write_spans_jsonl()`, SDK client patching, Gantt chart rendering
- `tinker_cookbook/utils/trace_test.py` — 32 tests covering: scope, iteration window, trace_iteration, scope_span (async/sync/perfetto/noop/exception), write_spans_jsonl, save_gantt_chart_html, SDK instrumentation tripwire
- `tinker_cookbook/rl/train.py` — Integrated `trace_iteration` into `do_sync_training()`, added `span_chart_every` to Config
- `tinker_cookbook/rl/rollouts.py` — Added `scope_span` instrumentation in `do_single_rollout` and `do_group_rollout` for semantic phase timing

## Test plan

- [x] 32 unit tests passing (`uv run pytest tinker_cookbook/utils/trace_test.py`)
- [x] Pyright clean (no new errors)

### DIFF
--- a/tinker_cookbook/utils/trace.py
+++ b/tinker_cookbook/utils/trace.py
@@ -89,7 +89,33 @@ class SpanRecord:
 
 
 class IterationWindow:
-    """Collects span records during a single training iteration for aggregation."""
+    """Collects span records during a single training iteration for aggregation.
+
+    Use with :func:`trace_iteration` to automatically capture all ``@scope`` and
+    ``scope_span`` timings within a training iteration. After the block exits,
+    call :meth:`get_timing_metrics` for a flat dict of timing metrics ready to log.
+
+    Example — GRPO training loop::
+
+        for i_batch in range(n_batches):
+            with trace_iteration(step=i_batch) as window:
+                # All @scope-decorated calls inside this block are recorded
+                await run_evals(sampling_client, ...)
+                trajectory_groups = await gather_rollouts(sampling_client, ...)
+                await train_step(training_client, trajectory_groups, ...)
+
+            # Aggregated metrics: time/total, time/run_evals, time/sample_async:total, ...
+            metrics.update(window.get_timing_metrics())
+
+            # Persist per-span data for post-hoc analysis
+            window.write_spans_jsonl(log_path / "timing_spans.jsonl", step=i_batch)
+
+            # Optional: save a Gantt chart every K steps
+            if i_batch % 10 == 0:
+                save_gantt_chart_html(window, i_batch, log_path / f"gantt_{i_batch}.html")
+
+            ml_logger.log_metrics(metrics, step=i_batch)
+    """
 
     def __init__(self) -> None:
         self.spans: list[SpanRecord] = []


tinker_cookbook/utils/trace.py
```diff
@@ -154,22 +267,76 @@ def _atexit_trace_shutdown():
...
+            "load_state_async",
+            "save_weights_for_sampler_async",
+            "save_weights_and_get_sampling_client_async",
+            "create_sampling_client_async",
+        ],
+        tinker.SamplingClient: [
+            "sample_async",
+            "compute_logprobs_async",
+        ],
+    }
```
danobi: this feels fragile - i think we should design profiling holistically rather than everyone independently coming from different angles

YujiaBao: Fair point on the hardcoded list — that'll go stale if the SDK adds methods. I think the simplest fix is adding a test that checks this list against the SDK's actual methods so we get a failure if they diverge. Alternatively I could introspect the classes at runtime (patch all public `*_async` methods), but that feels less explicit. Wdyt?

Re holistic design — that's actually what this PR is trying to do. Right now we have `@scope` for Perfetto, `@profiled` from #420 for wandb metrics, and `timed()` all annotating the same phases separately. This collapses them into one: `@scope` produces Perfetto + wandb metrics + Gantt charts. The auto-instrumentation is just so SDK calls get traced without needing a second decorator on every method.

YujiaBao: follow up:
  - Added `test_sdk_client_instrumentation_covers_all_async_methods` — introspects `TrainingClient` / `SamplingClient` at test time and asserts every public async method `has _scope_instrumented`. Acts as a tripwire when bumping the tinker dependency.
  - Implementation still uses an explicit list with `hasattr` guards, so removed methods are handled gracefully (skipped). Additions are caught by the test.

tinker_cookbook/utils/trace.py
```diff
@@ -154,22 +267,79 @@ def _atexit_trace_shutdown():
 atexit.register(_atexit_trace_shutdown)
 
 
+def _instrument_sdk_clients() -> None:
+    """Patch Tinker SDK client classes with @scope for automatic tracing."""
+    try:
+        import tinker
```
joschu: this shouldn't happen

tinker_cookbook/utils/trace.py
```diff
@@ -155,13 +302,55 @@ def _atexit_trace_shutdown():
 atexit.register(_atexit_trace_shutdown)
 
 
+def _instrument_sdk_clients() -> None:
```
kennyyu, line 331: I remember I tried doing this when I originally added scope, and for some larger/longer runs, this added a lot of overhead (I don't remember why). In your testing, how was the overhead?

tinker_cookbook/utils/trace.py
```diff
@@ -60,6 +67,146 @@ class ScopeContext:
...
+
+    name: str
+    start_time: float  # seconds (perf_counter, process-local)
+    end_time: float  # seconds (perf_counter, process-local)
+    wall_start: float  # seconds since epoch (time.time, cross-process comparable)
+    wall_end: float  # seconds since epoch (time.time, cross-process comparable)
+
+
+class IterationWindow:
+    """Collects span records during a single training iteration for aggregation."""
```
kennyyu: Would be great to include an example on how to use IterationWindow and the intended usage! (seems the main benefit is aggregations within the window period)

tinker_cookbook/utils/trace.py
```diff
@@ -422,6 +726,87 @@ async def foo(step: int):
...
+    )
+    fig.update_layout(
+        xaxis_title="Time (relative)",
+        yaxis_title="",
+        showlegend=False,
+    )
+    return fig
+
+
+def save_gantt_chart_html(window: IterationWindow, step: int, path: Path | str) -> None:
```
kennyyu, line 782: do you have an example/screenshot of a gantt chart? would be useful to include in the PR!

tinker_cookbook/rl/train.py
```diff
@@ -1199,77 +1207,81 @@ async def do_sync_training(
             "optim/lr": cfg.learning_rate,
             "progress/done_frac": (i_batch + 1) / num_batches,
         }
-        t_start = time.time()
 
-        # Run evaluations
-        if cfg.eval_every > 0 and i_batch % cfg.eval_every == 0:
-            with timed("run_evals", metrics):
+        with trace_iteration(step=i_batch) as window:
```
kennyyu, line 1211: note: iteration window gets a bit tricky to define once we consider async RL :) If you plan to add that to the async RL loop, you might have to define separate iteration windows for the trainer side + sampler/rollout side

tinker_cookbook/utils/trace.py
YujiaBao: Good question — I ran a microbenchmark (10k calls per config):

| Config | Median | p99 |
|--------|--------|-----|
| Baseline (no @scope) | 0.04 µs | 0.08 µs |
| @scope no-op | 0.08–0.13 µs | 0.17–0.21 µs |
| @scope + IterationWindow | 0.42–0.50 µs | 0.71–1.42 µs |
| @scope + Perfetto + Window | 3.25 µs | ~14 µs |

Even at p99 with full tracing, the overhead is ~14 µs per call. SDK calls take 1–60s each, so this is <0.001% of a real call. The hot path when there's no trace collector and no iteration window is just a ContextVar.get() — essentially free.

Do you remember if the overhead you saw was on the Perfetto trace file I/O side (large trace files)? That would be the flusher thread writing to disk, which is separate from the per-call wrapper cost.

<details>
<summary>Benchmark script</summary>

```python
"""Benchmark the per-call overhead of @scope instrumentation.

Measures overhead across four configurations:
  1. Baseline      – direct function call (no @scope)
  2. No-op @scope  – @scope with no trace collector, no iteration window
  3. Window only   – @scope with IterationWindow active (no Perfetto)
  4. Full tracing  – @scope with Perfetto + IterationWindow

Each configuration is tested for both sync and async functions.

Usage:
    uv run scripts/bench_scope_overhead.py
"""

import asyncio
import statistics
import tempfile
import time

from tinker_cookbook.utils.trace import (
    scope,
    trace_init,
    trace_iteration,
    trace_shutdown,
)

N_CALLS = 10_000


# --------------- Sync benchmarks ---------------

def bare_sync():
    """Baseline: no decorator."""
    pass


@scope
def scoped_sync():
    """Decorated with @scope."""
    pass


def bench_sync_baseline() -> list[float]:
    timings = []
    for _ in range(N_CALLS):
        t0 = time.perf_counter()
        bare_sync()
        timings.append(time.perf_counter() - t0)
    return timings


def bench_sync_noop() -> list[float]:
    timings = []
    for _ in range(N_CALLS):
        t0 = time.perf_counter()
        scoped_sync()
        timings.append(time.perf_counter() - t0)
    return timings


def bench_sync_window_only() -> list[float]:
    timings = []
    with trace_iteration(step=0):
        for _ in range(N_CALLS):
            t0 = time.perf_counter()
            scoped_sync()
            timings.append(time.perf_counter() - t0)
    return timings


def bench_sync_full() -> list[float]:
    with tempfile.NamedTemporaryFile(suffix=".jsonl") as f:
        trace_init(output_file=f.name)
        timings = []
        with trace_iteration(step=0):
            for _ in range(N_CALLS):
                t0 = time.perf_counter()
                scoped_sync()
                timings.append(time.perf_counter() - t0)
        trace_shutdown()
    return timings


# --------------- Async benchmarks ---------------

async def bare_async():
    """Baseline: no decorator."""
    pass


@scope
async def scoped_async():
    """Decorated with @scope."""
    pass


async def bench_async_baseline() -> list[float]:
    timings = []
    for _ in range(N_CALLS):
        t0 = time.perf_counter()
        await bare_async()
        timings.append(time.perf_counter() - t0)
    return timings


async def bench_async_noop() -> list[float]:
    timings = []
    for _ in range(N_CALLS):
        t0 = time.perf_counter()
        await scoped_async()
        timings.append(time.perf_counter() - t0)
    return timings


async def bench_async_window_only() -> list[float]:
    timings = []
    with trace_iteration(step=0):
        for _ in range(N_CALLS):
            t0 = time.perf_counter()
            await scoped_async()
            timings.append(time.perf_counter() - t0)
    return timings


async def bench_async_full() -> list[float]:
    with tempfile.NamedTemporaryFile(suffix=".jsonl") as f:
        trace_init(output_file=f.name)
        timings = []
        with trace_iteration(step=0):
            for _ in range(N_CALLS):
                t0 = time.perf_counter()
                await scoped_async()
                timings.append(time.perf_counter() - t0)
        trace_shutdown()
    return timings


# --------------- Report ---------------

def report(label: str, timings: list[float]) -> None:
    med = statistics.median(timings) * 1e6  # µs
    p99 = sorted(timings)[int(len(timings) * 0.99)] * 1e6
    total = sum(timings) * 1e3  # ms
    print(f"  {label:30s}  median={med:8.2f} µs   p99={p99:8.2f} µs   total={total:8.1f} ms")


def main():
    print(f"\n@scope overhead benchmark ({N_CALLS:,} calls per config)\n")

    print("Sync functions:")
    report("baseline (no @scope)", bench_sync_baseline())
    report("@scope no-op", bench_sync_noop())
    report("@scope + IterationWindow", bench_sync_window_only())
    report("@scope + Perfetto + Window", bench_sync_full())

    print("\nAsync functions:")
    report("baseline (no @scope)", asyncio.run(bench_async_baseline()))
    report("@scope no-op", asyncio.run(bench_async_noop()))
    report("@scope + IterationWindow", asyncio.run(bench_async_window_only()))
    report("@scope + Perfetto + Window", asyncio.run(bench_async_full()))

    print("\nContext: SDK calls (sample_async, forward_backward_async) typically take")
    print("1-60 seconds each. Even at p99, the overhead is <0.01% of a real call.\n")


if __name__ == "__main__":
    main()
```

</details>

tinker_cookbook/utils/trace.py
YujiaBao: Added a GRPO training loop example to the `IterationWindow` docstring showing the `trace_iteration` → `get_timing_metrics` → `write_spans_jsonl` → `save_gantt_chart_html` flow.

tinker_cookbook/utils/trace.py
YujiaBao: here's a sample from a synthetic GRPO iteration (16 concurrent sample_async calls, eval, forward_backward, optim_step, checkpoint save).
image

tinker_cookbook/rl/train.py
YujiaBao: Yeah, good call — `trace_iteration` is inherently an iteration-level abstraction, and async RL doesn't have a clean iteration boundary since sampling and training are fully decoupled.

PR #454, opened 2026-03-14
Fix Kimi K2.5 parsing when thinking is prefixed

joschu:
## Summary
- restore the synthetic `<think>` prefill before Kimi K2.5 parses sampled tokens
- apply the same normalization to streaming parsing so thinking/text deltas stay structured
- add regression tests for batch and streaming parse paths

## Testing
- uv run pytest -q tinker_cookbook/tests/test_renderers.py -k "kimi_k25_parse_response or test_eot_parsing"


tinker_cookbook/renderers/renderers_test.py
```diff
@@ -1137,6 +1140,476 @@ def test_eot_parsing(model_name: str, renderer_name: str):
...
+            # Output should not start with </think>
+            output_chunk = rendered.output[0]
+            assert isinstance(output_chunk, tinker.EncodedTextChunk), "Expected EncodedTextChunk"
+            output_str = tokenizer.decode(list(output_chunk.tokens))
+            assert not output_str.startswith("</think>"), (
+                f"Post-tool assistant should not have </think> prefix: {output_str}"
+            )
+
+
+def test_kimi_k2_thinking_stripped_when_no_suffix_messages():
```
YujiaBao: shall we put the tests for kimi k2 thinking under `tinker_cookbook/renderers/kimi_k2_test.py`

joschu:
Attaching the original bug repro used for this PR. This was provided locally as `/Users/joschu/code/tinker-cookbook/k25_think_bug.py` and demonstrates the parser/prefill mismatch directly.

```python
"""MWE: KimiK25Renderer.parse_response drops thinking because <think> is in the prefill.

The kimi_k25 renderer adds <think> as a generation prefill, but parse_response
only sees the response tokens (which start after the prefill). So the response
contains `thinking...</think>answer` without an opening <think> tag.
parse_think_blocks can't match the pattern and returns the full string as content.

Expected: content should be just "2" (with thinking in a ThinkingPart)
Actual:   content is the entire thinking + "</think>2" as a plain string
"""

import asyncio

import tinker
from tinker_cookbook.renderers import get_renderer, get_text_content
from tinker_cookbook.tokenizer_utils import get_tokenizer

tokenizer = get_tokenizer("moonshotai/Kimi-K2.5")
renderer = get_renderer("kimi_k25", tokenizer)

messages = [{"role": "user", "content": "What is 1+1? Reply with just the number."}]


async def main():
    service_client = tinker.ServiceClient()
    sampling_client = service_client.create_sampling_client(
        base_model="moonshotai/Kimi-K2.5"
    )

    model_input = renderer.build_generation_prompt(messages)

    # Show prompt tail — note <think> is in the prefill
    all_tokens = []
    for chunk in model_input.chunks:
        all_tokens.extend(chunk.tokens)
    print("Prompt tail:", repr(tokenizer.decode(all_tokens[-10:])))

    resp = await sampling_client.sample_async(
        model_input,
        num_samples=1,
        sampling_params=tinker.SamplingParams(
            temperature=1.0, max_tokens=256, stop=renderer.get_stop_sequences()
        ),
    )
    seq = resp.sequences[0]

    # Raw response tokens — no opening <think>, only </think>
    raw = tokenizer.decode(seq.tokens)
    print(f"\nRaw response: {repr(raw)}")

    # parse_response — thinking leaks into content
    parsed, success = renderer.parse_response(seq.tokens)
    print(f"\nparse_response success: {success}")
    print(f"content type: {type(parsed['content']).__name__}")
    print(f"content: {repr(parsed['content'])}")
    print(f"\nget_text_content: {repr(get_text_content(parsed))}")


asyncio.run(main())
```

tinker_cookbook/renderers/renderers_test.py
joschu: Moved the K2.5 parse regressions into `tinker_cookbook/renderers/kimi_k25_test.py` and removed them from `renderers_test.py`.

YujiaBao (approved):
One thought for a follow-up: once we have the Qwen 3 streaming parser in, it might be worth restructuring so the prefill normalization lives in the base class rather than each thinking renderer independently (as the behavior should be the same)

PR #463, opened 2026-03-15
Harbor eval standardization

ZitongYang:
- Add standalone Harbor evaluation (eval.py) — run any Harbor dataset through an agent loop and grade results, with per-task trajectory logs and summary files 
  - Generalize task loading: replace terminal-bench-specific launch_terminal_bench.py with load_harbor_tasks(dataset) in harbor_env.py, supporting any Harbor    
  dataset                                                                                                                                                        
  - Add scripts/ with ready-to-run eval and training entry points                                                                                                
  - Include baseline results for Kimi-K2-Thinking on Terminal-Bench 2.0 and SWE-Bench Verified 1.0 (32K context, naive agent harness)

### DIFF
--- a/tinker_cookbook/recipes/harbor_rl/eval.py
+++ b/tinker_cookbook/recipes/harbor_rl/eval.py
@@ -179,7 +179,6 @@ async def run_eval(
         config: Evaluation configuration.
         tasks: List of HarborTask to evaluate.
         sandbox_factory: Factory for creating sandboxes (defaults to Modal).
-        output_path: Base directory for results (default: scripts/results).
 
     Returns:
         List of per-task results.


YujiaBao:
Hey Zitong! The direction here makes sense — moving `load_harbor_tasks()` into `harbor_env.py` is cleaner and more reusable. A few things to address:

**Breaking import:** `tinker_cookbook/recipes/distillation/on_policy_distillation_harbor_multi_turn.py` (line 45) imports `load_terminal_bench_tasks` from the deleted `launch_terminal_bench.py`. The distillation README also references it. This PR will break that code path — those need to be updated to use the new `load_harbor_tasks()`.

**Design suggestions for reusability:**

1. Consider making `dataset` a `CLIConfig` parameter in `train_terminal_bench.py` instead of hardcoding `"terminal-bench-2.0"`. That way the same script works for any Harbor-formatted dataset without code changes.

2. The `-o ~/.cache/harbor/tasks/terminal-bench-2.0` in the download command couples the download path to the loader's expectations. If `harbor` already has a default output location, it'd be better to align with that rather than requiring a specific `-o` flag.

3. Minor: the new `load_harbor_tasks()` doesn't handle the case where the `tasks_dir` doesn't exist (e.g., user forgot to download). A clear error message would help here vs the current `FileNotFoundError` from `iterdir()`.

ZitongYang:
> Hey Zitong! The direction here makes sense — moving `load_harbor_tasks()` into `harbor_env.py` is cleaner and more reusable. A few things to address:
> 
> **Breaking import:** `tinker_cookbook/recipes/distillation/on_policy_distillation_harbor_multi_turn.py` (line 45) imports `load_terminal_bench_tasks` from the deleted `launch_terminal_bench.py`. The distillation README also references it. This PR will break that code path — those need to be updated to use the new `load_harbor_tasks()`.
> 
> **Design suggestions for reusability:**
> 
> 1. Consider making `dataset` a `CLIConfig` parameter in `train_terminal_bench.py` instead of hardcoding `"terminal-bench-2.0"`. That way the same script works for any Harbor-formatted dataset without code changes.
> 2. The `-o ~/.cache/harbor/tasks/terminal-bench-2.0` in the download command couples the download path to the loader's expectations. If `harbor` already has a default output location, it'd be better to align with that rather than requiring a specific `-o` flag.
> 3. Minor: the new `load_harbor_tasks()` doesn't handle the case where the `tasks_dir` doesn't exist (e.g., user forgot to download). A clear error message would help here vs the current `FileNotFoundError` from `iterdir()`.

Thanks! The breaking point is now fixed

tinker_cookbook/recipes/harbor_rl/eval.py
```diff
@@ -0,0 +1,240 @@
...
+        temperature=config.temperature,
+    )
+
+    if config.max_tasks is not None:
+        tasks = random.sample(tasks, min(config.max_tasks, len(tasks)))
+
+    logger.info("Starting evaluation of %d tasks", len(tasks))
+
+    task_results = list(
+        await asyncio.gather(
```
YujiaBao, line 222: shall we run it in a semaphore? right now the concurrency isnot bounded

tinker_cookbook/recipes/harbor_rl/eval.py
```diff
@@ -0,0 +1,240 @@
...
+) -> list[TaskResult]:
+    """Run evaluation on a list of Harbor tasks.
+
+    Results are written to files in <output_path>/<timestamp>/ as each task completes.
+
+    Args:
+        config: Evaluation configuration.
+        tasks: List of HarborTask to evaluate.
+        sandbox_factory: Factory for creating sandboxes (defaults to Modal).
+        output_path: Base directory for results (default: scripts/results).
```
YujiaBao: this is stale

YujiaBao (approved):

tinker_cookbook/recipes/harbor_rl/eval.py
ZitongYang: Thanks for mentioning it. I had that first version, and then I removed it to make the code more hackable and readable, since the intention is to make people build on top of it. For eval this should be fine, since it's essentially always group_size=1. Also eval data is small.

tinker_cookbook/recipes/harbor_rl/eval.py
ZitongYang: Great catch, Thanks!

PR #470, opened 2026-03-16
Add DeepSeek V3.1 experts-only export support

dphuang2:
## Summary
- add a DeepSeek-specific export path that bypasses full-model loading and writes shard-by-shard outputs on CPU for resumable exports
- quantize only routed expert weights to FP8 while keeping non-expert weights in higher precision, and emit the compressed-tensors metadata and DeepSeek custom files needed for serving
- add DeepSeek export coverage for shard/index correctness and resume behavior, and include the optional merge dependencies needed for this path

## Test plan
- [ ] Run `pytest tests/weights/test_export.py`
- [ ] Serve the merged DeepSeek artifacts on stock vLLM and compare eval behavior

### DIFF
--- a/tinker_cookbook/weights/_deepseek.py
+++ b/tinker_cookbook/weights/_deepseek.py
@@ -34,7 +34,11 @@ from safetensors import safe_open
 from safetensors.torch import load_file, save_file
 from transformers import PreTrainedModel
 
-from tinker_cookbook.weights._merge import apply_merged_weight
+from tinker_cookbook.weights._merge import (
+    apply_merged_weight,
+    expand_expert_lora_tensors,
+    merge_lora_matrices,
+)
 
 __all__ = ["is_deepseek_config", "is_deepseek_model", "save_deepseek_model"]
 
@@ -472,27 +476,6 @@ def _build_name_remaps(model_keys: set[str]) -> dict[str, str]:
     return name_remaps
 
 
-def _expand_expert_lora_tensors(
-    lora_A: torch.Tensor,
-    lora_B: torch.Tensor,
-) -> tuple[torch.Tensor, torch.Tensor]:
-    if lora_A.shape[0] == 1:
-        if lora_B.shape[0] <= 1:
-            raise ValueError(
-                "Cannot broadcast expert LoRA: both A and B have 1 expert "
-                f"(lora_A: {lora_A.shape}, lora_B: {lora_B.shape})"
-            )
-        lora_A = lora_A.expand(lora_B.shape[0], -1, -1)
-    elif lora_B.shape[0] == 1:
-        if lora_A.shape[0] <= 1:
-            raise ValueError(
-                "Cannot broadcast expert LoRA: both A and B have 1 expert "
-                f"(lora_A: {lora_A.shape}, lora_B: {lora_B.shape})"
-            )
-        lora_B = lora_B.expand(lora_A.shape[0], -1, -1)
-    return lora_A, lora_B
-
-
 def _build_merge_ops(
     *,
     adapter_weights: dict[str, torch.Tensor],
@@ -531,7 +514,7 @@ def _build_merge_ops(
             )
             continue
 
-        lora_A, lora_B = _expand_expert_lora_tensors(lora_A, lora_B)
+        lora_A, lora_B = expand_expert_lora_tensors(lora_A, lora_B)
         target_key = target_key.replace(".w1.weight", ".gate_proj.weight")
         target_key = target_key.replace(".w3.weight", ".up_proj.weight")
         target_key = target_key.replace(".w2.weight", ".down_proj.weight")
@@ -562,7 +545,7 @@ def _apply_merge_ops_to_state_dict(
             if target_key not in merge_ops:
                 continue
             for op in merge_ops[target_key]:
-                merged_lora = op.lora_B @ op.lora_A
+                merged_lora = merge_lora_matrices(op.lora_A, op.lora_B)
                 apply_merged_weight(target_tensor, merged_lora)
 
 
--- a/tinker_cookbook/weights/_export.py
+++ b/tinker_cookbook/weights/_export.py
@@ -136,7 +136,7 @@ def build_hf_model(
             adapter_weights, adapter_config = _load_adapter_weights(Path(adapter_path))
 
             logger.info("Loading base model: %s (dtype=%s)", base_model, dtype)
-            config = _build_config(config_dict, base_model, trust_remote_code=resolved_trust)
+            config = AutoConfig.from_pretrained(base_model, trust_remote_code=resolved_trust)
             is_multimodal = _is_multimodal(config)
             hf_model = _load_model(
                 config, base_model, torch_dtype=torch_dtype, trust_remote_code=resolved_trust
@@ -215,28 +215,6 @@ def _load_config_dict(model_path: str) -> dict:
         return json.loads((candidate / "config.json").read_text())
     config_dict, _ = PretrainedConfig.get_config_dict(model_path)
     return config_dict
-
-
-def _build_config(
-    config_dict: dict,
-    model_path: str,
-    *,
-    trust_remote_code: bool,
-) -> PretrainedConfig:
-    model_type = config_dict.get("model_type")
-    if model_type:
-        try:
-            return AutoConfig.for_model(
-                model_type,
-                **{key: value for key, value in config_dict.items() if key != "model_type"},
-            )
-        except Exception:
-            logger.info(
-                "Falling back to AutoConfig.from_pretrained for model_type=%s from %s",
-                model_type,
-                model_path,
-            )
-    return AutoConfig.from_pretrained(model_path, trust_remote_code=trust_remote_code)
 
 
 def _load_adapter_weights(adapter_dir: Path) -> tuple[dict[str, torch.Tensor], dict]:
--- a/tests/weights/test_export.py
+++ b/tests/weights/test_export.py
@@ -34,6 +34,8 @@ from transformers import (
 import tinker_cookbook.weights._deepseek as deepseek_export
 from tinker_cookbook.weights import build_hf_model
 
+pytestmark = pytest.mark.integration
+
 # ---------------------------------------------------------------------------
 # Shared helpers
 # ---------------------------------------------------------------------------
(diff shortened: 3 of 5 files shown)


tinker_cookbook/weights/_deepseek.py
```diff
@@ -0,0 +1,858 @@
...
+
+    rows, cols = weight.shape
+    block_rows, block_cols = _DEEPSEEK_BLOCK_SIZE
+    scale_rows = math.ceil(rows / block_rows)
+    scale_cols = math.ceil(cols / block_cols)
+    scales = torch.empty((scale_rows, scale_cols), dtype=torch.float32)
+    quantized = torch.empty_like(weight, dtype=_DEEPSEEK_FP8_DTYPE)
+    max_fp8 = _get_fp8_max()
+
+    for row_idx in range(scale_rows):
```
YujiaBao: **Performance: pure Python double loop will be very slow at scale**

This double loop over `(scale_rows, scale_cols)` blocks runs once per routed expert per shard. DeepSeek V3.1 has 256 routed experts with large weight matrices — this will dominate export wall time and be orders of magnitude slower than a vectorized approach.

A vectorized version could reshape the weight into blocks, compute `abs().max()` and scale per block in one pass:
```python
# sketch — pad, reshape into blocks, vectorize scale + quantize
padded = F.pad(weight, (0, -cols % block_cols, 0, -rows % block_rows))
blocks = padded.reshape(scale_rows, block_rows, scale_cols, block_cols).permute(0, 2, 1, 3)
max_abs = blocks.abs().amax(dim=(-1, -2))
scales = torch.where(max_abs == 0, torch.ones_like(max_abs), max_abs / max_fp8)
quantized = (blocks / scales.unsqueeze(-1).unsqueeze(-1)).clamp(-max_fp8, max_fp8).to(fp8_dtype)
```

tinker_cookbook/weights/_deepseek.py
```diff
@@ -0,0 +1,858 @@
...
+
+    weights = load_file(str(safetensors_path), device="cpu")
+    config = json.loads(config_path.read_text())
+    return weights, config
+
+
+def _should_skip_checkpoint_key(name: str) -> bool:
+    # DeepSeek V3.1 HF checkpoints include a MultiTokenPrediction layer that is
+    # not part of the base causal LM weights we want to merge and serve.
+    return "model.layers.61" in name or "rotary_emb.inv_freq" in name
```
YujiaBao, line 921: **Hardcoded layer index is fragile**

`"model.layers.61" in name` is specific to DeepSeek V3/V3.1's 61-layer architecture. A future DeepSeek model with a different depth would silently get wrong behavior.

The MTP layer is always at index `num_hidden_layers` (i.e. the layer right after the last real transformer layer). Consider deriving this from the config:
```python
def _should_skip_checkpoint_key(name: str, num_hidden_layers: int) -> bool:
    return f"model.layers.{num_hidden_layers}" in name or "rotary_emb.inv_freq" in name
```

tinker_cookbook/weights/_export.py
```diff
@@ -83,32 +90,44 @@ def build_hf_model(
         raise ValueError(f"Unsupported dtype {dtype!r}. Choose from: {list(_DTYPE_MAP.keys())}")
     torch_dtype = _DTYPE_MAP[dtype]
     resolved_trust = _resolve_trust_remote_code(trust_remote_code)
-
-    # Validate adapter exists before loading the (potentially huge) base model
-    adapter_weights, adapter_config = _load_adapter_weights(Path(adapter_path))
-
     out = Path(output_path)
-    out.mkdir(parents=True, exist_ok=False)
+    preserve_partial_output = _is_deepseek_model_path(base_model)
```
YujiaBao: **Unconditional network call for all models**

`preserve_partial_output = _is_deepseek_model_path(base_model)` runs before the main export and calls `snapshot_download` for non-local models just to read `config.json`. This means every non-DeepSeek export (Qwen, GPT-OSS, etc.) now pays an extra HF Hub round-trip.

Consider deferring this check — e.g. load the config once and branch on it, or check `is_deepseek_config` on the config that's already fetched in the non-DeepSeek path.

tinker_cookbook/weights/_deepseek.py
dphuang2: for now, i'll just keep this because it works and I don't want to touch it. We can revisit when add more models with DS architecture to tinker

YujiaBao:
**Consider a model-family registry pattern for export paths**

Right now the export logic branches on `is_deepseek_config` in `_export.py`, and all DeepSeek-specific code lives in a flat `_deepseek.py` sibling file. As more model families need custom export paths (each with their own quantization strategy, shard layout, custom files, etc.), this pattern will lead to a growing chain of `if/elif` checks in `_export.py` and an accumulation of `_<model>.py` files at the same level.

A cleaner structure might be:
```
tinker_cookbook/weights/
  __init__.py
  _export.py          # common export logic + model registry dispatch
  _merge.py           # shared merge helpers (e.g. _load_adapter_weights)
  _models/
    __init__.py        # registry (e.g. get_exporter(config) -> ModelExporter)
    _deepseek.py       # DeepSeek-specific export
    _qwen.py           # future Qwen-specific export
    ...
```

Each model family would implement a common interface (e.g. `save_model(base_model, adapter_path, output_path)`), and `_export.py` would dispatch via registry lookup rather than hardcoded conditionals. `_merge.py` already has a note about this pattern in its docstring. Shared helpers like `_load_adapter_weights` (currently duplicated between `_export.py` and `_deepseek.py`) would live in `_merge.py`.

Not necessarily blocking for this PR, but worth considering before more model families land — it'll be harder to refactor once there are several parallel implementations.

tinker_cookbook/weights/_deepseek.py
```diff
@@ -0,0 +1,858 @@
...
+        }
+        _write_json_atomic(state_path, state)
+
+    completed = {
+        shard_name
+        for shard_name in state.get("completed_shards", [])
+        if (output_path / shard_name).exists()
+    }
+    for shard_path in output_path.glob("*.safetensors"):
+        completed.add(shard_path.name)
```
YujiaBao: **Resume logic trusts on-disk `.safetensors` files unconditionally**

Lines 683-684 add any `.safetensors` file found in the output directory to `completed`, even if it wasn't tracked in `merge_state.json`. If a user accidentally points `output_path` to a directory with existing safetensors files from a different run (but with a matching `merge_state.json`), these get silently incorporated into the output, producing a corrupt checkpoint.

Consider only trusting shards listed in `merge_state.json["completed_shards"]` rather than also scanning the directory.

tinker_cookbook/weights/_deepseek.py
```diff
@@ -0,0 +1,858 @@
...
+
+def _expand_expert_lora_tensors(
+    lora_A: torch.Tensor,
+    lora_B: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    if lora_A.shape[0] == 1:
+        lora_A = lora_A.expand(lora_B.shape[0], -1, -1)
+    elif lora_B.shape[0] == 1:
+        lora_B = lora_B.expand(lora_A.shape[0], -1, -1)
+    return lora_A, lora_B
```
YujiaBao: **Missing validation when both lora_A and lora_B have shape[0]==1**

Unlike `_merge.py` (lines 100-113) which explicitly raises a `ValueError` when both A and B have `shape[0] == 1`, this function silently returns them as-is. This inconsistency could hide bugs where a single-expert adapter is mistakenly treated as valid.

tinker_cookbook/weights/_deepseek.py
```diff
@@ -0,0 +1,858 @@
...
+
+def is_deepseek_config(config: object) -> bool:
+    """Return True when a config should use the DeepSeek export path."""
+    if isinstance(config, dict):
+        model_type = str(config.get("model_type", "")).lower()
+        class_name = ""
+    else:
+        model_type = str(getattr(config, "model_type", "")).lower()
+        class_name = config.__class__.__name__.lower()
+    return model_type.startswith("deepseek") or class_name.startswith("deepseek")
```
YujiaBao: **`startswith("deepseek")` detection is overly broad**

This matches any model whose `model_type` or class name starts with "deepseek". A future model with a different architecture but a similar prefix (e.g. a third-party "deepseek_custom" variant) would be silently routed through this export path. Consider matching against a known set of model types (e.g. `{"deepseek_v3", "deepseek_v2"}`) instead of a prefix match.

tinker_cookbook/weights/_deepseek.py
```diff
@@ -0,0 +1,858 @@
...
+    extra_fields = {"scale_dtype", "zp_dtype"}
+    for group in config_dict.get("config_groups", {}).values():
+        if not isinstance(group, dict):
+            continue
+        for section in ("weights", "input_activations", "output_activations"):
+            args = group.get(section)
+            if not isinstance(args, dict):
+                continue
+            for key in extra_fields:
+                args.pop(key, None)
```
YujiaBao: **Fragile compatibility shim with hardcoded field names**

Hardcoding `{"scale_dtype", "zp_dtype"}` as fields to strip will need maintenance as `compressed-tensors` evolves. Each new version could add or rename fields that vLLM rejects. This is another reason to pin `compressed-tensors` to a tested version range — then you'd know exactly which fields exist and this stripping logic could be removed.

tinker_cookbook/weights/_deepseek.py
```diff
@@ -0,0 +1,858 @@
...
+import torch
+from huggingface_hub import snapshot_download
+from safetensors import safe_open
+from safetensors.torch import load_file
+from safetensors.torch import save_file
+from transformers import PreTrainedModel
+
+from tinker_cookbook.weights._merge import apply_merged_weight
+
+logger = logging.getLogger(__name__)
```
YujiaBao, line 45: **Consider adding `__all__` to clarify the public surface**

The module is `_`-prefixed (private) but exports `is_deepseek_config`, `is_deepseek_model`, and `save_deepseek_model` via `_export.py`. An explicit `__all__` would make the intended public surface clear and prevent accidental imports of internal helpers.

tinker_cookbook/weights/_deepseek.py
dphuang2: this is fine, I feel the developer should know what path they are passing so by default its "trusted"

tinker_cookbook/weights/_deepseek.py
dphuang2: can revisit if this is an issue

tinker_cookbook/weights/_deepseek.py
dphuang2: hm, this seems a bit too defensive but I can add it for completeness

dphuang2:
> **Consider a model-family registry pattern for export paths**
> 
> Right now the export logic branches on `is_deepseek_config` in `_export.py`, and all DeepSeek-specific code lives in a flat `_deepseek.py` sibling file. As more model families need custom export paths (each with their own quantization strategy, shard layout, custom files, etc.), this pattern will lead to a growing chain of `if/elif` checks in `_export.py` and an accumulation of `_<model>.py` files at the same level.
> 
> A cleaner structure might be:
> 
> ```
> tinker_cookbook/weights/
>   __init__.py
>   _export.py          # common export logic + model registry dispatch
>   _merge.py           # shared merge helpers (e.g. _load_adapter_weights)
>   _models/
>     __init__.py        # registry (e.g. get_exporter(config) -> ModelExporter)
>     _deepseek.py       # DeepSeek-specific export
>     _qwen.py           # future Qwen-specific export
>     ...
> ```
> 
> Each model family would implement a common interface (e.g. `save_model(base_model, adapter_path, output_path)`), and `_export.py` would dispatch via registry lookup rather than hardcoded conditionals. `_merge.py` already has a note about this pattern in its docstring. Shared helpers like `_load_adapter_weights` (currently duplicated between `_export.py` and `_deepseek.py`) would live in `_merge.py`.
> 
> Not necessarily blocking for this PR, but worth considering before more model families land — it'll be harder to refactor once there are several parallel implementations.

makes sense as a follow up PR

tinker_cookbook/weights/_export.py
```diff
@@ -83,32 +89,45 @@ def build_hf_model(
...
     torch_dtype = _DTYPE_MAP[dtype]
     resolved_trust = _resolve_trust_remote_code(trust_remote_code)
-
-    # Validate adapter exists before loading the (potentially huge) base model
-    adapter_weights, adapter_config = _load_adapter_weights(Path(adapter_path))
-
     out = Path(output_path)
-    out.mkdir(parents=True, exist_ok=False)
+    config_dict = _load_config_dict(base_model)
+    preserve_partial_output = _is_deepseek_config(config_dict)
```
YujiaBao, line 117: The DeepSeek path silently changes the behavioral contract of `build_hf_model`:

1. **`dtype` is ignored** — DeepSeek always uses BF16/FP8 regardless of what the user passes. A user calling `build_hf_model(..., dtype="float32")` for max precision gets silently ignored.
2. **`trust_remote_code` is ignored** — the DeepSeek path does its own config loading.
3. **`output_path` no longer raises `FileExistsError`** — the docstring says "Must not already exist" but DeepSeek sets `exist_ok=True`.

Since this is a public API, we should either:
- Log a warning when `dtype`/`trust_remote_code` are overridden by the DeepSeek path
- Or raise a `ValueError` if the user passes incompatible options
- And update the docstring to document the DeepSeek resume behavior for `output_path`

tinker_cookbook/weights/_deepseek.py
```diff
@@ -0,0 +1,931 @@
...
+
+def is_deepseek_config(config: object) -> bool:
+    """Return True when a config should use the DeepSeek export path."""
+    if isinstance(config, dict):
+        model_type = str(config.get("model_type", "")).lower()
+        class_name = ""
+    else:
+        model_type = str(getattr(config, "model_type", "")).lower()
+        class_name = config.__class__.__name__.lower()
+    return model_type.startswith("deepseek") or class_name.startswith("deepseek")
```
YujiaBao: I think we should address this before merging. Because `is_deepseek_config` gates the entire export path in `build_hf_model` — it decides whether `dtype`, `trust_remote_code`, and `output_path` semantics change — a false positive silently produces wrong behavior with no error. Matching specific known model types (e.g. `{"deepseek_v3"}`) is a one-line change and prevents a class of silent failures.

tinker_cookbook/weights/_deepseek.py
YujiaBao: The concern isn't about malicious input — it's about accidental misuse. If a user resumes an interrupted export but points `output_path` to a directory that has leftover `.safetensors` from a previous different run, lines 683-684 silently incorporate those into the output. The result is a corrupt checkpoint that loads fine but produces garbage at inference time. That's very expensive to debug. We should only trust shards listed in `merge_state.json["completed_shards"]`, not anything found on disk.

pyproject.toml
```diff
@@ -79,6 +80,11 @@ all = [
     "tinker_cookbook[inspect]",
     "tinker_cookbook[litellm]",
 ]
+merge = [
+    "accelerate",
+    "compressed-tensors==0.14.0.1",
+    "psutil>=7.2.2",
```
YujiaBao, line 89: A few issues with this extra:

1. **Naming**: `tinker_cookbook[merge]` leaks an implementation detail. Users call `build_hf_model` — they shouldn't need to know about internal merge steps. If this extra is specifically for DeepSeek FP8 export, `tinker_cookbook[deepseek]` would be clearer, or fold the deps into the base package.
2. **Exact pin**: `compressed-tensors==0.14.0.1` will cause install conflicts for users with other packages depending on a different version. Use `>=0.14,<0.15` or `~=0.14.0` unless there's a known breakage.
3. **Unused dep**: `accelerate` is listed but never imported anywhere in the PR.

tinker_cookbook/weights/_deepseek.py
```diff
@@ -0,0 +1,931 @@
...
+1. Quantization selection: only routed expert MLP weights are quantized to FP8,
+   while dense, shared-expert, embedding, attention, and norm weights stay in
+   higher precision.
+2. Checkpoint packaging: process one reference shard at a time on CPU, preserve
+   the upstream shard names and tensor placement, and emit a resumable output
+   index that tracks the shards already written.
+
+Preserving the reference shard layout is not a mathematical requirement for
+checkpoint correctness. It is the chosen implementation strategy here because it
+makes the work unit "one reference shard", keeps DeepSeek cross-shard scale
```
YujiaBao, line 14: This module is 931 lines of new code with no colocated unit tests (`deepseek_test.py`). The existing test coverage in `tests/weights/test_export.py` requires downloading from `deepseek-ai/DeepSeek-V3.1`, so it can't run offline or in CI without network access.

The rest of the weights module follows a pattern where unit tests sit next to the code (e.g. `merge_test.py` next to `_merge.py`) using synthetic tensors and dicts — no network, no real models. We should add a `deepseek_test.py` following the same pattern, covering at least:
- `_is_routed_expert_weight` — key name classification
- `_quantize_weight_blockwise` / `_dequantize_fp8_blockwise` — round-trip correctness
- `_build_merge_ops` — name remapping and expert expansion
- `_load_resume_state` — resume logic with synthetic on-disk state
- `_build_output_shard_matching_reference` — shard assembly and dtype selection
- `is_deepseek_config` — detection logic

Without offline unit tests, we won't be able to confidently refactor this code in followup PRs.

YujiaBao:
closed as we merged https://github.com/thinking-machines-lab/tinker-cookbook/pull/478

PR #475, opened 2026-03-17
Widen transformers to >=4.57.6,<5.3.0 for Qwen 3.5 support

YujiaBao:
## Summary
- **Raise transformers upper bound** from `<5.0.0` to `<5.3.0` — Qwen 3.5 support landed in v5.2.0 (Feb 2026); the old cap was a precautionary guard with no known incompatibility
- **Upper bound at `<5.3.0`** because transformers 5.3.0 reintroduced a known DeepSeek tokenizer regression ([huggingface/transformers#43975](https://github.com/huggingface/transformers/issues/43975)) where spaces are stripped during decode (e.g. `"hello world"` → `"helloworld"`). This was originally fixed in 5.2.0 but regressed in 5.3.0.
- **Fix pyright error:** `transformers.tokenization_utils` module was removed in 5.x — switch to top-level `from transformers import PreTrainedTokenizer`
- **Drop deprecated `local_dir_use_symlinks`** from `hf_hub_download` call (parameter removed in newer `huggingface_hub`)
- **Fix renderer tests for transformers 5.x:** `apply_chat_template(tokenize=True)` now returns `BatchEncoding` instead of `list[int]`. Add shared `extract_token_ids()` helper to normalize the return type across both versions.
- **Fix pyright errors from 5.x type stubs:** `tokenizer.decode()` return type changed from `str` to `str | list[str]` in 5.x stubs. Wrap affected call sites with `str()`.
- **Add CI version matrix:** pyright and pytest now run against both transformers `4.57.6` (lower bound) and `5.2.0` (latest allowed) to catch version-specific regressions.

## Compatibility analysis

We audited all transformers usage in the codebase. The library only uses stable public APIs (`Auto*` classes, `PreTrainedModel`, `PretrainedConfig`, `.from_pretrained()`, `.save_pretrained()`). `apply_chat_template` is only used in tests, not production code. All changes are verified to work on both transformers 4.57.6 and 5.2.0.

| File | Change | 4.x | 5.x (<5.3.0) |
|------|--------|-----|-----|
| `pyproject.toml` | `>=4.57.6,<5.3.0` | ✅ | ✅ |
| `tokenizer_utils.py` | `from transformers import PreTrainedTokenizer` (was `from transformers.tokenization_utils`) | ✅ runtime + pyright | ✅ runtime + pyright |
| `search_env.py` | Drop `local_dir_use_symlinks` | ✅ | ✅ |
| `renderers/*_test.py` | Use `extract_token_ids()` for `apply_chat_template` results | ✅ | ✅ |
| `renderers/base.py`, `gpt_oss.py`, etc. | Wrap `decode()` with `str()` for pyright | ✅ | ✅ |

**Other transformers usages verified — no changes needed:**
- `image_processing_utils.py` — internal imports resolve on both versions
- `weights/_export.py` — stable `Auto*` APIs; `dtype` is the correct documented parameter
- `hyperparam_utils.py` — `AutoConfig.from_pretrained` + `config.hidden_size`, stable API
- No `apply_chat_template` usage in production code or recipes — only in tests

**DeepSeek tokenizer regression in 5.3.0:**
Bisected across 5.0.0–5.3.0. Versions 5.0.0–5.2.0 decode correctly; 5.3.0 strips spaces. This is a known upstream bug ([#43975](https://github.com/huggingface/transformers/issues/43975)) that was fixed in 5.2.0 but regressed in 5.3.0. We can raise the upper bound once the fix lands upstream.

## Test plan
- [x] `uv run pyright tinker_cookbook` — 0 errors on both 4.57.6 and 5.2.0
- [x] `uv run pytest tinker_cookbook/` — 462 passed, 0 failed on both 4.57.6 and 5.2.0
- [ ] CI matrix passes (pyright × {4.57.6, 5.2.0}, pytest × {4.57.6, 5.2.0})

### DIFF
--- a/.github/workflows/pyright.yaml
+++ b/.github/workflows/pyright.yaml
@@ -9,6 +9,12 @@ jobs:
   type-check:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        transformers-version: ["4.57.6", "5.2.0"]
+
+    name: type-check (transformers ${{ matrix.transformers-version }})
+
     steps:
       - name: checkout
         uses: actions/checkout@v4
@@ -20,6 +26,9 @@ jobs:
 
       - name: venv
         run: uv venv && uv sync --all-extras
+
+      - name: pin transformers
+        run: uv pip install transformers==${{ matrix.transformers-version }}
 
       - name: pyright
         run: uv run pyright tinker_cookbook
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -10,6 +10,12 @@ jobs:
   test:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        transformers-version: ["4.57.6", "5.2.0"]
+
+    name: test (transformers ${{ matrix.transformers-version }})
+
     steps:
       - name: checkout
         uses: actions/checkout@v4
@@ -21,6 +27,9 @@ jobs:
 
       - name: venv
         run: uv venv && uv sync --all-extras
+
+      - name: pin transformers
+        run: uv pip install transformers==${{ matrix.transformers-version }}
 
       - name: pytest (unit)
         run: uv run pytest tinker_cookbook/


YujiaBao:
> would our tokenizer tests catch any regressions? Should we do a matrix-based integration test for this change?

i think our tests already caught errors. there are some behavior changes in transformer 5.3.0

tinker_cookbook/renderers/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+
+def extract_token_ids(result: Any) -> list[int]:
+    """Extract token IDs from apply_chat_template result.
+
+    transformers 4.x returns list[int], while 5.x returns BatchEncoding (dict-like
+    with 'input_ids' and 'attention_mask' keys). This helper normalizes both to list[int].
+    """
+    if hasattr(result, "input_ids"):
+        return list(result["input_ids"])
+    return list(result)
```
dphuang2, line 16: :(

dphuang2 (approved):

PR #373, opened 2026-02-19
Enable merging of large models

erikwijmans:
Updates tinker_cookbook/scripts/merge_tinker_adapter_to_hf_model.py to enable large models that don't fit within CPU memory.

PR #476, opened 2026-03-17
Add shard-by-shard merging and modular merge architecture

YujiaBao:
## Summary

Adds shard-by-shard LoRA merging for large models that don't fit in CPU memory, and restructures the weights module into a clean two-layer architecture that supports pluggable model-specific export strategies.

**Motivated by:** #373 (shard-by-shard merge for large models) and #470 (DeepSeek V3.1 FP8 export). Both PRs revealed that the existing monolithic merge function doesn't scale — #373 because it loads the full model into memory, #470 because model-specific export behavior was forced into a separate 900+ line module with duplicated merge logic.

## Problem

Two related problems in `tinker_cookbook.weights`:

1. **Memory**: `build_hf_model` loads the entire base model into CPU memory via `from_pretrained()`. For 70B+ models, this exceeds available RAM on many machines.

2. **Extensibility**: Model-specific merge behavior was controlled by ad-hoc boolean flags (`is_gpt_oss`, `is_fused_experts`, inline vision prefix checks). Adding a new model meant scattering more `if/else` branches across the code, and #470 had to build a completely separate module because there was no clean extension point.

### Two-layer architecture

Model-specific differences fall into two distinct layers:

**Layer 1 — Merge-level (generalizable):** Key remapping, expert layout, LoRA delta math. Different models use the same algorithm with different configuration. A `MergeProfile` dataclass captures this, with per-model detectors registered in `_PROFILE_DETECTORS`.

**Layer 2 — Export-level (per-strategy):** How the merged model is assembled and written to disk. The standard path (full/shard) and model-specific paths (e.g. DeepSeek FP8) differ fundamentally, so each gets its own module under `_export/`.

```
tinker_cookbook/weights/
├── __init__.py              # Public API (unchanged)
├── _merge.py                # Layer 1: MergeProfile, MergeOp, plan/apply, shared math
├── _artifacts.py            # Shared I/O: ShardWriter, adapter loading, model resolution
├── _download.py             # Download from Tinker (unchanged)
├── _publish.py              # Publish to HF Hub (unchanged)
├── _export/                 # Layer 2: Export strategies
│   ├── __init__.py          #   build_hf_model dispatch + shared helpers
│   ├── _full.py             #   Full-model strategy (original behavior)
│   └── _shard.py            #   Shard-by-shard strategy (low memory)
```

### Adding a new model family

**If the model only needs different key remapping or expert layout:**
1. Add a detector function to `_PROFILE_DETECTORS` in `_merge.py`
2. Done — both `_full.py` and `_shard.py` work automatically

**If the model needs a fundamentally different output format** (like DeepSeek FP8):
1. Create `_export/_<model>.py` with its own export pipeline
2. Add detection + dispatch in `_export/__init__.py` (hook is already in place)
3. Reuse shared utilities from `_merge.py` and `_artifacts.py`

### API change

Two new optional parameters on `build_hf_model` (fully backward compatible):

```python
def build_hf_model(
    *,
    base_model: str,
    adapter_path: str,
    output_path: str,
    dtype: str = "bfloat16",
    trust_remote_code: bool | None = None,
    merge_strategy: str = "auto",   # NEW: "auto" | "shard" | "full"
    dequantize: bool = False,       # NEW: placeholder (NotImplementedError)
) -> None:
```

- `merge_strategy="auto"` (default) uses shard-by-shard — identical output, strictly less memory
- `merge_strategy="full"` forces full-model loading (original behavior, escape hatch)
- `dequantize=True` raises `NotImplementedError` — reserves the API for future dequantization support

### Key components

**`_merge.py` — plan-then-execute merge pipeline:**
- `MergeProfile` — frozen dataclass describing model-specific merge config (expert layout, key remaps, vision prefix)
- `detect_merge_profile()` — dispatches to per-model detectors (`_detect_gpt_oss_profile`, `_detect_default_profile`). New models add a detector to `_PROFILE_DETECTORS`
- `MergeOp` — stores small LoRA matrices; full delta computed on-the-fly during apply
- `plan_merge_ops()` — maps adapter keys to model keys, validates upfront before heavy I/O
- `apply_merge_op()` — applies one merge to a tensor dict (works for full state dict or single shard)
- `validate_merge_op_shapes()` — checks all merge op shapes against target shapes before shard processing (fails fast)
- `merge_lora_matrices()` / `expand_expert_lora_tensors()` — extracted math utilities
- `merge_adapter_weights()` — backward-compatible wrapper, unchanged signature

**`_artifacts.py` — shared model artifact I/O:**
- `ShardWriter` — accumulates tensors and writes numbered safetensors shards with temp→final renaming
- `get_model_state_shapes()` — reads shapes from safetensors headers without loading tensor data
- `get_shard_files()` — discovers shard files from index or glob
- `resolve_model_dir()` — local path or HF Hub download
- `load_adapter_weights()` — shared adapter loading (was duplicated)
- `copy_model_code_files()` — copies only `*.py` files for trust_remote_code support

**`_export/__init__.py` — dispatcher:**
- Loads model config once (lightweight, no weight download)
- Has commented hook for model-family specific override (PR #470 plugs in here)
- Dispatches to `_full.py` or `_shard.py` based on `merge_strategy`
- Logs warning when `dtype` is specified with shard strategy (shard preserves on-disk dtype)

**`_export/_shard.py` — shard-by-shard merge flow:**
1. Fail fast if output exists → 2. Load adapter (small) → 3. Read shapes from safetensors headers → 4. Detect merge profile → 5. Plan all merge ops → 6. Validate shapes upfront → 7. For each shard: load, apply ops, write via ShardWriter → 8. Verify all ops consumed → 9. Finalize shards + write index → 10. Copy config + model code + save tokenizer

### Robustness

- **Upfront shape validation** — all merge op shapes checked against target tensor shapes before any shard I/O begins (via safetensors header reading, no tensor loading)
- **Early output-exists check** — `FileExistsError` raised before any expensive model/adapter loading
- **Empty adapter warning** — logs warning when adapter has zero LoRA weights (catches wrong adapter format)
- **Helpful error for non-safetensors models** — detects `.bin` files and suggests `merge_strategy='full'`
- **Missing config.json error** — clear `FileNotFoundError` for local dirs without config
- **Atomic shard writes** — `ShardWriter` uses temp files, renames on finalize; shard count only incremented after successful write
- **Cleanup on failure** — partial output directory removed on exception

## What stays unchanged

- `__init__.py` — public API surface unchanged
- `_download.py`, `_publish.py` — unchanged
- `merge_adapter_weights()` — same signature, same behavior (now delegates to plan+apply internally)
- `build_hf_model()` — existing calls with no new params work identically
- All 15 original `merge_test.py` tests pass without modification

### Unit tests (102 total, no network needed)
- [x] 15 original backward-compat tests pass unchanged
- [x] 29 new `merge_test.py` tests: `MergeProfile`, `detect_merge_profile`, `plan_merge_ops`, `apply_merge_op`, `validate_merge_op_shapes`, `merge_lora_matrices`, `expand_expert_lora_tensors`
- [x] 20 new `artifacts_test.py` tests: `ShardWriter`, `get_model_state_keys`, `get_model_state_shapes`, `get_shard_files`, `load_adapter_weights`, `copy_model_code_files`
- [x] 9 new `export_test.py` tests: end-to-end shard merge (single shard, multi-shard, separate experts, fused experts, single-shard no index), strategy dispatch validation, cleanup on failure
- [x] 19 new `stress_test.py` tests: numerical correctness (exact values, scaling, higher rank, bfloat16 precision), cross-path consistency (plan+apply matches merge_adapter_weights bit-for-bit), expert edge cases, shard path edge cases, shape validation, profile detection, config loading, ShardWriter edge cases

### Integration tests (require network)
- [x] All 6 existing integration tests pass with shard-by-shard default (GPT-OSS, Qwen3-VL MoE, Qwen3 MoE, Qwen3 dense)
- [x] **Strategy consistency**: full and shard produce **bit-identical** weights on tiny Qwen3-8B model with bfloat16 weights

### DIFF
--- a/tinker_cookbook/weights/stress_test.py
+++ b/tinker_cookbook/weights/stress_test.py
@@ -146,7 +146,7 @@ class TestCrossPathConsistency:
 
         # Path 1: merge_adapter_weights (backward-compat wrapper)
         sd1 = {k: v.clone() for k, v in state_dict.items()}
-        model1 = type("Model", (), {"state_dict": lambda self: sd1})()
+        model1 = type("Model", (torch.nn.Module,), {"state_dict": lambda self: sd1})()
         merge_adapter_weights(model1, adapter, config)
 
         # Path 2: plan + apply (new API)


tinker_cookbook/weights/_artifacts.py
```diff
@@ -0,0 +1,258 @@
...
+        config = json.load(f)
+    return weights, config
+
+
+# ---------------------------------------------------------------------------
+# Non-weight file copying
+# ---------------------------------------------------------------------------
+
+
+def copy_non_weight_files(model_dir: Path, output_path: Path) -> None:
```
erikwijmans: Doing this has potentially has some anti-patterns, like accidentally copying the `model.safetensors.index.json` and really messing up the load for vllm/sglang if any of the keys changed file location (which this will do). I'd copy just the `*.py` files and then load+save the config, tokenizer, and image encoder (if there is one). Ideally we'd not copy the python files too and save the model but I don't think there's a way to tell HF to save everything but the weights

dphuang2 (approved):

PR #480, opened 2026-03-18
Fix async RL training hang on data exhaustion via cascading shutdown

YujiaBao:
## Summary

- Replace centralized `shutdown_event` + `shutdown_loops()` in `do_async_training` with a cascading shutdown protocol: **dataloader -> workers -> training loop -> evaluation loop**
- When the dataloader exhausts data, it enqueues `_Shutdown` sentinels to each worker; the last worker signals the training loop, which signals the eval loop
- Fix potential deadlock from requeuing stale samples into a full bounded queue during shutdown — stale samples are now discarded when the dataloader is done
- Handle `_Shutdown` in `do_train_step_streaming_and_get_sampling_client` (returns `None` on shutdown)
- Replace `tests/smoke_tests.py` with proper recipe-level tests in `tests/recipes/test_recipe_math_rl.py`

## Motivation

Previously, async RL runs could hang indefinitely when the dataloader ran out of data. The training loop would block on `trajectory_groups_queue.get()` waiting for items that would never arrive, because `shutdown_loops()` was only called *after* the training loop completed — a deadlock. See #107 for the original report.

### New internal types (private, not part of public API)

- **`_Shutdown`** — A `@dataclass` sentinel value enqueued through async queues to signal graceful shutdown. Replaces the overloaded use of `None` (which means "failed rollout") as a shutdown signal.
- **`_AsyncCounter`** — An async-safe counter that tracks the number of alive worker tasks. The last worker to decrement to zero enqueues a `_Shutdown` to the training queue.

### Modified functions

- **`do_async_training`** — Removed `shutdown_event` (`asyncio.Event`) and `shutdown_loops()`. Added three new coordination primitives:
  - `dataloader_done_event` — set when dataloader finishes; prevents requeuing stale samples during shutdown (avoids deadlock on full bounded queue)
  - `evaluation_loop_should_shutdown_event` — set by training loop to signal eval loop to exit
  - `worker_alive_counter` (`_AsyncCounter`) — tracks alive workers; last worker signals training loop
  
  Each nested loop was updated:
  - `dataloader_loop`: enqueues `_Shutdown` × `groups_per_batch` on completion
  - `trajectory_group_worker_loop`: checks for `_Shutdown` instead of `None`; last worker enqueues `_Shutdown` to trajectory queue
  - `training_loop`: checks for `_Shutdown` from queue before processing; skips stale-sample requeue when `dataloader_done_event` is set
  - `evaluation_loop`: uses `evaluation_loop_should_shutdown_event` instead of `shutdown_event`

- **`do_train_step_streaming_and_get_sampling_client`** — Queue parameter type widened to `asyncio.Queue[WrappedTrajectoryGroup | _Shutdown | None]`. Return type widened to `... | None`. Returns `None` when a `_Shutdown` sentinel is received mid-minibatch.

- **`do_sync_training_with_stream_minibatch`** — Added assertion that streaming result is not `None` (safety net; `_Shutdown` cannot appear in this path's local queue).

### Math RL recipe

- Added `stream_minibatch` CLI flag to `tinker_cookbook/recipes/math_rl/train.py`

### Test migration

- Deleted `tests/smoke_tests.py` — confusing name, predated the recipe test pattern
- Added `tests/recipes/test_recipe_math_rl.py` with 3 tests (sync, async, stream minibatch) — auto-discovered by nightly CI via `smoke-test-recipes.yaml`

### No public API changes

`Config`, `AsyncConfig`, `StreamMinibatchConfig`, `WrappedTrajectoryGroup`, and all recipe-facing interfaces are unchanged. All new types are underscore-prefixed (private).

## Test plan

- [x] 8 new unit tests in `tinker_cookbook/rl/shutdown_test.py` (no API key needed)
- [x] All existing unit tests pass (503 passed)
- [x] 3 new recipe tests in `tests/recipes/test_recipe_math_rl.py`:
  - [x] `test_math_rl_sync`
  - [x] `test_math_rl_async` — exercises the shutdown cascade
  - [x] `test_math_rl_stream_minibatch` — exercises sync streaming path

### DIFF
--- a/tinker_cookbook/recipes/math_rl/train.py
+++ b/tinker_cookbook/recipes/math_rl/train.py
@@ -64,7 +64,7 @@ class CLIConfig:
 
     # Stream minibatch: train on minibatches as soon as they are ready
     # instead of waiting for the full batch.
-    stream_minibatch: bool = False
+    stream_minibatch_config: StreamMinibatchConfig | None = None
 
     # Loss function and configuration.
     # See https://tinker-docs.thinkingmachines.ai/losses
@@ -160,9 +160,9 @@ async def cli_main(cli_config: CLIConfig):
         else None,
         stream_minibatch_config=StreamMinibatchConfig(
             groups_per_batch=cli_config.groups_per_batch,
-            num_minibatches=2,
+            num_minibatches=cli_config.stream_minibatch_config.num_minibatches,
         )
-        if cli_config.stream_minibatch
+        if cli_config.stream_minibatch_config is not None
         else None,
         loss_fn=cli_config.loss_fn,
         loss_fn_config=cli_config.loss_fn_config,
--- a/tinker_cookbook/rl/train.py
+++ b/tinker_cookbook/rl/train.py
@@ -693,10 +693,14 @@ async def do_async_training(
         ttl_seconds=cfg.ttl_seconds,
     )
 
-    # Shutdown coordination:
-    # - dataloader_done_event: set when dataloader exhausts data, prevents requeuing stale samples
-    # - evaluation_loop_should_shutdown_event: set by training loop to signal eval loop to exit
-    # - worker_alive_counter: tracks alive workers; last worker signals training loop
+    # Shutdown coordination — cascading sequence:
+    # 1. Dataloader exhausts data → sets dataloader_done_event (prevents requeuing stale
+    #    samples) and enqueues one _Shutdown sentinel per worker into env_group_builders_queue.
+    # 2. Each trajectory worker receives its _Shutdown sentinel → exits and decrements
+    #    worker_alive_counter. The last worker enqueues a _Shutdown into trajectory_groups_queue.
+    # 3. Training loop receives _Shutdown from trajectory_groups_queue → finishes current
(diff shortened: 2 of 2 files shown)


tinker_cookbook/recipes/math_rl/train.py
```diff
@@ -62,6 +62,10 @@ class CLIConfig:
 
     max_steps_off_policy: int | None = None
 
+    # Stream minibatch: train on minibatches as soon as they are ready
+    # instead of waiting for the full batch.
+    stream_minibatch: bool = False
```
kennyyu: It might make sense to inline the mini batch config here to avoid duplicating params:

```
stream_minibatch_config: StreamMinibatchConfig | None = None
```

tinker_cookbook/rl/train.py
```diff
@@ -666,37 +693,46 @@ async def do_async_training(
         ttl_seconds=cfg.ttl_seconds,
     )
 
+    # Shutdown coordination:
```
kennyyu: If may be helpful to document the full shutdown sequence, something like
1. dataloader runs out of data => XX happens
2. then trajectory workers do YY
3. ZZ happens

PR #492, opened 2026-03-18
Nemotron3 model and renderer

opherlieber:


tinker_cookbook/renderers/nemotron3.py
```diff
@@ -0,0 +1,327 @@
...
+   newlines (Qwen3.5 uses <think>\\n\\n</think>\\n\\n).
+
+6. Empty system message injection: Nemotron-3's HF template always outputs
+   a system message block even when none is provided (it always sets
+   system_message = "" which is "defined" in Jinja2). Our renderer
+   prepends an empty system message in build_generation_prompt and
+   build_supervised_example to match this behavior.
+
+Reference: chat_template.jinja in
+  /gcs/tml-us-south1-ckpt-01/checkpoints/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/
```
dphuang2:
```suggestion
Reference: https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/blob/main/chat_template.jinja
```

tinker_cookbook/renderers/nemotron3_test.py
```diff
@@ -0,0 +1,591 @@
...
+    Nemotron3Renderer,
+    _format_nemotron3_tool_declaration,
+)
+from tinker_cookbook.tokenizer_utils import get_tokenizer
+
+NEMOTRON3_NANO_MODEL = "nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16"
+NEMOTRON3_SUPER_MODEL = "nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16"
+NEMOTRON3_TOKENIZER_PATH = (
+    "/gcs/tml-us-south1-ckpt-01/checkpoints/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16"
+)
```
dphuang2: this will cause test failures, should we just pull from hugging face directly

tinker_cookbook/renderers/nemotron3.py
```diff
@@ -0,0 +1,325 @@
...
+            # Qwen3_5Renderer already stripped \n\n; put back one \n that was
+            # over-stripped (Nemotron-3 has one separator newline, not two).
+            # Only re-add if the Qwen3.5 parent stripped exactly two newlines.
+            if first_text_after_thinking is not None and not first_text_after_thinking[
+                "text"
+            ].startswith("\n"):
+                # Parent stripped \n\n; we need \n separator, so add one back.
+                first_text_after_thinking["text"] = "\n" + first_text_after_thinking["text"]
+
+        return message, success
```
YujiaBao: Looks like `_parse_response_for_streaming` isn't overridden here — the inherited version from `Qwen3_5Renderer` calls `_postprocess_parsed_message` which strips `\n\n` after `</think>`, but unlike `parse_response` there's nothing adding the `\n` back for Nemotron-3's single-newline separator. So streaming responses with thinking would end up with the wrong separator (0 newlines instead of 1).

Maybe the cleanest fix is to override `_postprocess_parsed_message` to strip only `\n` instead of `\n\n`, so both `parse_response` and `_parse_response_for_streaming` get the right behavior without needing the post-hoc correction?

tinker_cookbook/renderers/nemotron3.py
```diff
@@ -0,0 +1,325 @@
...
+        if has_think and not is_historical:
+            return ""
+        # For historical messages with stripped thinking and non-empty text:
+        # add \n separator to match HF template's c.split('</think>')[-1] behavior.
+        if is_historical and has_think:
+            has_nonempty_text = isinstance(content, list) and any(
+                p["type"] == "text" and p.get("text", "") for p in content
+            )
+            if has_nonempty_text:
+                return "<think></think>\n"
```
YujiaBao, line 174: I think there's a subtle mismatch here for the specific case of a historical assistant message that has thinking + non-empty text + tool_calls.

In the [HF Jinja template](https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16/raw/main/chat_template.jinja), the tool_calls branch does:
```jinja2
{%- set c = "<think></think>" ~ c | trim %}
```
Because Jinja2's `|` binds tighter than `~`, the `trim` applies to just `c` (which is `"\ntext"` after the split), stripping the leading `\n` before concatenation. So the template produces `<think></think>text` (no newline).

But the renderer returns `"<think></think>\n"` here, which gives `<think></think>\ntext`.

The non-tool_calls branch in the template applies `| trim` on a separate line after concatenation, so the middle `\n` is preserved there — the renderer is correct for that path. It's only the tool_calls + historical + thinking + non-empty text combo that diverges.

Pretty narrow edge case, but worth noting since the existing tests don't cover it (they use empty text `""` for tool call messages).

tinker_cookbook/renderers/nemotron3.py
```diff
@@ -0,0 +1,325 @@
...
+    RenderContext,
+    RenderedMessage,
+    Role,
+    TextPart,
+    ToolSpec,
+)
+from tinker_cookbook.renderers.qwen3_5 import Qwen3_5Renderer
+
+
+def _format_nemotron3_tool_declaration(tool: ToolSpec) -> str:
```
YujiaBao, line 65: Minor: the HF Jinja template has a `render_extra_keys` macro that outputs additional parameter fields beyond name/type/description/enum (e.g. `default`, `minimum`, `items`). This doesn't handle those, so tools with richer JSON Schema definitions would diverge. Probably fine for now, just flagging it.

tinker_cookbook/model_info.py
```diff
@@ -106,6 +107,19 @@ def get_moonshot_info() -> dict[str, ModelAttributes]:
     }
 
 
+@cache
+def get_nvidia_info() -> dict[str, ModelAttributes]:
+    org = "nvidia"
+    return {
```
YujiaBao, line 113: FYI the [Super model's template](https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16/raw/main/chat_template.jinja) has a `low_effort` flag that appends `\n\n{reasoning effort: low}` to the last user message. No action needed now, just something to be aware of if someone asks for it later.

tinker_cookbook/renderers/nemotron3_test.py
```diff
@@ -0,0 +1,589 @@
...
+    hf_messages = [
+        {"role": "system", "content": system_prompt},
+        {"role": "user", "content": "What's the weather in NYC?"},
+    ]
+
+    if build_mode == "generation":
+        cookbook = nemotron_renderer.build_generation_prompt(prefix_messages + [user_msg]).to_ints()
+        hf = _hf_generation_tokens(nemotron_tokenizer, hf_messages, tools=openai_tools)
+    else:
+        assistant_msg = Message(role="assistant", content="Let me check that for you.")
```
YujiaBao, line 580: CI is failing on transformers 4.57.6 — the 4 HF-comparison tests hit `TypeError: argument 'ids': 'str' object cannot be interpreted as an integer` because the Nemotron-3 tokenizer/template isn't fully compatible with older transformers. These should probably `pytest.skip` when the template isn't available or the tokenizer behaves unexpectedly, so we don't break CI for users on older versions.

opherlieber: Yep, weird though since it passes for me on 4.57.6

tinker_cookbook/renderers/nemotron3.py
opherlieber: added test-case + fix

tinker_cookbook/model_info.py
opherlieber: Will add reasoning-effort in a follow-up

tinker_cookbook/renderers/nemotron3_test.py
opherlieber: fixed, it was just the divergence in 5.0 apply_chat_template returning a dict instead of tokens

tinker_cookbook/renderers/nemotron3.py
opherlieber: Added test+support

PR #494, opened 2026-03-18
Allow older transformers version CI to fail without blocking PRs

YujiaBao:
## Summary
- Add `continue-on-error: true` for transformers 4.57.6 matrix entries in pytest and pyright workflows
- The latest version (5.2.0) remains a hard failure
- Older version test results are still visible in workflow logs for debugging

## Motivation
We support `transformers>=4.57.6,<5.3.0`, but development targets the latest version. Older version failures shouldn't block PRs — developers would have to manually install/uninstall different transformers versions locally to debug issues that are often just stale API differences. With `continue-on-error`, the older version job still runs and surfaces failures clearly in the workflow UI, but doesn't block the PR or show a red X at the top level.

## Test plan
- [ ] Verify PR CI shows green overall even if the 4.57.6 job fails
- [ ] Verify 5.2.0 job failure still blocks (shows red)

### DIFF
--- a/.github/workflows/pyright.yaml
+++ b/.github/workflows/pyright.yaml
@@ -9,6 +9,17 @@ jobs:
   type-check:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        include:
+          - transformers-version: "5.2.0"
+            allow-failure: false
+          - transformers-version: "4.57.6"
+            allow-failure: true
+
+    name: type-check (transformers ${{ matrix.transformers-version }})
+    continue-on-error: ${{ matrix.allow-failure }}
+
     steps:
       - name: checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


YujiaBao:
> what was the delta in time?

PR updated. I think the focus here is actually not about time, but about tolerance on breaking outdated transformer versions. Features with newer models will definitely break older transformers, and if the old tests is causing the failure, we should set it as warnings rather than failure.

.github/workflows/pyright.yaml
```diff
@@ -11,9 +11,14 @@ jobs:
 
     strategy:
       matrix:
-        transformers-version: ["4.57.6", "5.2.0"]
+        include:
+          - transformers-version: "5.2.0"
+            allow-failure: false
+          - transformers-version: "4.57.6"
+            allow-failure: true
```
dphuang2, line 18: seems like an easy way to never catch failures 🤔

dphuang2:
Screenshot 2026-03-18 at 5 16 02 PM

does it make sense to completely ignore? (context: https://pepy.tech/projects/transformers?timeRange=threeMonths&category=version&includeCIDownloads=true&granularity=daily&viewType=line&versions=5.3.0%2C5.2.0%2C5.1.0%2C4.57.6%2C4.57.3)

PR #497, opened 2026-03-19
Add rollout error resilience for RL training

YujiaBao:
## Problem

During RL training, per-trajectory errors (container crashes, sandbox flakes, transient API failures) crash the entire training run. There's no error tolerance or retry mechanism in the rollout pipeline.

## Solution

A pluggable `RolloutStrategy` abstraction that controls how trajectories are collected within a group, with a built-in retry strategy for env-level failures.

### Design Rationale

Per joschu's feedback: skipping failed trajectories causes bias that is hard to reason about in RL training. Instead, **failed trajectories should be retried with fresh environments**. When the retry budget is exhausted, the group fails entirely (cancel remaining + re-raise) to avoid partial-group bias from training on an incomplete set of trajectories. Context overflow issues should be handled at the env level (separate concern, not in scope here).

Per kennyyu's suggestion: the rollout collection logic should be a pluggable strategy interface, since there are many possible approaches (retry, over-provisioning, variance-based collection, etc.).

### Config

```python
# Tolerance for errors during rollouts (container crashes, sandbox flakes, etc.).
# False (default): crash on any error (FailFast).
# True: retry failed trajectories with default budget (RetryOnFailure(max_retries=3)).
# RolloutStrategy instance: custom strategy (e.g. RetryOnFailure(max_retries=5)).
rollout_error_tolerance: bool | RolloutStrategy = False
```

### RolloutStrategy

Pluggable strategy for trajectory collection (`rl/rollout_strategy.py`):

```python
class RolloutStrategy(ABC):
    @property
    def catches_group_errors(self) -> bool: ...
    @abstractmethod
    async def execute(self, env_group_builder, policy) -> RolloutResult: ...
```

The strategy owns the full trajectory collection lifecycle including env creation (via `make_envs()`), so strategies like retry can create fresh envs as needed. Group reward computation and logging remain in `do_group_rollout`.

**Built-in strategies:**
- `FailFast` — original `asyncio.gather(...)` behavior, any error crashes
- `RetryOnFailure(max_retries=3)` — retry failed trajectories with fresh environments using `asyncio.wait(FIRST_COMPLETED)`. Retries start immediately when a failure is detected. When retry budget is exhausted, cancels remaining tasks and re-raises to avoid partial-group bias.

### Error Tracking

- `TrajectoryGroup.rollout_errors: list[RolloutError]` — carries error info through return values (pickle-safe, works across ProcessPoolExecutor/Ray)
- `RolloutErrorCounter` — aggregates in the main event loop only, no shared mutable state
- Cumulative metrics per batch: `rollout_errors/total`, `rollout_errors/{ErrorType}`, `rollout_errors/groups_skipped`

## Scope

This PR handles **env-level failures** (container crashes, sandbox flakes, transient errors) where retrying the trajectory with a fresh environment is appropriate.

**Not in scope:** Context overflow from `prompt + max_tokens > context_window`. Per joschu's feedback, this should be handled at the env level by terminating the episode before exceeding the context limit — not by catching the error after the fact. That fix belongs in a separate PR.

## Files Changed

| File | Changes |
|------|---------|
| `tinker_cookbook/rl/rollout_strategy.py` | **New.** `RolloutStrategy` ABC, `FailFast`, `RetryOnFailure`, config mapping. |
| `tinker_cookbook/rl/types.py` | `RolloutError` dataclass, `rollout_errors` field on `TrajectoryGroup`. |
| `tinker_cookbook/rl/rollouts.py` | `do_group_rollout` delegates to strategy. `RolloutErrorCounter`. |
| `tinker_cookbook/rl/train.py` | Config: `rollout_error_tolerance`. Strategy + counter threaded to all 3 training modes. |
| `tinker_cookbook/rl/metric_util.py` | `RLTestSetEvaluator` accepts strategy, uses `RolloutErrorCounter`. |
| `tinker_cookbook/exceptions.py` | `AllTrajectoriesFailedError` as `TrainingError` subclass. |
| `tinker_cookbook/__init__.py` | Re-export. |
| `tinker_cookbook/recipes/code_rl/train.py` | `rollout_max_retries` CLI field. |
| `tinker_cookbook/rl/rollout_error_resilience_test.py` | **New.** 31 tests. |

## Backward Compatibility

All changes are backward compatible — default behavior is identical to pre-PR:

- `rollout_error_tolerance=False` (default) produces `FailFast` which executes the exact same `asyncio.gather(*[do_single_rollout(...)])` code path as before
- All function signatures add optional params with defaults (`strategy=None`, `error_counter=None`)
- `TrajectoryGroup.rollout_errors` field has `default_factory=list` — existing positional construction unaffected
- `Config` new field has default `False` — all 5 recipes use keyword construction, unaffected
- Distillation module calls `do_group_rollout_and_filter_constant_reward` and `RLTestSetEvaluator` without new kwargs — defaults apply
- Verifiers recipe monkey-patch pattern is pre-existing and unaffected

## Distributed / Multiprocess Safety

- `FailFast` and `RetryOnFailure` are `@dataclass(frozen=True)` with primitive fields — trivially pickleable for `ProcessPoolExecutor`/Ray dispatch via `_RolloutTask`
- `RetryOnFailure` uses `asyncio.create_task` + `asyncio.wait(FIRST_COMPLETED)` which works in the fresh event loop created by `asyncio.run()` in worker processes
- `RolloutErrorCounter` lives only in the main event loop — error info flows back via `TrajectoryGroup.rollout_errors` (pickle-safe return values)
- All three training modes (sync, async, stream minibatch) correctly handle `None` results from error tolerance

## Test plan

- [x] 31 new unit tests (strategies, retry behavior, budget exhaustion, cancellation, pickling, config mapping, error counter, make_envs failure)
- [x] All existing tests pass (758 unit)
- [x] pyright clean, ruff clean

### DIFF
--- a/tinker_cookbook/rl/metric_util.py
+++ b/tinker_cookbook/rl/metric_util.py
@@ -1,5 +1,6 @@
 import asyncio
 import itertools
+import logging
 from collections import defaultdict
 
 import numpy as np
@@ -7,11 +8,14 @@ import tinker
 
 from tinker_cookbook.completers import TinkerTokenCompleter, TokenCompleter
 from tinker_cookbook.eval.evaluators import SamplingClientEvaluator
+from tinker_cookbook.exceptions import AllTrajectoriesFailedError
 from tinker_cookbook.rl.rollout_logging import (
     RolloutSummaryExportConfig,
     write_rollout_summaries_jsonl,
 )
+from tinker_cookbook.rl.rollout_strategy import RolloutStrategy
 from tinker_cookbook.rl.rollouts import (
+    RolloutErrorCounter,
     do_group_rollout,
     do_group_rollout_and_filter_constant_reward,
     get_rollout_executor,
@@ -19,6 +23,8 @@ from tinker_cookbook.rl.rollouts import (
 from tinker_cookbook.rl.types import EnvGroupBuilder, RLDataset, TrajectoryGroup
 from tinker_cookbook.utils import logtree
 from tinker_cookbook.utils.misc_utils import all_same, dict_mean
+
+logger = logging.getLogger(__name__)
 
 
 def _compute_by_group_metrics(trajectory_groups_P: list[TrajectoryGroup], good_thresh: float = 0.5):
@@ -117,11 +123,13 @@ class RLTestSetEvaluator(SamplingClientEvaluator):
         max_tokens: int,
         name: str = "test",
         num_groups_to_log: int = 4,
+        strategy: RolloutStrategy | None = None,
     ):
         self.env_group_builders_P = dataset_to_env_group_builders(dataset)
         self.max_tokens = max_tokens
         self.name = name
         self.num_groups_to_log = num_groups_to_log
+        self.strategy = strategy
 
     async def eval_token_completer(
         self,
@@ -129,33 +137,34 @@ class RLTestSetEvaluator(SamplingClientEvaluator):
         *,
         rollout_summary_export: RolloutSummaryExportConfig | None = None,
     ) -> dict[str, float]:
-        async def run_group_rollout(builder, i):
-            enable_logging = i < self.num_groups_to_log
-            with logtree.optional_enable_logging(enable=enable_logging):
-                return await do_group_rollout(builder, policy)
+        async def run_group_rollout(
+            builder: EnvGroupBuilder, group_idx: int
+        ) -> TrajectoryGroup | None:
+            enable_logging = group_idx < self.num_groups_to_log
+            try:
+                with logtree.optional_enable_logging(enable=enable_logging):
+                    result = await do_group_rollout(
+                        builder,
+                        policy,
+                        strategy=self.strategy,
+                    )
+            except AllTrajectoriesFailedError as e:
+                logger.warning(f"Eval: {e}")
+                result = None
+            except Exception as e:
+                if self.strategy is None or not self.strategy.catches_group_errors:
+                    raise
+                logger.warning(f"Eval rollout error ({type(e).__name__}): {e}")
+                result = None
+            return result
 
-        trajectory_groups_P = await asyncio.gather(
-            *[run_group_rollout(builder, i) for i, builder in enumerate(self.env_group_builders_P)]
+        results = await asyncio.gather(
+            *[
+                run_group_rollout(builder, group_idx)
+                for group_idx, builder in enumerate(self.env_group_builders_P)
+            ]
         )
-        taglist_P = [builder.logging_tags() for builder in self.env_group_builders_P]
-        if rollout_summary_export is not None:
-            sampling_client_steps_P = (
-                [rollout_summary_export.sampling_client_step] * len(trajectory_groups_P)
-                if rollout_summary_export.sampling_client_step is not None
-                else None
-            )
-            write_rollout_summaries_jsonl(
-                rollout_summary_export.path,
-                split=rollout_summary_export.split,
-                iteration=rollout_summary_export.iteration,
-                trajectory_groups_P=trajectory_groups_P,
-                taglist_P=taglist_P,
-                sampling_client_steps_P=sampling_client_steps_P,
-            )
-        metrics = compute_trajectory_metrics(trajectory_groups_P, taglist_P)
-
-        metrics = {f"{self.name}/{k}": v for k, v in metrics.items()}
-        return metrics
+        return self._collect_eval_metrics(results, rollout_summary_export)
 
     async def __call__(
         self,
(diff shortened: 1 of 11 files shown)


tinker_cookbook/rl/rollouts.py
```diff
@@ -106,14 +148,66 @@ async def do_single_rollout(policy: TokenCompleter, env: Env) -> Trajectory:
 
 @logtree.scope_header_decorator("Group Rollout")
 async def do_group_rollout(
-    env_group_builder: EnvGroupBuilder, policy: TokenCompleter
+    env_group_builder: EnvGroupBuilder,
+    policy: TokenCompleter,
+    skip_rollout_errors: bool = False,
```
kennyyu: it seems these are either or:

no skip | allow skip with some threshold fraction

maybe we can use a union instead?

tinker_cookbook/rl/rollouts.py
```diff
@@ -106,14 +148,66 @@ async def do_single_rollout(policy: TokenCompleter, env: Env) -> Trajectory:
 
 @logtree.scope_header_decorator("Group Rollout")
 async def do_group_rollout(
```
kennyyu, line 144: in general, folks may want to explore different strategies on how to do the rollouts.

For example one thing I implemented recently was to allow doing some retries:

```
# Example: by default we do 16 attempts
# user can also specify up to 3 retries if needed
# if any rollout  fails and we still have a budget, try again

            retries_remaining = max_retry
            retry_count = 0
            timed_results: list[tuple[Sequence[Rollout], float]] = []
            pending: set[asyncio.Task[tuple[Sequence[Rollout], float]]] = set(tasks)

            while pending:
                done, pending = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)

                failures: list[Exception] = []
                for task in done:
                    try:
                        timed_results.append(task.result())
                    except Exception as exc:
                        failures.append(exc)

                for exc in failures:
                    if retries_remaining > 0:
                        retries_remaining -= 1
                        retry_count += 1
                        logger.warning("attempt failed (retry %d/%d): %s", retry_count, self.max_retry_spp, exc)
                        retry_task = asyncio.create_task(do_rollout())
                        pending.add(retry_task)
                    else:
                        logger.error("retry budget exhausted, cancelling remaining tasks")
                        for t in pending:
                            t.cancel()
                        await asyncio.gather(*pending, return_exceptions=True)
                        raise exc
```
There's also a bunch of other strategies you could do as well:
- start 20 at the beginning, and exit as soon as you have 16 and cancel the other 4
- keep generating enough trajectories until there's enough variance in reward
- and a lot moer

So instead of trying to put all the logic into this function, maybe we can have some interface/protocol/Callable interface where people can implement different strategies?

YujiaBao: make sense!

tinker_cookbook/rl/metric_util.py
```diff
@@ -117,45 +123,42 @@ def __init__(
...
+        self.strategy = strategy
 
     async def eval_token_completer(
         self,
         policy: TokenCompleter,
         *,
         rollout_summary_export: RolloutSummaryExportConfig | None = None,
     ) -> dict[str, float]:
-        async def run_group_rollout(builder, i):
+        async def run_group_rollout(builder: EnvGroupBuilder, i: int) -> TrajectoryGroup | None:
```
joschu: can you improve the variable name of `i`?

joschu: i see that's not new to this PR :)

tinker_cookbook/rl/rollout_strategy.py
```diff
@@ -0,0 +1,176 @@
...
+    skip_rollout_errors: bool | float | RolloutStrategy,
+) -> RolloutStrategy:
+    """Convert a ``Config.skip_rollout_errors`` value to a :class:`RolloutStrategy`.
+
+    - ``False`` → :class:`FailFast` (crash on any error, the default)
+    - ``True``  → :class:`SkipWithThreshold` with ``min_fraction=0.0``
+      (always keep the group as long as ≥1 trajectory succeeds)
+    - A float in ``(0, 1]`` → :class:`SkipWithThreshold` with that fraction
+    - A :class:`RolloutStrategy` instance → passed through as-is
+    """
```
dphuang2, line 213: Is this `min_fraction` flexibility necessary? When would somebody know the exact number to choose.

tinker_cookbook/rl/train.py
```diff
@@ -1445,9 +1492,19 @@ async def main(
 
     # Create dataset from thunk
     dataset, maybe_test_dataset = await cfg.dataset_builder()
+    # Build rollout strategy and error counter from config
+    strategy = rollout_strategy_from_config(cfg.skip_rollout_errors)
```
dphuang2: can we make this dynamically configurable in an environment using environment variables? `TINKER_SKIP_ROLLOUT_ERRORS=true|false`

tinker_cookbook/rl/rollout_strategy.py
YujiaBao: The sampler may fail randomly. For larger rollouts, we may want to crash only when there are significant amount of errors on rollouts.

tinker_cookbook/rl/train.py
YujiaBao: I think this should not be an env variable as the strategy could be part of the hyper parameter to choose from.

tinker_cookbook/rl/rollout_strategy.py
YujiaBao: mhmm but I do feel that allowing it to be `float` is quite confusing. Made some updates on this design.

tinker_cookbook/rl/data_processing.py
```diff
@@ -20,14 +21,37 @@
...
 
-def compute_advantages(trajectory_groups_P: list[TrajectoryGroup]) -> list[torch.Tensor]:
-    """Compute advantages for each trajectory, centered within groups."""
+def compute_advantages(
+    trajectory_groups_P: list[TrajectoryGroup],
+    error_reward_handler: RolloutErrorRewardHandler | None = None,
+) -> list[torch.Tensor]:
+    """Compute advantages for each trajectory, centered within groups.
+
+    When *error_reward_handler* is provided and a group has rollout errors,
```
YujiaBao: @joschu  thoughts on this?

YujiaBao:
> I feel the min success rate is a little too specific of a knob. Realistically, people don't know how models are going to perform on average for a particular task.

fixed

PR #500, opened 2026-03-19
Fix bug in `ModalSandbox.write_file` method

puma314:
## Summary

**Fixed a bug in `ModalSandbox.write_file`**, which hangs for the full exec timeout (~30-60s) on files because `proc.stdin.write_eof()` buffers the EOF signal without flushing it. Adding `await proc.stdin.drain.aio()` after `write_eof()` ensures the EOF is sent immediately.

## The bug

When writing files via stdin in a Modal sandbox, the sequence is:

1. Write content chunks (each followed by `drain()` ✅)
2. `proc.stdin.write_eof()` — signals end of input
3. ❌ **Missing `drain()`** — the EOF sits in the buffer and is never flushed

The process blocks waiting for stdin to close, and the call hangs until the exec timeout fires.

### How to reproduce

Comment out line 185 in `tinker_cookbook/sandbox/modal_sandbox.py` (`await proc.stdin.drain.aio()` after `write_eof()`) and run the test — `write_file` will take ~30s (the timeout) instead of completing in 1-2s.

## Test plan

- [ ] Run `pytest tinker_cookbook/sandbox/modal_sandbox_test.py` (requires Modal auth) — all 3 tests should pass with write_file completing in <15s
- [ ] Verify regression by commenting out the drain line and observing the timeout hang

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,6 +105,7 @@ markers = [
     "slow: marks tests as slow (deselect with '-m \"not slow\"')",
     "integration: marks tests requiring TINKER_API_KEY and network access",
     "downstream_compat: marks tests that verify public API contracts for downstream consumers",
+    "timeout: per-test timeout in seconds (requires pytest-timeout)",
 ]
 
 [tool.ruff]


tinker_cookbook/sandbox/modal_sandbox_test.py
```diff
@@ -0,0 +1,131 @@
+"""Smoke tests for ModalSandbox operations.
```
YujiaBao: Colocated `_test.py` files are intended to be unit tests that run fast without external dependencies (`pytest tinker_cookbook/`). These tests require a live Modal connection, making them smoke tests — they'd be a better fit in `tests/` where other integration/smoke tests live.

Also, each test spins up a new `ModalSandbox.create()`, which is the slowest part. Consider sharing a single sandbox via a module-scoped async fixture to cut total test time ~3x.

tinker_cookbook/sandbox/modal_sandbox_test.py
```diff
@@ -0,0 +1,131 @@
...
+import pytest
+
+from tinker_cookbook.sandbox.modal_sandbox import ModalSandbox
+
+_has_modal_auth = bool(
+    os.environ.get("MODAL_TOKEN_ID")
+    or os.path.exists(os.path.expanduser("~/.modal.toml"))
+)
+
+requires_modal = pytest.mark.skipif(not _has_modal_auth, reason="Modal not configured locally")
```
YujiaBao: Nit: `~/.modal.toml` existing doesn't guarantee valid auth (could be expired). Fine for skip logic, just noting it.

tinker_cookbook/sandbox/modal_sandbox_test.py
```diff
@@ -0,0 +1,131 @@
...
+        assert elapsed < 15, (
+            f"write_file took {elapsed:.1f}s — likely stdin EOF hang (expected <15s)"
+        )
+
+        # Verify size
+        size_result = await sandbox.run_command("wc -c < /tmp/binary.bin")
+        assert size_result.exit_code == 0
+        assert int(size_result.stdout.strip()) == 256
+
+        print(f"\nwrite_file (binary) latency: {elapsed:.2f}s")
```
YujiaBao: This test's `< 15s` assertion is already covered by `test_write_file_latency`. The relative comparison is interesting as documentation but may not justify the extra sandbox spin-up cost. Consider folding the comparison into a comment/docstring on one of the other tests instead.

tinker_cookbook/sandbox/modal_sandbox_test.py
```diff
@@ -0,0 +1,131 @@
...
+
+async def _timed(coro):
+    """Await a coroutine and return (result, elapsed_seconds)."""
+    start = time.monotonic()
+    result = await coro
+    return result, time.monotonic() - start
+
+
+@requires_modal
+@pytest.mark.asyncio
```
YujiaBao: Minor: if the drain fix regresses, the test hangs for the full 30s `write_file` timeout before the `< 15s` assertion fires. A `@pytest.mark.timeout(20)` on each test would fail faster and more clearly.

tinker_cookbook/sandbox/modal_sandbox_test.py
puma314: Got it, moved the test file and used an async fixture

tinker_cookbook/sandbox/modal_sandbox_test.py
puma314: Yeah removed this test

tinker_cookbook/sandbox/modal_sandbox_test.py
puma314: done

puma314:
@YujiaBao thanks for the comments, I was in the middle of addressing all of your comments before I saw your latest comment to just merge the bug fix. 

Happy for you to take a look at the latest diff that should address your comments OR just remove the test entirely and merge the bug-fix. Will defer to your approach.

pyproject.toml
```diff
@@ -104,6 +105,7 @@ markers = [
     "slow: marks tests as slow (deselect with '-m \"not slow\"')",
     "integration: marks tests requiring TINKER_API_KEY and network access",
     "downstream_compat: marks tests that verify public API contracts for downstream consumers",
+    "timeout: per-test timeout in seconds (requires pytest-timeout)",
```
YujiaBao, line 108: Nit: `pytest-timeout` registers the `timeout` marker automatically — this manual registration isn't needed and the description is slightly misleading (it's a plugin-provided marker, not a custom one). Fine to drop this line.

PR #505, opened 2026-03-20
Add cleanup() lifecycle method to EnvGroupBuilder

puma314:
## Summary

Currently the way cleanup is done for environments with expensive resources is jank and shoe-horned into `compute_group_rewards()`. This poses a problem when a rollout raises an exception mid-episode or the training process is stopped after starting rollouts but before reaching `compute_group_rewards()`, leaving dangling resources.

## Fix

- Adds `async cleanup()` no-op method to `EnvGroupBuilder` base class, called from a `try/finally` in `do_group_rollout()` so resources are released regardless of success or failure
- Moves `HarborEnvGroupBuilder` sandbox cleanup from `compute_group_rewards()` to a `cleanup()` override — previously, if a rollout failed before reaching `compute_group_rewards()`, sandboxes would leak
- No behavior change on the happy path; subclasses without expensive resources inherit the default no-op

## Test plan
- [x] All 34 existing RL unit tests pass
- [ ] Verify Harbor RL recipe still cleans up sandboxes on successful runs
- [ ] Verify sandbox cleanup occurs when a rollout raises an exception mid-episode


tinker_cookbook/rl/rollouts.py
```diff
@@ -109,22 +109,25 @@ async def do_group_rollout(
...
-
-    with logtree.scope_header("Trajectory Details"):
-        for traj_idx, (traj, final_reward) in enumerate(
-            zip(trajectories_G, rewards_G, strict=True)
-        ):
-            with logtree.scope_header(f"Trajectory {traj_idx} Episode"):
-                _log_single_trajectory_details(traj, final_reward)
-
-    return TrajectoryGroup(trajectories_G, list(rewards_G), list(metrics_G))
+    try:
```
YujiaBao, line 158: `make_envs()` should be inside the `try` block. `HarborEnvGroupBuilder.make_envs()` creates sandboxes incrementally in a loop — if sandbox creation fails midway, `self._sandboxes` already has entries but `cleanup()` won't run because we haven't entered the `try` block yet. This is safe since the default `cleanup()` is a no-op, and Harbor's cleanup handles an empty `_sandboxes` list.

```suggestion
    try:
        envs_G: Sequence[Env] = await env_group_builder.make_envs()
```

tinker_cookbook/rl/rollouts.py
```diff
@@ -109,22 +109,25 @@ async def do_group_rollout(
...
+        with logtree.scope_header("Trajectory Details"):
+            for traj_idx, (traj, final_reward) in enumerate(
+                zip(trajectories_G, rewards_G, strict=True)
+            ):
+                with logtree.scope_header(f"Trajectory {traj_idx} Episode"):
+                    _log_single_trajectory_details(traj, final_reward)
+
+        return TrajectoryGroup(trajectories_G, list(rewards_G), list(metrics_G))
+    finally:
+        await env_group_builder.cleanup()
```
YujiaBao, line 181: If `cleanup()` raises, it will mask the original exception from the `try` block. Consider wrapping this in a `try/except` that logs and suppresses, or documenting that `cleanup()` implementations must not raise (like `HarborEnvGroupBuilder` does with its per-sandbox try/except).

tinker_cookbook/rl/types.py
```diff
@@ -124,6 +124,18 @@ async def compute_group_rewards(
...
 
+    async def cleanup(self) -> None:
+        """Clean up resources created by make_envs().
+
+        Called after rollouts and reward computation complete, regardless
+        of success or failure. Override this to release expensive resources
+        like cloud sandboxes, remote browsers, etc.
+
+        Default is a no-op. Implementations should be idempotent (safe to
+        call multiple times).
```
YujiaBao: Worth noting in the docstring that exceptions from `cleanup()` are not caught by `do_group_rollout` — implementations should handle their own exceptions internally (like `HarborEnvGroupBuilder` does with its per-sandbox try/except).

tinker_cookbook/rl/rollouts.py
puma314: good point, fixing

puma314:
@YujiaBao PTAL, I merged in the latest main which I saw added a `strategy` argument to `do_group_rollout`.

PR #506, opened 2026-03-20
Handle context limits and max-tokens truncation in multi-turn RL environments

YujiaBao:
## Problem

In multi-turn RL (e.g., code_rl with tool use), the accumulated conversation context grows each turn. There were four related problems with how context/length limits were handled:

**P1: Context overflow crashes training.** The existing `max_trajectory_tokens` check only tested `observation.length > limit`, but the next sampling call sends `observation + max_tokens` to the Tinker API. When `observation + max_tokens > max_model_len`, the server returns HTTP 400 `PromptTooLongException` — an unrecoverable crash mid-training.

**P2: Max-tokens truncation misclassified as parse error.** When a model hits `max_tokens` without producing a stop sequence (`stop_reason="length"`), the `stop_reason` was discarded. The truncated response (missing stop token) then failed `parse_response()`, applying `failed_parse_reward` (-1.0 or -0.1) instead of treating it as a capacity issue.

**P3: Overflow reward not configurable.** The reward for context overflow was hardcoded to `0.0`. Users couldn't choose between neutral (0) or negative (penalize verbose trajectories).

**P4: Overflow check fired on completed episodes.** When the model gave a final answer (episode done), the context overflow check still fired on the rendered conversation. If the full conversation exceeded the limit, the real reward from `reward_fn` was silently replaced with the overflow reward — even though there was no next sampling call to protect against.

### Example (P1)

With Qwen3.5-4B (32K context), `max_tokens=4096`:
```
Turn 1: ob=5K   → check: 5K < 32K ✓  → sample: 5K+4K=9K  < 32K ✓
Turn 2: ob=18K  → check: 18K < 32K ✓ → sample: 18K+4K=22K < 32K ✓
Turn 3: ob=29K  → check: 29K < 32K ✓ → sample: 29K+4K=33K > 32K 💥
```

## Before (main)

```
do_single_rollout(policy, env)
│
│  1. ob, stop = env.initial_observation()
│     └─ EnvFromMessageEnv: render messages → model_input
│        (no size check)
│
│  LOOP:
│  2. ac = policy(ob, stop)
│     └─ TinkerTokenCompleter.sample_async(prompt=ob, max_tokens=M)
│        │
│        ├─ stop_reason="stop"   → normal (hit stop sequence)
│        ├─ stop_reason="length" → TRUNCATED (hit max_tokens)
│        │   ⚠ stop_reason is DISCARDED — not captured on TokensWithLogprobs
│        │
│        └─ Server validates: prompt_tokens + max_tokens > max_model_len
│           → 💥 HTTP 400 PromptTooLongException → tinker.BadRequestError
│              (unrecoverable, crashes the rollout)
│
│  3. step_result = env.step(ac.tokens)     ← only tokens, no stop_reason
│     └─ EnvFromMessageEnv:
│        a. parse_response(tokens)
│        │  ├─ If truncated (no stop token): success=False
│        │  │   → reward = failed_parse_reward (-1.0 or -0.1)
│        │  │     ⚠ WRONG: conflates parse errors with max_tokens truncation
│        │  │     ⚠ BUT: previous turns' logprobs ARE preserved (good)
│        │  │
│        │  └─ If normal: success=True → continue
│        │
│        b. msg_env.step(message) → reward, episode_done, next_messages
│        c. render(next_messages) → next_observation
│        d. CHECK: next_observation.length > max_trajectory_tokens?
│           ├─ YES → reward=0, done=True
│           │   ⚠ TOO LATE: doesn't account for max_tokens budget
│           │   ⚠ Next iteration's policy() call can still overflow the API
│           │   ⚠ ALSO FIRES WHEN episode_done=True, overriding real reward
│           └─ NO  → continue
│
│  4. ob = step_result.next_observation → goto LOOP step 2
```

## After (this PR)

```
do_single_rollout(policy, env)
│
│  1. ob, stop = env.initial_observation()
│     └─ EnvFromMessageEnv: render messages → model_input
│        CHECK 1: model_input.length + max_generation_tokens > max_trajectory_tokens?
│        ├─ YES → raise ValueError
│        │        (prompt too long — config/data bug, no logprobs to preserve)
│        └─ NO  → return (model_input, stop)
│
│  LOOP:
│  2. ac = policy(ob, stop)
│     └─ TinkerTokenCompleter.sample_async(prompt=ob, max_tokens=M)
│        │
│        ├─ stop_reason="stop"   → ✅ normal
│        └─ stop_reason="length" → ✅ CAPTURED on ac.stop_reason
│        │
│        └─ SAFE: we verified ob.length + M ≤ limit, so no API 400
│
│  3. step_result = env.step(ac.tokens, stop_reason=ac.stop_reason)
│     └─ EnvFromMessageEnv:
│        ┌──────────────────────────────────────────────────────────┐
│        │  CHECK 2: stop_reason == "length"?                       │
│        │  Model hit max_tokens without producing a stop sequence  │
│        │  ├─ YES → reward = context_overflow_reward (configurable)│
│        │  │        done = True                                    │
│        │  │        metric: max_tokens_reached = 1.0               │
│        │  │        msg_env.step() NOT called (short-circuit)      │
│        │  │        Previous turns' logprobs preserved for training │
│        │  └─ NO  → continue to parse                              │
│        └──────────────────────────────────────────────────────────┘
│        │
│        a. parse_response(tokens)
│           ├─ success=False → reward = failed_parse_reward
│           │   (now only genuine parse errors, not max_tokens truncation)
│           └─ success=True  → continue
│        │
│        b. msg_step = msg_env.step(message)
│           → reward, episode_done, next_messages
│        c. render(next_messages) → next_observation
│        │
│        ┌──────────────────────────────────────────────────────────┐
│        │  CHECK 3: episode NOT done AND                           │
│        │           next_ob + max_gen_tokens > limit?              │
│        │                                                          │
│        │  ├─ episode_done=True → SKIP CHECK                      │
│        │  │   No next sampling call; real reward preserved.       │
│        │  │                                                       │
│        │  ├─ Exceeds limit → reward = context_overflow_reward     │
│        │  │   done = True, metric: context_overflow = 1.0         │
│        │  │   Previous turns' logprobs preserved for training     │
│        │  │                                                       │
│        │  └─ Within limit → continue normally                     │
│        └──────────────────────────────────────────────────────────┘
│
│  4. ob = step_result.next_observation → goto LOOP step 2
│     └─ SAFE: ob.length + M ≤ limit (guaranteed by check 3)
```

### The three checks

| Check | Where | Condition | Action |
|-------|-------|-----------|--------|
| **1** | `initial_observation()` | `prompt + max_gen_tokens > limit` | `ValueError` (config/data bug) |
| **2** | `step()` entry | `stop_reason == "length"` | Terminate, `context_overflow_reward`, metric: `max_tokens_reached` |
| **3** | `step()` after env | `NOT episode_done AND next_ob + max_gen_tokens > limit` | Terminate, `context_overflow_reward`, metric: `context_overflow` |

**Invariant:** At every point where `policy(ob, stop)` is called, we have already verified that `ob.length + max_generation_tokens ≤ max_trajectory_tokens` and the API call cannot overflow.

## Design rationale

Per discussion with joschu: context overflow should be handled **in the env itself** as graceful episode termination, not via exception handling or retry strategies. This is decoupled from rollout error resilience (#497) which handles env crashes (container failures, sandbox flakes).

When an episode terminates due to context overflow or max_tokens truncation:
- Previous turns' logprobs are **preserved** in the trajectory for training
- A configurable reward (`context_overflow_reward`, default -0.1) is assigned
- A metric (`context_overflow` or `max_tokens_reached`) is logged for monitoring

## Files changed

| File | Changes |
|------|---------|
| `completers.py` | Capture `stop_reason` from `tinker.SampledSequence` onto `TokensWithLogprobs` |
| `rl/types.py` | Add `stop_reason` and `**kwargs` to `Env.step()` protocol |
| `rl/rollouts.py` | Pass `stop_reason` from sampled action to `env.step()` |
| `rl/message_env.py` | Add `max_generation_tokens`, `context_overflow_reward`; handle `stop_reason="length"`; skip overflow check when `episode_done` |
| `rl/message_env_test.py` | 12 new tests (generation budget, overflow reward, max_tokens_reached, episode_done skip) |
| `tool_use/agent_tool_message_env.py` | Thread `max_generation_tokens`, `context_overflow_reward` through `build_agent_tool_env` |
| `recipes/code_rl/` | Thread params through builders; default `max_tokens` 5→4096 |
| `recipes/search_tool/` | Thread `max_generation_tokens`, `context_overflow_reward` through builders |
| `recipes/harbor_rl/` | Thread `max_generation_tokens`, `context_overflow_reward` through builders |
| All other envs | Accept `**kwargs` in `step()` for forward compatibility |
| `tests/downstream_compat/test_rl_types.py` | Update `Env.step` signature check |

## Backward compatibility

- `max_generation_tokens=None` (default): no generation budget check, degrades to original `observation > limit`
- `context_overflow_reward=-0.1` (default): matches the existing `failed_parse_reward` default in `build_agent_tool_env`, so max_tokens truncation behavior is preserved for existing users
- `stop_reason="stop"` (default on `TokensWithLogprobs`): no change for custom completers
- `Env.step(**kwargs)`: existing custom envs need to add `**kwargs` to their `step()` signature (one-line change). The `**kwargs` also provides forward compatibility for future parameters.

## Usage

```python
build_agent_tool_env(
    ...,
    max_trajectory_tokens=32 * 1024,
    max_generation_tokens=4096,        # reserve room for generation
    context_overflow_reward=-0.1,      # or 0.0 for neutral, -1.0 to penalize
)
```

For code_rl, `max_generation_tokens` is automatically set from `Config.max_tokens`. The `context_overflow_reward` is configurable at the env builder level for all three tool-use recipes (code_rl, search_tool, harbor_rl).

## Test plan

- [x] All 26 `message_env_test.py` tests pass (15 existing + 11 new)
- [x] 811 tests pass across the full test suite
- [x] 199 downstream compat tests pass
- [x] ruff format/check clean
- [ ] Manual test: code_rl with Qwen3.5-4B and `max_tokens=4096` — no more 400 errors

### DIFF
--- a/tinker_cookbook/tool_use/agent_tool_message_env.py
+++ b/tinker_cookbook/tool_use/agent_tool_message_env.py
@@ -150,7 +150,7 @@ def build_agent_tool_env(
             terminates if the trajectory + generation budget would exceed
             *max_trajectory_tokens*, preventing context overflow errors.
         context_overflow_reward: Reward assigned when the episode is terminated due to
-            context overflow. Defaults to 0.0.
+            context overflow. Defaults to -0.1.
 
     Returns:
         An EnvFromMessageEnv ready for RL training.


puma314:
+1 I ran into this as well, would be great to see this merged!

tinker_cookbook/rl/message_env.py
```diff
@@ -77,13 +82,44 @@ async def _render_in_thread(self, messages: list[Message], **kwargs) -> tinker.M
...
+        # If the model hit max_tokens without producing a stop sequence, terminate
+        # the episode early. Previous turns' logprobs are preserved in the trajectory.
+        if stop_reason == "length":
+            return types.StepResult(
+                reward=self.context_overflow_reward,
+                episode_done=True,
+                next_observation=tinker.ModelInput.empty(),
+                next_stop_condition=self._base_stop_condition,
+                metrics={"max_tokens_reached": 1.0},
+            )
```
dphuang2, line 121: do we need to explicitly need to create a `context_overflow_reward` attribute in `EnvGroupBuilder` and `EnvFromMessageEnv`?

Why not pass `stop_reason` to the environment itself and let the environment do what it wants with the `stop_reason`? That way we don't have to add a new attribute `context_overflow_reward` and we can remove some hard-coded business logic in `EnvFromMessageEnv`

YujiaBao: Good question! We do pass `stop_reason` through to `Env.step()` — the protocol is now `step(self, action, *, stop_reason=..., **kwargs)`, so any direct `Env` implementation can handle it however they want.

The reason `EnvFromMessageEnv` has a built-in default is that it's the adapter layer between token-level concerns (like `stop_reason`, parse errors, context overflow) and message-level `MessageEnv` implementations. All three tool-use recipes (code_rl, search_tool, harbor_rl) go through `build_agent_tool_env()` → `EnvFromMessageEnv` → `AgentToolMessageEnv`, and the right behavior is the same for all of them: terminate the episode, preserve previous turns' logprobs, assign a configurable reward. `MessageEnv` authors shouldn't need to know about `stop_reason` — they work at the message level.

Re `context_overflow_reward` being configurable — I think this is important. Different setups may want different signals: -0.1 (default, matches `failed_parse_reward`), 0.0 (neutral), or more negative (penalize verbose trajectories). Hardcoding it would require subclassing to change.

Direct `Env` implementations (ProblemEnv, game envs, etc.) receive `stop_reason` via `**kwargs` and can opt in to handling it if needed — but for now they don't, since truncation already produces natural penalties through their own logic (failed format checks, illegal moves, etc.).

tinker_cookbook/tool_use/agent_tool_message_env.py
```diff
@@ -144,6 +146,11 @@ def build_agent_tool_env(
         max_turns: Maximum turns before episode ends.
         failed_parse_reward: Reward when model output fails to parse.
         max_trajectory_tokens: Maximum tokens in trajectory before terminating episode.
+        max_generation_tokens: Maximum tokens per generation. When set, the episode
+            terminates if the trajectory + generation budget would exceed
+            *max_trajectory_tokens*, preventing context overflow errors.
+        context_overflow_reward: Reward assigned when the episode is terminated due to
+            context overflow. Defaults to 0.0.
```
dphuang2:
```suggestion
        context_overflow_reward: Reward assigned when the episode is terminated due to
            context overflow. Defaults to -0.1.
```

dphuang2 (approved):
nice!

PR #507, opened 2026-03-20
Fix crash when all RL advantages are zero in a batch

YujiaBao:
## Summary
- Guard `forward_backward` and `optim_step` when `datums_D` is empty (all questions in the batch had zero advantages)
- Log a warning so users can see when this happens
- Reward metrics still logged for observability

## Test plan
- [x] Run rl_loop with a small model where all-zero-advantage batches are likely (e.g., Qwen3-4B with `batch_size=128`)
- [x] Verify training continues past previously-crashing batches

### DIFF
--- a/tinker_cookbook/recipes/rl_loop.py
+++ b/tinker_cookbook/recipes/rl_loop.py
@@ -226,7 +226,7 @@ def main(config: Config):
                 datums_D.append(datum)
 
         # Training step
-        if not datums_D:
+        if len(datums_D) == 0:
             logger.warning("Batch %d: all advantages zero, skipping training step", batch_idx)
         else:
             fwd_bwd_future = training_client.forward_backward(


tinker_cookbook/recipes/rl_loop.py
```diff
@@ -226,13 +226,18 @@ def main(config: Config):
                 datums_D.append(datum)
 
         # Training step
-        fwd_bwd_future = training_client.forward_backward(datums_D, loss_fn="importance_sampling")
-        optim_step_future = training_client.optim_step(adam_params)
-        _fwd_bwd_result = fwd_bwd_future.result()
-        optim_result = optim_step_future.result()
+        if not datums_D:
```
dphuang2: Is there a more explicit check? One that doesn’t assume the reader knows what falsy means

tinker_cookbook/recipes/rl_loop.py
dphuang2, line 199: are we not checking it here? Also, we should probably log a warning here

dphuang2: also, isn't this check wrong? what if all advantages == 1.0? Isn't that the same issue for GRPO?

YujiaBao: this is fine because we are just skipping the questions if there is no diff in rewards among the group -> no signal for the model to learn

PR #512, opened 2026-03-21
Extract OpenAI compat utilities from LiteLLM provider into renderers/openai_compat

luyuzhe111:
## Summary                                                                                                 
                                                                                                 
- Extract OpenAI ↔ Tinker conversion utilities (`openai_messages_to_tinker`, `openai_tools_to_tinker`,     
`prepare_messages_with_tools`, `sample_chat_completion`, `sampling_result_to_openai_dict`,                 
`extract_sampling_params`, `SamplingResult`) from `third_party/litellm/provider.py` into a new             
`renderers/openai_compat` module.
- Re-export the new public API from `renderers/__init__.py`.
- Move corresponding tests to colocated `renderers/openai_compat_test.py`.

## Motivation                                                                                              
                                                                                                             
These helpers were originally written as private (`_`-prefixed) functions inside the LiteLLM provider. However, [rllm-org/rllm#451](https://github.com/rllm-org/rllm/pull/451) — which adds tool-use support for the Tinker renderer path in rllm — needs the same OpenAI ↔ Tinker conversion logic. Rather than duplicating the code across repos, this PR lifts the utilities into a shared module that both the LiteLLM provider and rllm's Tinker engine, among many others, can import.

## Type of change 

- Refactor                                                                                             
                                                                         
## What changed                                                                                            
                                                                                                             
- New module `tinker_cookbook/renderers/openai_compat.py` with 7 public functions/types (previously private
 in LiteLLM provider)
- `tinker_cookbook/third_party/litellm/provider.py` loses ~200 lines of extracted code and imports from the
 new module                                                                                                
- Tests relocated from `provider_test.py` to `openai_compat_test.py` following colocated `*_test.py`
convention                                                                                                 
- `renderers/__init__.py` updated with re-exports and `__all__` entries
                                                                                                             
## Validation

- Followed the contribution guide to run all tests.      
                                                                                           
---

### DIFF
--- a/tinker_cookbook/third_party/openai_compat_test.py
+++ b/tinker_cookbook/third_party/openai_compat_test.py
@@ -13,7 +13,7 @@ from tinker_cookbook.third_party.openai_compat import (
 # ---------------------------------------------------------------------------
 
 
-class TestOpenaiMessagesToTinker:
+class TestOpenAIMessagesToTinker:
     def test_basic_messages(self) -> None:
         messages = [
             {"role": "system", "content": "You are helpful."},
@@ -69,7 +69,7 @@ class TestOpenaiMessagesToTinker:
 # ---------------------------------------------------------------------------
 
 
-class TestOpenaiToolsToTinker:
+class TestOpenAIToolsToTinker:
     def test_basic_tool(self) -> None:
         tools = [
             {


YujiaBao:
**Suggested restructuring: `third_party/openai_compat` instead of `renderers/openai_compat`**

These conversion utilities aren't really part of the rendering system — they're an API format translation layer that happens to *use* renderers. Placing them in `renderers/` and re-exporting from `renderers/__init__` makes them look like core rendering functionality.

Additionally, the support and testing standards for `third_party/` integrations are likely going to differ from the core renderer library — and the dependency footprint may differ too. It's cleaner to keep them apart so we don't blur that boundary.

Suggested layout:

```
tinker_cookbook/third_party/
├── __init__.py
├── openai_compat.py          # shared OpenAI ↔ Tinker conversion (this PR)
├── litellm/                   # LiteLLM provider (imports from openai_compat)
│   ├── __init__.py
│   └── provider.py
└── rllm/                      # future: rllm Tinker engine support
    └── ...
```

This keeps the third-party integration boundary clear: `openai_compat` is shared infra for any OpenAI-compatible consumer (LiteLLM, rllm, etc.), while each integration gets its own subpackage. The `renderers/__init__.py` re-exports and `__all__` additions would be dropped.

luyuzhe111:
@YujiaBao Thanks for the detailed feedback and explanation! They make a lot of sense. We actually don't need a separate rllm folder since it's OpenAI compatible, and the existing functionalities already suffice.

I have updated the PR based on the review.

YujiaBao (changes requested):
Thanks for moving this to `third_party/` — that addresses the location concern.

Looking at the module again, I think we're exposing too much as public API. Right now it mixes pure format conversion with orchestration, and I'd rather keep the public surface tight.

The converters make sense as public API — `openai_messages_to_tinker()`, `openai_tools_to_tinker()`, `sampling_result_to_openai_dict()`. These are stateless, don't depend on the tinker SDK, and are clearly "format conversion."

The rest I'd keep private:

- `sample_chat_completion()` is really a render→sample→parse pipeline, not a format converter. We already have this pattern in `TinkerMessageCompleter`, and rllm has its own sampling engine — I don't think anyone needs a second public pipeline function living in "openai_compat." It also pulls in `tinker.SamplingClient` as a dependency for the module.
- `SamplingResult` only exists to pass data between `sample_chat_completion` and `sampling_result_to_openai_dict`. If the pipeline stays private, this can too — either keep it in the provider or have `sampling_result_to_openai_dict` take explicit args.
- `prepare_messages_with_tools()` bundles conversion with renderer interaction. Callers can just do `openai_tools_to_tinker()` + `renderer.create_conversation_prefix_with_tools()` themselves.
- `extract_sampling_params()` is fine but it's ~15 lines — I'd keep it private for now and promote later if there's demand.

General principle: `openai_compat` = stateless format conversion, no tinker SDK dependency. Orchestration belongs closer to the consumer.

luyuzhe111:
@YujiaBao thanks for the catch. pulling in tinker sdk dependencies in `openai_compat` is definitely not ideal. i have updated the PR. only `openai_messages_to_tinker()` and `openai_tools_to_tinker()` are factored out now!

YujiaBao:
Looks good — thanks for iterating on the location and public surface.

Issue #401, opened 2026-03-08
ConflictError (409) on checkpoint save retry should be handled internally


ditonogo:
Summary

  When save_checkpoint_async is retried (e.g. due to a transient network issue), the server returns a 409 Conflict because
  the checkpoint was already persisted by the first attempt. The SDK surfaces this as an unhandled tinker.ConflictError
  instead of treating it as an idempotent success.

  Environment

  - tinker SDK version: 0.14.0
  - Python: 3.12

  Stack trace 

```
  File "tinker/lib/public_interfaces/training_client.py", line 527, in _save_state_async
      future = await self.holder.execute_with_retries(_send_request)
  File "tinker/lib/internal_client_holder.py", line 416, in execute_with_retries
      raise e
  File "tinker/lib/internal_client_holder.py", line 377, in execute_with_retries
      return await func(*args, **kwargs)
  File "tinker/lib/public_interfaces/training_client.py", line 522, in _send_request
      return await client.weights.save(
  File "tinker/resources/weights.py", line 108, in save
      return await self._post(
  File "tinker/_base_client.py", line 1031, in request
      raise self._make_status_error_from_response(err.response) from None
  tinker.ConflictError: Error code: 409 - Checkpoint with the same name already exists.
      Please choose a different name to avoid overwriting.
```

  Expected behavior

  execute_with_retries should catch ConflictError (409) for weights.save and treat it as a successful no-op, since the
  conflict means the previous attempt already persisted the checkpoint. The error should not propagate to the caller.

  Suggested fix

  In the retry logic (internal_client_holder.py:execute_with_retries), add a special case for 409 on idempotent save
  operations:

```
  try:
      return await func(*args, **kwargs)
  except ConflictError:
      # Previous attempt succeeded; treat as idempotent success.
      return ...
```

  Alternatively, the server could support an If-None-Match / upsert semantic so retries are naturally idempotent.

dphuang2:
thanks for reporting this! just curious, how often were you seeing this issue? And what sort of transient network issue was it?

ditonogo:
I just observed once for last dozens of trainings.  The network is quite stable most of times.

dphuang2:
@ditonogo do you mind sharing a session ID next time this happens? Then we can investigate more concretely.

ditonogo:
> [@ditonogo](https://github.com/ditonogo) do you mind sharing a session ID next time this happens? Then we can investigate more concretely.

[...]

PR #513, opened 2026-03-21
Handle checkpoint save 409 conflicts with idempotent local recovery fix checkpoint conflict

nishanthp:
This PR makes checkpoint saving resilient to 409 Conflict errors when a checkpoint with the same name already exists (issue #401). Instead of failing immediately, save_checkpoint_async now attempts idempotent recovery from local checkpoints.jsonl.

  ## What changed

- Updated tinker_cookbook/checkpoint_utils.py:
- Added conflict-aware save path for each checkpoint kind (state, sampler).
- On tinker.ConflictError, recover the existing checkpoint path from the most recent matching name in local checkpoints.jsonl.
- Preserve current behavior (re-raise) if no recoverable local record exists.
- Avoid duplicate checkpoints.jsonl entries when conflict recovery resolves to an identical record.
- Added tests in tinker_cookbook/checkpoint_utils_test.py:
   - test_save_checkpoint_async_recovers_conflict_from_existing_state_record
   - test_save_checkpoint_async_conflict_without_local_record_raises

  ## Why

  Training can be interrupted/retried around checkpoint boundaries. Reusing checkpoint names should be safe and idempotent when local metadata already confirms the checkpoint path.
  This improves robustness of resume/retry flows across training loops that share checkpoint_utils.save_checkpoint_async.

  ## Validation

- uv run pytest -q tinker_cookbook/checkpoint_utils_test.py
- uv run ruff check tinker_cookbook/checkpoint_utils.py tinker_cookbook/checkpoint_utils_test.py

Both pass locally.

## Scope / compatibility

- No public function signatures changed.
- Default behavior is unchanged except for conflict recovery in idempotent cases.
- If recovery is not possible, errors still propagate.

### DIFF
--- a/tinker_cookbook/checkpoint_utils.py
+++ b/tinker_cookbook/checkpoint_utils.py
@@ -326,6 +326,72 @@ def get_last_checkpoint(log_dir: str, required_key: str = "state_path") -> Check
         return None
 
 
+def _find_last_checkpoint_by_name(log_dir: str, name: str) -> CheckpointRecord | None:
+    """Return the most recent checkpoint record with the given name."""
+    checkpoints = load_checkpoints_file(log_dir)
+    for checkpoint in reversed(checkpoints):
+        if checkpoint.name == name:
+            return checkpoint
+    return None
+
+
+def _recover_path_from_conflict(
+    *,
+    log_path: str,
+    checkpoint_name: str,
+    path_key: Literal["state_path", "sampler_path"],
+) -> str | None:
+    """Recover existing checkpoint path from local records after a 409 conflict."""
+    existing = _find_last_checkpoint_by_name(log_path, checkpoint_name)
+    if existing is None:
+        return None
+    recovered_path = existing.get(path_key)
+    if isinstance(recovered_path, str):
+        return recovered_path
+    return None
+
+
+async def _save_one_kind_with_conflict_recovery(
+    *,
+    training_client: tinker.TrainingClient,
+    checkpoint_name: str,
+    ttl_seconds: int | None,
+    log_path: str,
+    kind: Literal["state", "sampler"],
+) -> tuple[str, bool]:
+    """Save one checkpoint kind and recover idempotently from name conflicts."""
+    try:
+        if kind == "state":
+            response = await (
+                await training_client.save_state_async(checkpoint_name, ttl_seconds=ttl_seconds)
+            ).result_async()
+        else:
+            response = await (
+                await training_client.save_weights_for_sampler_async(
+                    checkpoint_name, ttl_seconds=ttl_seconds
+                )
+            ).result_async()
+        return response.path, False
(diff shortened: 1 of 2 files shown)


nishanthp:
Tagging a reviewer as I don't have a way to add reviewers @YujiaBao

YujiaBao:
Thanks for the PR — the problem you identified is real. RL async training resume does hit a 409 when `do_async_training` tries to re-save the initial checkpoint at `start_batch`.

After digging into the code, I think we can fix this more simply at the call site in `rl/train.py:708` rather than adding generic 409-recovery to `save_checkpoint_async`. The sync training path (`save_checkpoint_and_get_sampling_client`) already avoids this — it checks `i_batch > start_batch` before saving and falls back to `save_weights_and_get_sampling_client_async()` when it's the initial call. The async path just needs similar treatment: check for an existing checkpoint record before attempting the save.

That keeps the fix to ~5 lines at the one call site that has the bug, avoids changing the behavior of `save_checkpoint_async` (which has 23+ callers), and avoids the parallelism regression for `kind="both"` that the current approach introduces.

I'm going to put up a separate PR with this approach. Feel free to close this one once that's up — happy to discuss if you see cases I'm missing.

nishanthp:
@YujiaBao   

Thanks for the detailed review and yeah agreed on scope of change. 
You’re right that this bug is at the async RL call site, and the sync path already has the correct guard pattern.

I can update this PR to use the same call-site fix in rl/train.py (around the do_async_training initial checkpoint save), and revert the generic save_checkpoint_async changes so behavior stays unchanged for other callers and we avoid the kind="both" parallelism regression.

I can push that revision here so we can keep the discussion in one place?

YujiaBao:
https://github.com/thinking-machines-lab/tinker-cookbook/pull/514

YujiaBao:
Thanks for offering to revise, @nishanthp. After digging deeper, we realized the original premise was wrong — training resume creates a new training run on the Tinker server, so checkpoint names from the old run can't conflict with the new run. The 409 isn't a resume problem.

Looking at issue #401, the actual bug is in the SDK's internal retry logic: `execute_with_retries` retries a `save_state_async` call after a transient network error, but the first attempt already succeeded server-side. The retry hits a 409 because the checkpoint already exists. The right fix is in the SDK — `execute_with_retries` should treat 409 on save operations as idempotent success.

Closing #514 as well since it was based on the same incorrect assumption. Thanks for surfacing this — it led us to understand the real issue.

nishanthp:
I see, since the cookbook creates a new training run on the Tinker server, checkpoint names from the previous run don’t share the same run namespace. So, the 409 conflict isn’t a cross-run resume collision. That makes sense. 

I’ve opened a PR in the SDK repo with a fix  https://github.com/thinking-machines-lab/tinker/pull/21. Please take a look.
Thanks @YujiaBao

PR #514, opened 2026-03-22
Fix 409 Conflict on async RL training resume

YujiaBao:
## Summary
- When async RL training resumes, `do_async_training` unconditionally saves an initial checkpoint at `start_batch` to get a `sampler_path` — but that name already exists on the server from the original run, causing a 409 Conflict.
- Fix: check for an existing checkpoint record before saving, reusing it on resume. This matches how the sync path (`save_checkpoint_and_get_sampling_client`) already avoids this issue.
- Adds a smoke test that runs async training, simulates a crash by trimming `checkpoints.jsonl`, and verifies resume succeeds.
- Supersedes #513 with a simpler approach scoped to the one call site that triggers the bug.

## Test plan
- [x] Unit tests pass (`tinker_cookbook/checkpoint_utils_test.py` — 13 passed, `tinker_cookbook/rl/` — 62 passed)
- [x] Async RL smoke test passes (`tests/recipes/test_recipe_math_rl.py::test_math_rl_async`)
- [x] New resume smoke test passes (`tests/recipes/test_recipe_math_rl.py::test_math_rl_async_resume`)

### DIFF
--- a/tests/recipes/test_recipe_math_rl.py
+++ b/tests/recipes/test_recipe_math_rl.py
@@ -1,21 +1,26 @@
+import json
+import tempfile
+from pathlib import Path
+
 import pytest
 
 from tests.helpers import run_recipe
 
 MODULE = "tinker_cookbook.recipes.math_rl.train"
 
+_COMMON_ARGS = [
+    "model_name=Qwen/Qwen3.5-4B",
+    "groups_per_batch=8",
+    "group_size=4",
+    "max_tokens=5",
+]
+
 
 @pytest.mark.integration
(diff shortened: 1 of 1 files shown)


YujiaBao:
Closing this — after investigating issue #401 more carefully, the 409 Conflict is not caused by training resume (which creates a new server-side training run with its own checkpoint namespace). The actual bug is in the Tinker SDK's `execute_with_retries`: a transient network error causes a retry of `save_state_async`, but the first attempt already succeeded, so the retry hits a 409. The fix belongs in the SDK, not the cookbook.

PR #517, opened 2026-03-22
Group per-iteration output files into subdirectories

joschu:
## Summary

- Adds `iteration_dir()` utility that creates and returns `log_path/iteration_NNNNNN/` subdirectories
- Moves all per-iteration files (logtree HTML/JSON, rollout summaries, timing gantt) into these subdirectories
- Applies to all training loops: RL, SL, DPO, and distillation
- Updates docs (`rl-logging.mdx`), AGENTS.md, and the logging skill to reflect the new layout

**Before:** 308 files flat in `log_path/` for an 84-iteration RL run
**After:** 5 top-level files + 84 tidy subdirectories

## Test plan

- [x] All 773 unit tests pass
- [x] Imports verified for all training modules
- [ ] Run an RL training job and verify iteration subdirectories are created with correct file layout


tinker_cookbook/rl/types.py
```diff
@@ -17,7 +17,7 @@
 Observation: TypeAlias = tinker.ModelInput
 Logprobs: TypeAlias = list[float]
 Metrics: TypeAlias = dict[str, float | int]
-Logs: TypeAlias = dict[str, str | int | float]
+Logs: TypeAlias = dict[str, Any]
```
joschu: might roll this back -- this was for a change that's no longer in the PR

tinker_cookbook/supervised/train.py
```diff
@@ -358,9 +359,9 @@ async def finish_and_log(submitted: SubmittedBatch, window: trace.IterationWindo
         submitted.metrics.update(window.get_timing_metrics())
         window.write_spans_jsonl(log_path / "timing_spans.jsonl", step=submitted.step)
         if config.span_chart_every > 0 and submitted.step % config.span_chart_every == 0:
-            trace.save_gantt_chart_html(
-                window, submitted.step, log_path / f"timing_gantt_{submitted.step:06d}.html"
-            )
+            iter_dir = iteration_dir(log_path, submitted.step)
+            assert iter_dir is not None
```
YujiaBao: The RL code handles `iter_dir is None` gracefully, but the SL/DPO/distillation paths just `assert iter_dir is not None`. That'll be a confusing crash if `log_path` is ever unset, and asserts get stripped under `-O`. Can we match what the RL code does, or at least add a message to the assert?

tinker_cookbook/utils/misc_utils.py
```diff
@@ -93,3 +94,17 @@ def concat_lists(list_of_lists: list[list[Any]]) -> list[Any]:
...
+def iteration_dir(log_path: str | Path | None, step: int) -> Path | None:
+    """Return the per-iteration subdirectory, creating it if it doesn't exist.
+
+    Output files for each training step are grouped under
+    ``log_path/iteration_NNNNNN/`` to keep the top-level directory manageable.
+    Returns ``None`` when *log_path* is ``None`` or empty.
+    """
+    if not log_path:
+        return None
+    d = Path(log_path) / f"iteration_{step:06d}"
```
YujiaBao: This always calls `mkdir()`, so you'll get empty `iteration_NNNNNN/` dirs even when nothing gets written (e.g. `num_groups_to_log=0` and `rollout_json_export=False`). Not a big deal, but worth considering deferring the mkdir to the actual write sites.

YujiaBao (approved):

tinker_cookbook/supervised/train.py
joschu: resolved

PR #519, opened 2026-03-23
feat: add trajectory_hook to Env for post-rollout callbacks

puma314:
## Summary

- Add `trajectory_hook` as an optional attribute on `Env` (defaults to `None`)
- `do_single_rollout()` calls the hook after the `Trajectory` is fully assembled
- Add `TrajectoryHook` protocol (`async def(Trajectory) -> None`)
- 5 tests covering normal episodes, single-step episodes, class-based observers, and metrics/logs inspection

## Motivation

The existing logging infrastructure (logtree, rollout_summaries) cannot capture trajectories that terminate via adapter-level exits in `EnvFromMessageEnv`:

- **Parse errors** return a `StepResult` immediately without calling `MessageEnv.step()` — any logging attached to the message-level env never knows the episode happened.
- **Context overflow** calls `MessageEnv.step()` but overrides `episode_done=True` — the message-level env thinks the episode is still running.

Both paths produce valid `Transition` objects in the rollout loop, so the final `Trajectory` is complete. But logging attached at the `MessageEnv` layer misses these episodes entirely. In our training runs, this meant **0 out of 42 episodes** were captured because most terminated via parse_error.

`trajectory_hook` runs at the rollout level — after the `Trajectory` is assembled, regardless of termination path. This is the right layer for trajectory-level persistence (DB writes, JSONL logging, W&B tables, etc.).

## Changes

| File | Change |
|---|---|
| `tinker_cookbook/rl/types.py` | Add `trajectory_hook: TrajectoryHook \| None = None` to `Env` |
| `tinker_cookbook/rl/rollouts.py` | 2 lines: call the hook after trajectory construction |
| `tinker_cookbook/rl/trajectory_hook.py` | `TrajectoryHook` protocol definition |
| `tinker_cookbook/rl/trajectory_hook_test.py` | 5 tests with a simple `CountingEnv` example |

## Test plan

- [x] All 70 existing RL tests pass
- [x] 5 new tests covering: hook receives complete trajectory, no hook is fine, single-step episodes, class-based observers, metrics/logs inspection
- [x] `CountingEnv` in the test file serves as a usage example

### DIFF
--- a/tinker_cookbook/rl/rollouts.py
+++ b/tinker_cookbook/rl/rollouts.py
@@ -137,7 +137,10 @@ async def do_single_rollout(policy: TokenCompleter, env: Env) -> Trajectory:
         stop_condition = step_result.next_stop_condition
         if step_result.episode_done:
             break
-    return Trajectory(transitions=transitions, final_ob=ob)
+    trajectory = Trajectory(transitions=transitions, final_ob=ob)
+    if env.trajectory_hook is not None:
+        await env.trajectory_hook(trajectory)
+    return trajectory
 
 
 @logtree.scope_header_decorator("Group Rollout")
--- a/tinker_cookbook/rl/trajectory_hook_test.py
+++ b/tinker_cookbook/rl/trajectory_hook_test.py
@@ -1,0 +1,150 @@
+"""Tests for trajectory_hook on Env.
+
+The trajectory_hook is an optional callback on Env that do_single_rollout()
+invokes after the Trajectory is fully assembled. It runs for every episode
+termination — including parse_error and context_overflow cases that bypass
+MessageEnv.step() — making it the right place for trajectory-level logging,
+persistence, or analysis.
+
+Example use case: an RL training system that logs every trajectory to a
+database for post-hoc analysis. Without the hook, trajectories that end
+via parse errors or context overflow would be silently dropped from the
+log, because the inner MessageEnv never sees those termination events.
+"""
+
+import asyncio
+from dataclasses import dataclass, field
+
+import tinker
+
+from tinker_cookbook.completers import StopCondition, TokenCompleter, TokensWithLogprobs
+from tinker_cookbook.rl.rollouts import do_single_rollout
+from tinker_cookbook.rl.trajectory_hook import TrajectoryHook
+from tinker_cookbook.rl.types import Action, Env, StepResult, Trajectory
+
+# ---------------------------------------------------------------------------
+# A simple counting env with a trajectory_hook
+# ---------------------------------------------------------------------------
+
+
+@dataclass
+class CountingEnv(Env):
+    """Env that counts down from max_steps, then terminates.
+
+    Demonstrates trajectory_hook usage: the hook receives the complete
+    Trajectory after the rollout loop finishes, regardless of how the
+    episode ended.
+    """
+
+    max_steps: int = 3
+    trajectory_hook: TrajectoryHook | None = None
+
+    _step: int = field(default=0, init=False)
+
+    async def initial_observation(self) -> tuple[tinker.ModelInput, StopCondition]:
+        return tinker.ModelInput.from_ints([0]), ["<stop>"]
+
+    async def step(self, action: Action) -> StepResult:
+        self._step += 1
+        done = self._step >= self.max_steps
+        return StepResult(
+            reward=1.0 if done else 0.0,
+            episode_done=done,
+            next_observation=tinker.ModelInput.from_ints([self._step]),
+            next_stop_condition=["<stop>"],
+            metrics={"step": float(self._step)},
+            logs={"action_tokens": len(action)},
+        )
+
+
+class StubPolicy(TokenCompleter):
+    """Policy that always returns the same tokens."""
+
+    async def __call__(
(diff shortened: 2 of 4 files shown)


YujiaBao:
Thanks for the PR and the detailed write-up! The motivation around capturing adapter-level exits is worth investigating. After reviewing the implementation and tracing through the rollout pipeline, I have a couple of concerns.

**Picklability**

`Env` objects must be pickleable for distributed rollout execution via `ProcessPoolExecutor` (see the `_RolloutTask` path in `rollouts.py` and the picklability requirements documented on `EnvGroupBuilder`). Adding a callable `trajectory_hook` attribute to `Env` would break this when the hook is set — closures and most callables aren't pickleable. This would silently break any training run using `set_rollout_executor()`.

**The trajectories may not actually be lost**

I traced both adapter-level exit paths in `EnvFromMessageEnv.step()`:

- **Parse error** (line 89-96): Returns a `StepResult` with `episode_done=True`, `reward=failed_parse_reward`, and `metrics={"parse_error": 1.0}`. `MessageEnv.step()` is never called — this part of the PR description is accurate.
- **Context overflow** (line 102-114): `MessageEnv.step()` IS called first (line 98), then the adapter overrides `episode_done=True` and `reward=0.0`, adding `metrics["context_overflow"] = 1.0`.

In both cases, `EnvFromMessageEnv.step()` returns a valid `StepResult`. Back in `do_single_rollout` (line 127-139), this becomes a normal `Transition` appended to the trajectory. The resulting `Trajectory` is complete and flows into `TrajectoryGroup`.

From there, the existing infrastructure captures it:
- **Rollout summary JSONL** (`rollout_logging.py:77-106`): Iterates `trajectory_group.trajectories_G` and writes every trajectory with per-step metrics (including `parse_error` and `context_overflow` flags).
- **Logtree logging** (`rollouts.py:170-175`): Logs all trajectories via `_log_single_trajectory_details`.

So the "0 out of 42 episodes captured" issue shouldn't be coming from the rollout infrastructure itself. Could you share more about where the logging gap was? It's possible the issue is in analysis code filtering out short trajectories, or in custom logging attached inside the `MessageEnv` implementation that doesn't fire on early adapter exits.

tinker_cookbook/rl/rollouts.py
```diff
@@ -137,7 +137,10 @@ async def do_single_rollout(policy: TokenCompleter, env: Env) -> Trajectory:
         stop_condition = step_result.next_stop_condition
         if step_result.episode_done:
             break
-    return Trajectory(transitions=transitions, final_ob=ob)
+    trajectory = Trajectory(transitions=transitions, final_ob=ob)
+    if env.trajectory_hook is not None:
+        await env.trajectory_hook(trajectory)
+    return trajectory
```
YujiaBao, line 144: If the hook raises, the trajectory is never returned — the caller gets an exception instead of the `Trajectory`. For a hook intended for logging/persistence, this seems too fragile. A `try/except` with a warning log would be safer, so a buggy hook doesn't crash training.

tinker_cookbook/rl/types.py
```diff
@@ -62,6 +67,12 @@ class Env(ABC):
     Discard after running for one episode.
     """
 
+    trajectory_hook: TrajectoryHook | None = None
+    """Optional hook invoked with the completed Trajectory after the rollout
+    loop finishes.  Runs for every episode termination path — including
```
YujiaBao, line 72: Every existing `Env` subclass in the codebase uses a custom `__init__` (not dataclass). A class attribute on the ABC means users can't pass `trajectory_hook` as a constructor argument — they'd need to do `env.trajectory_hook = hook` after construction, which is awkward in `EnvGroupBuilder.make_envs()` where envs are typically created in a list comprehension.

tinker_cookbook/rl/trajectory_hook_test.py
```diff
@@ -0,0 +1,150 @@
...
+The trajectory_hook is an optional callback on Env that do_single_rollout()
+invokes after the Trajectory is fully assembled. It runs for every episode
+termination — including parse_error and context_overflow cases that bypass
+MessageEnv.step() — making it the right place for trajectory-level logging,
+persistence, or analysis.
+
+Example use case: an RL training system that logs every trajectory to a
+database for post-hoc analysis. Without the hook, trajectories that end
+via parse errors or context overflow would be silently dropped from the
+log, because the inner MessageEnv never sees those termination events.
```
YujiaBao, line 12: This claim isn't accurate — trajectories that end via parse errors or context overflow ARE captured in the rollout summary JSONL export (`rollout_logging.py:77-106`) and logtree logging (`rollouts.py:170-175`). Both iterate over all trajectories in `TrajectoryGroup`, which includes adapter-level exits. The per-step `metrics["parse_error"]` and `metrics["context_overflow"]` flags are preserved. See the main PR comment for the full trace.

puma314:
@YujiaBao I was previously only looking at the logtree logging and saw that it only captured some trajectories, not all. It is a very good point that the rollout summary json has all the information I might need (my use-case is creating a custom dashboard for rollouts based on top of all trajectories), so I will just use that. My mistake, and thanks for pointing this out.

PR #526, opened 2026-03-24
Add multimodal tool result support

puma314:
## Motivation

Right now, multimodal tools are not well-supported and if used in a `AgentToolMessageEnv` will raise an exception since `format_content_as_string` raises when a `ContentPart` is not text or thinking. This PR:

- Adds `multimodal_tool_result()` helper for tools that return text + images (e.g., browser screenshots)
- Adds `format_content_for_logging()` to safely log multimodal content with `[image]` placeholders
- Updates `AgentToolMessageEnv` tool result logging to use `format_content_for_logging` so multimodal results don't crash
- Adds tests in `agent_tool_message_env_test.py` and `renderers_test.py`

## Test plan
- [x] `pytest tinker_cookbook/tool_use/agent_tool_message_env_test.py` — 14 tests pass (6 existing + 8 new)
- [x] `pytest tinker_cookbook/renderers/renderers_test.py::TestFormatContentForLogging` — 8 tests pass
- [x] Full unit suite: 865 passed, 46 skipped, 0 failures

@YujiaBao I'm open-minded to passing a parameter into `format_content_as_string` like `allow_images` or adding this logic to it directly, instead of creating a new method. You have more context on whether the exception raising in `format_content_as_string` is intentional or not.

### DIFF
--- a/tinker_cookbook/renderers/renderers_test.py
+++ b/tinker_cookbook/renderers/renderers_test.py
@@ -1480,7 +1480,6 @@ def test_register_and_get_custom_tokenizer(cleanup_custom_tokenizer):
 
 class TestFormatContentAsStringImages:
     def test_image_pil_object(self) -> None:
-
         content: list[ContentPart] = [ImagePart(type="image", image=Image.new("RGB", (10, 10)))]
         assert format_content_as_string(content) == "<image>Image(10x10, RGB)</image>"
 
@@ -1489,7 +1488,6 @@ class TestFormatContentAsStringImages:
         assert format_content_as_string(content) == "<image>https://example.com/img.png</image>"
 
     def test_mixed_content(self) -> None:
-
         content = [
             TextPart(type="text", text="Before"),
             ImagePart(type="image", image=Image.new("RGB", (20, 15))),
@@ -1500,7 +1498,6 @@ class TestFormatContentAsStringImages:
         )
 
     def test_custom_separator(self) -> None:
-
         content = [
             TextPart(type="text", text="A"),
             ImagePart(type="image", image=Image.new("RGBA", (5, 5))),
--- a/tinker_cookbook/tool_use/agent_tool_message_env_test.py
+++ b/tinker_cookbook/tool_use/agent_tool_message_env_test.py
@@ -185,7 +185,6 @@ class TestSimpleToolResultMultimodal:
     """Unit tests for simple_tool_result() with list[ContentPart] content."""
 
     def test_with_text_and_images(self):
-
         img1 = Image.new("RGB", (10, 10), color="red")
         img2 = Image.new("RGB", (10, 10), color="blue")
         parts = [
@@ -216,7 +215,6 @@ class TestSimpleToolResultMultimodal:
         assert msg.get("name") == "t"
 
     def test_interleaved_order_preserved(self):
-
         img = Image.new("RGB", (10, 10))
         parts = [
             ImagePart(type="image", image=img),
@@ -233,7 +231,6 @@ class TestSimpleToolResultMultimodal:
         assert content[2] == {"type": "image", "image": "https://example.com/img.png"}
 
     def test_passthrough_fields(self):
-
         img = Image.new("RGB", (10, 10))
         result = simple_tool_result(
             [
@@ -254,7 +251,6 @@ class TestSimpleToolResultMultimodal:
         assert result.messages[0].get("name") == "screenshot"
 
     def test_defaults_with_list_content(self):
-
         img = Image.new("RGB", (10, 10))
         result = simple_tool_result(
             [TextPart(type="text", text="text"), ImagePart(type="image", image=img)]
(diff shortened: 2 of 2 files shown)


tinker_cookbook/tool_use/tools.py
```diff
@@ -112,6 +113,62 @@ def error_tool_result(
     )
 
 
+def multimodal_tool_result(
```
YujiaBao: Rather than a separate function, consider adding an optional `images` parameter to `simple_tool_result`. The fallback-to-string logic you already wrote is exactly right — it would just live inside `simple_tool_result` instead. This keeps image support native to the existing API without new surface area.

tinker_cookbook/renderers/base.py
```diff
@@ -689,6 +689,36 @@ def ensure_text(content: Content) -> str:
     raise RendererError(f"Expected text content, got multimodal content with {len(content)} parts")
 
 
+def format_content_for_logging(content: Content, separator: str = "\n") -> str:
```
YujiaBao: This is identical to `format_content_as_string` except for the image branch. I'd suggest fixing `format_content_as_string` directly in a follow-up PR instead of adding a parallel function:
```python
elif p["type"] == "image":
    parts.append("[image]")
else:
    parts.append(f"[{p['type']}]")
```

tinker_cookbook/tool_use/agent_tool_message_env.py
```diff
@@ -97,7 +102,9 @@ async def step(self, message: Message) -> MessageStepResult:
             tool_result_messages = await self._handle_tool_calls(tool_calls)
 
             for i, msg in enumerate(tool_result_messages):
-                logs[f"tool_result_{i}"] = get_text_content(msg)
+                # Use format_content_for_logging to handle multimodal tool results
```
YujiaBao: Should come in the follow-up PR alongside the `format_content_as_string` fix. No urgency — `get_text_content` won't crash on images, it just silently drops them.

puma314:
Ah, didn't realize you were open to changing `simple_tool_result`, this is indeed much cleaner!! I also fixed a few things like `simple_tool_result` takes in `images` as `list[Image.Image | str]`, as I realized the `ImagePart["image"]` type can have a str for URL cases.  

PTAL @YujiaBao

tinker_cookbook/tool_use/tools.py
```diff
@@ -27,16 +28,23 @@ def simple_tool_result(
     *,
     call_id: str = "",
     name: str = "",
+    images: list[Image.Image | str] | None = None,
     should_stop: bool = False,
```
YujiaBao, line 30: Rather than adding an `images` parameter, widen `content` to `str | list[ContentPart]` directly:

```python
def simple_tool_result(
    content: str | list[ContentPart],
    ...
) -> ToolResult:
```

The `images` param forces a text-first, images-after shape, but models support freely interleaved text + image parts and our VL renderers preserve ordering. `Message["content"]` is already `str | list[ContentPart]`, so this aligns with the existing type system and matches the pattern users already know from supervised data (e.g., the VLM classifier recipe builds `list[ContentPart]` with `ImagePart`/`TextPart` in any order).

With this change, tools.py shouldn't need to import `PIL`, `TextPart`, or `ImagePart` at all — callers construct the content parts.

Also worth a one-line docstring note: images in tool role messages are a cookbook extension — none of the standard APIs (OpenAI, Anthropic, vLLM, SGLang) document image content in tool/function messages. It works because our renderers handle all roles uniformly via `_preprocess_message_parts()`.

tinker_cookbook/renderers/base.py
```diff
@@ -783,8 +787,14 @@ def format_content_as_string(content: Content, separator: str = "\n") -> str:
             parts.append(f"<think>{p['thinking']}</think>")
         elif p["type"] == "text":
             parts.append(p["text"])
+        elif p["type"] == "image":
+            img = p["image"]
+            if isinstance(img, str):
+                parts.append(f"<image>{img}</image>")
```
YujiaBao, line 793: Adding the `ImagePart` branch is good, but the `else` should stay as `raise RendererError(...)` — not fall back to `[<type>]`. Every other function that exhausts `ContentPart` (`content_to_jsonable`, `to_openai_message`) raises on unknown types. Silent fallback masks bugs.

```python
        elif p["type"] == "image":
            img = p["image"]
            if isinstance(img, str):
                parts.append(f"<image>{img}</image>")
            else:
                parts.append(f"<image>Image({img.size[0]}x{img.size[1]}, {img.mode})</image>")
        else:
            raise RendererError(f"Unknown content part type: {p['type']}")
```

tinker_cookbook/tool_use/tools.py
```diff
@@ -14,11 +14,12 @@
     get_type_hints,
 )
 
+from PIL import Image
```
YujiaBao: With the `content` widening approach, this PIL import becomes unnecessary — callers construct `ContentPart` lists themselves. But if it stays for any reason, `tools.py` has `from __future__ import annotations` so type annotations are lazy; move `from PIL import Image` under `TYPE_CHECKING`.

puma314:
ok removed the method, PTAL @YujiaBao

YujiaBao (approved):

PR #527, opened 2026-03-24
Fix LoRA merging for Qwen3.5 dense models

yuxuan18:
## Problem

`build_hf_model()` crashes with `WeightsMergeError` on Qwen3.5 models due to two issues:

1. **Split vs fused QKV** — Tinker trains separate `in_proj_q/k/v` LoRA weights, but HF stores a single fused `in_proj_qkv` (Q‖K‖V, unequal dims: Q=K=2048, V=4096 for 4B).
2. **`unembed_tokens` remap** — Qwen3.5 has `vision_config`, so HF uses `model.language_model.*` keys. `lm_head` is weight-tied and absent from safetensors

## Changes (`_merge.py`)

- `MergeOp.slice_start` — new field for row-slice targeting within a fused weight.
- `plan_merge_ops` — intercepts `in_proj_q/k/v`, redirects to `in_proj_qkv` with offsets from sibling `lora_B` shapes; reorders vision remaps so prefix applies before `unembed_tokens → embed_tokens`.
- `apply_merge_op` — applies delta to `target[start:end]` when `slice_start` is set.
- `validate_merge_op_shapes` — accepts sliced ops (checks bounds, not full-tensor equality).

## Tests

New tests in `TestQwen35QkvFusion` covering offset correctness, value correctness, non-overlap, unequal dims, and vision prefix path. All tests pass.

## Verified on

`Qwen/Qwen3.5-4B` and `Qwen/Qwen3.5-27B`

### DIFF
--- a/tinker_cookbook/weights/_merge.py
+++ b/tinker_cookbook/weights/_merge.py
@@ -371,7 +371,9 @@ def plan_merge_ops(
                     k = adapter_weights[pfx + ".in_proj_k.lora_B.weight"].shape[0]
                     start = {"q": 0, "k": q, "v": q + k}[qkv_role]
                     ops.setdefault(fused_key, []).append(
-                        MergeOp(target_key=fused_key, lora_A=lora_A, lora_B=lora_B, slice_start=start)
+                        MergeOp(
+                            target_key=fused_key, lora_A=lora_A, lora_B=lora_B, slice_start=start
+                        )
                     )
                     continue
             _plan_non_expert_op(target_key, lora_A, lora_B, n, profile, model_state_keys, ops)
--- a/tinker_cookbook/weights/merge_test.py
+++ b/tinker_cookbook/weights/merge_test.py
@@ -790,8 +790,7 @@ class TestQwen35QkvFusion:
         # All three should land on in_proj_qkv, not on separate split keys
         assert "model.layers.0.linear_attn.in_proj_qkv.weight" in ops
         assert not any(
-            k.endswith((".in_proj_q.weight", ".in_proj_k.weight", ".in_proj_v.weight"))
-            for k in ops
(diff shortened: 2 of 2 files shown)


YujiaBao:
Thanks for identifying and fixing these two bugs — the split QKV fusion and tied-embeddings remap are real issues affecting all Qwen3.5 models, and your reproduction on Qwen3.5-4B and 27B was really helpful for confirming the problem.

After reviewing, we had some concerns about the design approach and decided to reimplement this following the existing profile-based architecture. The main issues were:

1. **QKV fusion logic inlined in the generic path**: The `_QKV` detection and offset computation were added directly to `plan_merge_ops`, which runs for every adapter weight on every model. The existing pattern (GPT-OSS, DeepSeek) captures model-specific behavior in `MergeProfile` + profile detectors, keeping the generic planner clean. We added a `_detect_qwen3_5_profile()` detector matching `model_type` in `("qwen3_5", "qwen3_5_moe")` and a `split_qkv_projections` field on `MergeProfile` to gate the logic.

2. **Helper extraction**: The QKV interception was extracted into `_plan_split_qkv_op()`, parallel to the existing `_plan_non_expert_op` and `_plan_expert_ops` helpers.

3. **Two concerns mixed in one diff**: The unembed_tokens fix is a generic vision-model bug, while the QKV fusion is Qwen3.5-specific. Separating them makes each easier to reason about.

We kept the core ideas from your PR — `slice_start` on `MergeOp`, the offset computation from sibling `lora_B` shapes, and the tied-embeddings detection via `lm_head.weight` presence — those were all correct. The changes are mostly about where the logic lives.

The replacement PR is #528. It also adds e2e tests through the full `build_hf_model()` shard pipeline with a tiny Qwen3.5-4B model, covering QKV fusion, slice isolation, and the tied-embeddings unembed path.

PR #570, opened 2026-03-31
Add InterleavedRLDatasetBuilder for multi-domain RL training

YujiaBao:
## Summary

Adds `InterleavedRLDatasetBuilder` to `tinker_cookbook.rl` for blending multiple RL datasets by weighted sampling. Enables multi-domain RL training where each batch contains `EnvGroupBuilder` instances drawn from multiple sources according to configured weights.

## API

```python
from tinker_cookbook.rl import InterleavedRLDatasetBuilder

builder = InterleavedRLDatasetBuilder(
    sources=[mcqa_builder, workbench_builder, structured_output_builder],
    weights=[0.55, 0.30, 0.15],
    groups_per_batch=8,
    total_batches=70,
)
```

- `sources` — list of `RLDatasetBuilder` instances to blend
- `weights` — relative weight per source (auto-normalized)
- `groups_per_batch` — number of `EnvGroupBuilder`s per output batch (required, no default)
- `total_batches` — fixed step count with cycling; `None` runs until smallest source exhausted
- `seed` — deterministic schedule for checkpoint recovery

## Design

- **Group-level blending**: each source is flattened into individual groups via arithmetic indexing (no eager materialization). Sources with different `groups_per_batch` compose correctly.
- **Shuffled cycling**: each pass through a source's groups uses a fresh shuffled permutation, so repeated epochs see problems in different order. All rollouts are on-policy — the schedule only determines which problems to serve, not pre-computed rollouts.
- **Deterministic schedule**: precomputed from seed for reproducibility and checkpoint-based recovery. Same `(seed, batch_index)` always produces the same groups.
- **Concurrent source building**: `asyncio.gather` builds all sources in parallel.
- **Per-domain metrics**: works with existing `logging_tags()` on `EnvGroupBuilder` — shows `env/mcqa/reward`, `env/workbench/reward` separately in wandb.
- **Ragged batch detection**: clear `IndexError` if a source returns variable-length batches.
- Only `InterleavedRLDatasetBuilder` is public; `_InterleavedRLDataset` is an implementation detail.
- Test datasets from individual sources are discarded — set up evaluation separately.

## Test plan

- [x] 20 unit tests covering: blending distribution, deterministic replay, shuffled cycling across epochs, all validation paths (empty sources, empty batches, ragged batches, negative index), multi-group sources (no data waste), different source batch sizes, builder `__call__` async path

### DIFF
--- a/tinker_cookbook/rl/interleaved_test.py
+++ b/tinker_cookbook/rl/interleaved_test.py
@@ -3,6 +3,7 @@
 import asyncio
 from collections import Counter
 from collections.abc import Sequence
+from typing import cast
 
 import pytest
 
@@ -277,7 +278,7 @@ class Test_InterleavedRLDataset:
         for i in range(20):
             batch = ds.get_batch(i)
             for builder in batch:
-                seen_indices.add(builder.idx)
+                seen_indices.add(cast(MockEnvGroupBuilder, builder).idx)
 
         # All 20 groups should be reachable (indices 0-19)
         assert len(seen_indices) == 20, (
@@ -299,7 +300,7 @@ class Test_InterleavedRLDataset:
         seen = set()
         for i in range(9):
             for builder in ds.get_batch(i):
-                seen.add(builder.idx)
+                seen.add(cast(MockEnvGroupBuilder, builder).idx)
 
         # Should see all 9 groups, not just indices 0, 3, 6 (first of each batch)
         assert seen == set(range(9))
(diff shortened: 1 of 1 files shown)


tinker_cookbook/rl/interleaved.py
```diff
@@ -0,0 +1,365 @@
...
+Provides :class:`InterleavedRLDatasetBuilder`, which blends multiple
+:class:`RLDatasetBuilder` sources into a single training dataset. Each
+training batch contains :class:`EnvGroupBuilder` instances drawn from
+different sources according to configured weights. Groups are never
+mixed across domains — each group comes from a single source,
+preserving GRPO's within-group advantage centering.
+
+Example::
+
+    builder = InterleavedRLDatasetBuilder(
```
hjc-puro, line 12: Is "Interleaved" a common term for dataset mixing? For me some naming like `MultiEnvRLDatasetBuilder` would be clearer but this might be from inexperience

YujiaBao: @hjc-puro  I was using the terminology from huggingface's `datasets`. We also added a similar feature for our SFT data earlier, so I want to be consistent on the naming. https://huggingface.co/docs/datasets/v4.8.4/process

PR #571, opened 2026-04-01
Filter empty EncodedTextChunks in RenderedMessage to fix SFT crash

YujiaBao:
## Summary
- Thinking-capable renderers (DeepSeek, Qwen3, Kimi, Nemotron) strip thinking content from historical assistant messages. When a message has only thinking content and no text, this produces `EncodedTextChunk(tokens=[])` which Tinker rejects with "Chunk N has empty tokens list".
- Moved the empty-chunk filter into `RenderedMessage.__post_init__` so all consumers (`build_supervised_example`, `build_generation_prompt`, and any future code) are protected at construction time. Removed the now-redundant ad-hoc filters from `build_generation_prompt`, `build_supervised_example`, and `KimiK2Renderer.build_supervised_example`.
- Added regression tests: parametrized test across all 6 thinking renderers with thinking-only assistant messages, unit tests for `RenderedMessage` filtering, and a DeepSeek-specific `render_message` test.

## Test plan
- [x] `test_rendered_message_filters_empty_encoded_text_chunks` — verifies `__post_init__` drops empty chunks
- [x] `test_rendered_message_preserves_non_empty_chunks` — verifies no over-filtering
- [x] `test_thinking_only_assistant_message_does_not_crash` — parametrized across Qwen3, Qwen3.5, DeepSeek, KimiK2, KimiK2.5, Nemotron3
- [x] `test_deepseek_render_message_thinking_only_produces_no_empty_chunks` — render_message level
- [x] Full renderer test suite: 381 passed, 71 skipped (all skips are pre-existing DeepSeek tokenizer bug)

### DIFF
--- a/tinker_cookbook/renderers/base.py
+++ b/tinker_cookbook/renderers/base.py
@@ -1076,6 +1076,20 @@ class RenderedMessage:
     stop_overlap: tinker.EncodedTextChunk | None = None
     """Tokens that overlap between stop sequence and next message's header."""
 
+    def __post_init__(self) -> None:
+        # Filter out empty EncodedTextChunks from output. These arise when renderers
+        # strip thinking content from historical assistant messages and the message has
+        # no remaining text. Tinker rejects empty token chunks with
+        # "Chunk N has empty tokens list", so we filter them at construction time to
+        # prevent any consumer from encountering them.
+        filtered = [
+            chunk
+            for chunk in self.output
+            if not isinstance(chunk, tinker.EncodedTextChunk) or chunk.tokens
+        ]
+        if len(filtered) != len(self.output):
+            object.__setattr__(self, "output", filtered)
+
 
 class TrainOnWhat(StrEnum):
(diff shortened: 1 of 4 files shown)


YujiaBao:
Closing this — the root cause is a broken DeepSeek V3.1 tokenizer configuration on HuggingFace, not a bug in our rendering code.

The DeepSeek tokenizer declares `byte_fallback: True` in its BPE config but has zero byte tokens (`<0x00>`–`<0xFF>`) in its vocabulary. The `tokenizers` library silently drops any characters that can't be represented (CJK, emoji, non-Latin scripts) instead of raising an error. ~6.6% of tulu3 rows contain CJK-only messages, which tokenize to `[]`, producing `EncodedTextChunk(tokens=[])` that Tinker rejects.

Other models (Qwen3, Llama3, Nemotron) work fine because their tokenizers include CJK characters in the BPE vocabulary. This is not related to thinking-block stripping.

PR #588, opened 2026-04-03
Generalize SandboxFactory to support non-Modal backends

ZitongYang:
- Change SandboxFactory signature from `(modal.Image, int)` to `(Path, int)` so any sandbox backend can be plugged in without depending on Modal types
  - Move `modal.Image` construction into `default_sandbox_factory`, making modal a deferred import
  - Simplify `load_harbor_tasks` to use flat directory layout (`<dataset>/<task_name>/`) instead of requiring the legacy `<dataset>/<uuid>/<task_name>/` nesting


danobi (approved):
lgtm!

PR #656, opened 2026-04-10
tutorials: add an optional text input for the Tinker API key

akshayka:
This PR adds a text input to each marimo tutorial notebook for inputting their API key, in case they did not set it in their environment before opening the notebook. If an API key was already present in the environment, then the notebook proceeds as before.

Additionally, this PR fixes an unrelated bug in `103_async_patterns.py`, which previously redefined a variable.

Finally, this PR adds "open in molab" links to the tutorials readme, making is easy for users to run the tutorials online should they wish.

image

P.S. I am the original creator of marimo. It was very cool to see you all use marimo for these tutorials. Thanks!

### DIFF
--- a/tutorials/101_hello_tinker.py
+++ b/tutorials/101_hello_tinker.py
@@ -62,8 +62,9 @@ def _(mo):
 @app.cell
 def _(mo):
     api_key = mo.ui.text(kind="password", label="Paste your Tinker API key")
-    api_key
+    api_key  # noqa: B018
     return (api_key,)
+
 
 @app.cell
 async def _(api_key, mo, tinker):
--- a/tutorials/101_hello_tinker.py
+++ b/tutorials/101_hello_tinker.py
@@ -60,7 +60,23 @@ def _(mo):
 
 
 @app.cell
-async def _(tinker):
+def _(mo):
+    api_key = mo.ui.text(kind="password", label="Paste your Tinker API key")
+    api_key
+    return (api_key,)
+
+@app.cell
+async def _(api_key, mo, tinker):
+    import os
+
+    mo.stop(
+        "TINKER_API_KEY" not in os.environ and not api_key.value,
+        "Paste your API key above",
+    )
+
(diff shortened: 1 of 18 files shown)


YujiaBao (changes requested):
Thanks for the PR — and for building marimo! The API key input is a great UX improvement for users running tutorials without a pre-set environment, and the molab links are a nice addition.

Two things to address:

**1. CI is failing — ruff B018 + formatting**

The bare `api_key` expression in each cell triggers ruff's B018 ("Found useless expression"). I know this is idiomatic marimo for displaying a widget, but our linter doesn't know that. Adding `# noqa: B018` to each of those lines should fix it. There are also some minor ruff formatting issues (missing blank lines between cells).

**2. Missing API key input in several tutorials**

Since this PR already adds molab links to all tutorials in the README, could you also add the API key input to the remaining notebooks that call the Tinker API? Without it, users opening these on molab will hit an error with no way to recover:

- `201_rendering.py` — calls `tinker.ServiceClient()`
- `303_sft_with_config.py` — calls `train.main()`
- `304_rl_with_config.py` — calls `rl_train.main()`
- `404_sequence_extension.py` — uses `tinker`
- `405_multi_agent.py` — uses `tinker`

(401, 402, 403 may be fine if they're purely local computation — happy to check if you're unsure. 503 might also need it depending on whether it calls the service.)

akshayka:
@YujiaBao, thanks for the review! I believe I've addressed your comments, let me know if I missed anything.

* I've updated the notebooks to satisfy precommit.
* I've added the API key elements to 303 and 304.
* I manually verified that the following notebooks work without an API key (if you still want me to add the input, let me know):

  * 201 does not use `ServiceClient`, that's only in the Markdown prose. 
  * 401
  * 402
  * 403
  * 404_sequence_extension (I fixed an unrelated bug.)
  * 405_multi_agent

YujiaBao (approved):

PR #660, opened 2026-04-12
Reduce tic-tac-toe default training to 40 steps

### LINKED ISSUE #130: suggested tweaks to tic-tac-toe example
The default settings in https://github.com/thinking-machines-lab/tinker-cookbook/blob/main/tinker_cookbook/recipes/multiplayer_rl/text_arena/train.py use `num_train_datapoints = 131072` with a batch size of 512, which amounts to 256 steps.  However, I found that you can get good results with many fewer steps (see figure), which is faster and cheaper. For example, running for 130 steps took me 6h (wall clock) and cost $40 (!).

I suggest changing `num_train_datapoints = 20_000`  which amounts to 39 steps (which should only take ~2h).
I also suggest showing the expected reward curve (as in example below), and explaining why a reward of 0 is good (since I assume it corresponds to always drawing, which is the expected return for player 0 under an optimal strategy).

[Of course, if we just use 39 steps, we won't see the instabilities that occur later in training, which seem to be due to going off policy.]

It would also be great if you could tweak the demo to support playing against a random opponent (instead of self play), either at training or testing or both. This may simplify the code, and also shows that the learned policy is general, and doesn't always deterministically execute the same sequence of actions from the initial board state. It would be nice to also allow the user to interactively play against the learned strategy.  (My [naive attempt](https://discord.com/channels/1435768626681741364/1435768628250415258/1442238316429705247) to interact with this TwoAgentEnv failed, due to async issues.) 

Image

dylantirandaz:
Closes #130.

Lower `num_train_datapoints` from 40960 (80 steps) to 20480 (40 steps) in the `multiplayer_rl/text_arena` recipe. The model reaches strong play by step 40, well before the self-play collapse at step 80+, so the default run is roughly half the wall-clock time and API cost with no loss in demonstrated performance.

README updates:
- Step table and training-duration estimate reflect the new default
- Collapse-dynamics section notes the 40-step default avoids the instability and shows how to opt into a longer run

### DIFF
--- a/tinker_cookbook/recipes/multiplayer_rl/text_arena/README.md
+++ b/tinker_cookbook/recipes/multiplayer_rl/text_arena/README.md
@@ -15,7 +15,7 @@ We show how to coordinate the steps of two *Environment* objects such that both 
 python -m tinker_cookbook.recipes.multiplayer_rl.text_arena.train
 ```
 
-With the default settings (batch size 512, ~40k datapoints = 80 steps), training takes roughly 1-2 hours.
+With the default settings (batch size 512, ~20k datapoints = 40 steps), training takes roughly 1 hour.
 
 ### Expected results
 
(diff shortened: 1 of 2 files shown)


YujiaBao:
Thanks for the PR, @dylantirandaz! I appreciate the motivation to reduce wall-clock time and cost.

I'd prefer to keep the default at 80 steps though. I've observed that training dynamics vary across different base models — while Qwen3-4B may plateau by step 40, other models may need more steps. Keeping 80 steps also serves an educational purpose: it puts users closer to the collapse region, which is one of the key takeaways of this recipe. The step 60 row in the results table is useful reference for that.

Let's keep the 80-step default and the existing results table as-is.

dylantirandaz:
Okay sounds good no worries!

Issue #268, opened 2025-12-30
tool_use/search example is misaligned with tinker's key idea


josancamon19:
```
If you launch the chroma service locally, you generally need 160+ GB RAM to load the vector index in memory for good performance.
```

I think every recipe should be runnable locally, at minimum this would require to ssh into a machine with this amount of ram available to host it, or download the file and upload it to chroma, which is probably best (probably pricey and slow), but still most laptops don't have 200GB to spare. 

Solutions
- easiest: to host it, and rate limit?
- a script to either upload to chroma easily or to host it in a machine with that ram available

I just think the most exciting premise is to not worry about infra, and one of the (most interesting) examples should not make me worry about it

joschu:
good point, agreed

joschu:
We're working on an easier-to-use recipe.

IgorGanapolsky:
## Load pattern +1

When sampling slows under fan-out we separate train vs eval traffic, cap concurrent `SamplingClient`s, prefer fewer fatter sessions, and use a **watermarked eval store** so retries do not recompute finished `example_id`s. Pure eval sometimes moves to local vLLM when provider latency dominates wall-clock.

Documented session caps and expected throughput under load would save the community a lot of guesswork.

_Issue #268: tool_use/search example is misaligned with tinker's key idea_

---
*Independent of Thinking Machines Lab. Related open work: [ThumbGate.app](https://thumbgate.app/) (Hermes agent control plane) and fail-closed GTM/cash validators in [mac-yolo-safeguards](https://github.com/IgorGanapolsky/mac-yolo-safeguards) (`tools/tinker-brain/`, `tools/governed-data-mesh.js`). No affiliation claimed with TML or model providers.*

PR #662, opened 2026-04-12
Add lightweight local BM25 variant to Search-R1 recipe

dylantirandaz:
Addresses #268.

Provides a laptop-friendly alternative to the existing Chroma + Gemini pipeline, which requires 160+ GB RAM to host the full Wikipedia vector index. The training loop, system prompt, reward function, and dataset are identical — only the retrieval backend changes.

New files under `recipes/search_tool/`:
- `local_tools.py` — `BM25Index` (pure numpy, ~100 lines) and `LocalSearchTool` exposing the same `@tool`-decorated `search` method as `ChromaTool`. Default corpus is `rag-datasets/rag-mini-wikipedia` (~3k passages, <10 MB), loaded once via the HuggingFace `datasets` library.
- `search_env_local.py` — `LocalSearchR1DatasetBuilder` / `LocalSearchEnvGroupBuilder` paralleling the Chroma builder; subsamples the SearchR1 train split to 2048 examples by default since the small corpus does not cover rare entities.
- `train_local.py` — CLI entry point with a simpler config (no Chroma host/port, no Gemini credentials, no vector-search optional deps).
- `local_tools_test.py` — colocated unit tests for tokenization and BM25 ranking against a toy corpus.

The existing Chroma pipeline is untouched — users replicating paper-level numbers still run `python -m tinker_cookbook.recipes.search_tool.train`. The README gains a "Lightweight alternative" section with a trade-offs table.

Trade-offs are explicit: retrieval quality is lower (BM25 over 3k docs vs dense embeddings over full Wikipedia), so this is suited for iterating on the training loop, not benchmarking.

Note: maintainer comment on #268 mentioned an easier-to-use recipe was in progress — happy to close this in favor of that or rework it to match the intended design.


YujiaBao (changes requested):
Thanks for the PR, @dylantirandaz! The BM25 tool itself is clean and the laptop-friendly angle is a good one.

The main feedback is on structure — the new files duplicate a lot of the existing recipe code. `LocalSearchEnvGroupBuilder`, `LocalSearchRLDataset`, `_initial_messages`, and most of `train_local.py` are near-identical copies of their Chroma counterparts. We'd prefer to follow the pattern used in the SDFT recipe (`recipes/sdft/train.py`), where a single `train.py` dispatches to different backends via a config option.

Concretely:

1. **Keep `local_tools.py`** — the BM25 implementation is genuinely new, looks good.
2. **Widen `SearchEnvGroupBuilder`** in the existing `search_env.py` — rename `chroma_tool` to `search_tool` and accept `ChromaTool | LocalSearchTool`. Same for `_initial_messages`. Both tools expose the same `@tool`-decorated `search` method, so this is a type annotation change. This eliminates `LocalSearchEnvGroupBuilder` and `LocalSearchRLDataset` entirely.
3. **Add `LocalSearchR1DatasetBuilder`** as a new builder class in `search_env.py` — the builder configs are genuinely different (chroma params vs local params), so a separate class is right (same pattern as `SciKnowEvalSDFTBuilder` vs `ToolAlpacaSDFTBuilder` in `recipes/sdft/datasets.py`).
4. **Fold `train_local.py` into `train.py`** — add a `backend: str = "chroma"` option to `CLIConfig` and dispatch in `cli_main()`. Users would run: `python -m tinker_cookbook.recipes.search_tool.train backend=local`

This should eliminate `search_env_local.py` and `train_local.py` while keeping all the new BM25 functionality.

(Review assisted by Claude)

dylantirandaz:
Thanks for the review, all addressed in the latest push (5a5631a). 

- `search_env_local.py` and `train_local.py` are gone.
- `SearchEnvGroupBuilder` now takes `search_tool: ChromaTool | LocalSearchTool`. The `chroma_tool` → `search_tool` rename is internal to the class; `SearchR1DatasetBuilder` passes it through.
- `LocalSearchR1DatasetBuilder` lives next to `SearchR1DatasetBuilder` in `search_env.py`, matching the `SciKnowEvalSDFTBuilder` / `ToolAlpacaSDFTBuilder` pattern.
- `train.py` adds `backend: Literal["chroma", "local"] = "chroma"` and dispatches via `_build_chroma_dataset_builder` / `_build_local_dataset_builder` helpers in `cli_main`. The run name now includes the backend name so logs are easy to distinguish.
- Usage: `python -m tinker_cookbook.recipes.search_tool.train backend=local`.

`local_tools.py` and `local_tools_test.py` are unchanged.

PR #663, opened 2026-04-12
Serialize conversation history in AgentToolMessageEnv 

tyler-griggs:
## Summary
- On episode end in `AgentToolMessageEnv.step()`, serializes `self.history` into `logs["conversation_history"]` as JSON
- Uses existing `message_to_jsonable()` to handle ToolCall pydantic objects and PIL images

### DIFF
--- a/tinker_cookbook/tool_use/agent_tool_message_env.py
+++ b/tinker_cookbook/tool_use/agent_tool_message_env.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import json
 from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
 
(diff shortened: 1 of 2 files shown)


YujiaBao:
Hey @tyler-griggs — the per-step `logs` in the rollout summaries JSONL already capture `assistant_content`, `tool_call_N`, and `tool_result_N` each turn (see `serialize_rollout_summaries` in `rollout_logging.py:113-145`). Adding `conversation_history` to the env logs would duplicate that data per trajectory.

If the missing piece is initial messages, I think the right place is to add them to the existing trajectory record in `serialize_rollout_summaries` rather than serializing the full history in the env. That way everything lives in one place without duplication.

(Reviewed with Claude)

PR #665, opened 2026-04-15
Add image support for preference learning

Jah-yee:
This PR adds image support for preference learning.

Changes:
- Removed the assertion that blocked images in preference learning
- Updated the truncation logic to work with both text chunks and image chunks  
- Image chunks are now properly handled during truncation

### DIFF
--- a/tinker_cookbook/preference/types.py
+++ b/tinker_cookbook/preference/types.py
@@ -175,19 +175,46 @@ class ComparisonRendererFromChatRenderer(ComparisonRenderer):
         convo = self._comparison_to_convo(labeled_comparison.comparison)
         convo_with_pref = convo + [{"role": "assistant", "content": labeled_comparison.label}]
         model_input, weights = self.convo_renderer.build_supervised_example(convo_with_pref)
-        # TODO: support images in preference learning
-        assert all(isinstance(c, tinker.types.EncodedTextChunk) for c in model_input.chunks), (
-            "Preference learning currently only supports text-only content."
-        )
-        # Truncate at the first weight==1 position + 1
-        tokens = model_input.to_ints()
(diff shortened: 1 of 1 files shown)


YujiaBao (changes requested):
Thanks for working on this! The image support for preference learning looks like the right direction, but this PR needs some cleanup before review.

1. **Rebase needed** — the Nemotron-3 and Qwen3.5 renderer code (commit 2) has already been merged to `main` (`0dd21a4`). Please rebase onto current `main` — this should drop ~1450 lines from the diff and leave just the preference learning change.

2. **PR description vs contents** — after rebasing, the `math.ceil` fix (commit `f67537a6`) will still be here. That same commit also appears in PR #666. Please split it into its own PR or drop it from one of the two.

3. **Image chunk truncation behavior** — the chunk-walking logic looks correct, but silently dropping an image chunk that overlaps the truncation point could be surprising. Worth adding a log warning when this happens, or documenting the intended behavior.

_Review assisted by Claude._

Jah-yee:
Good day,

Thank you for the thorough review. I have addressed all three points:

1. **Rebase**: The branch has been rebased onto current main. The Nemotron-3 and Qwen3.5 renderer code is no longer in this PR (those files are now identical to main). Only the preference learning changes remain.

2. **math.ceil duplicate**: The sl_loop.py change has been removed from this PR. Please refer to PR #666 for that fix.

3. **Image chunk truncation warning**: Added `logger.warning(f"Image chunk dropped due to truncation at index {first_weight_one_index}")` in types.py when an image chunk is excluded due to truncation overlap.

The diff now only contains the image support for preference learning.

Thank you for your attention to these changes.

Warmly, RoomWithOutRoof

PR #664, opened 2026-04-15
feat(renderers): add Nemotron-Cascade-2-30B-A3B support

Jah-yee:
Good day,

This PR adds renderer support for the nvidia/Nemotron-Cascade-2-30B-A3B model as requested in issue #511.

## Changes
- Add `nemotron_cascade_2` renderer (with thinking enabled, using Qwen3.5 format)
- Add `nemotron_cascade_2_disable_thinking` renderer (without thinking)
- Also includes Qwen3.5 and Nemotron-3 renderers from upstream

## Technical Details
Nemotron-Cascade-2-30B-A3B uses the same chat template as Qwen3.5 with `<|im_start|>` tokens and supports the `enable_thinking` parameter. The implementation reuses the existing Qwen3.5 renderer classes.

Example usage:
```python
from tinker_cookbook import renderers
renderer = renderers.get_renderer("nemotron_cascade_2", tokenizer)
```

Thank you for your work on this excellent library!

Warmly,
RoomWithRoof

PR #671, opened 2026-04-16
fix(sl_loop): use math.ceil for train batch count

Jah-yee:
Good day,

This PR changes the train batch count calculation in sl_loop.py from floor division to math.ceil.

## Change
- Import math module
- Change n_train_batches from len(train_dataset) // config.batch_size to math.ceil(len(train_dataset) / config.batch_size)

This ensures the total number of batches is correctly counted even when the dataset size is not evenly divisible by the batch size.

Note: This is submitted as a standalone PR as suggested in the review of PR #664, where the same math.ceil fix was found to be duplicated across PRs #665 and #666.

Thank you for your attention...

Warmly, RoomWithOutRoof

### DIFF
--- a/tinker_cookbook/recipes/sl_loop.py
+++ b/tinker_cookbook/recipes/sl_loop.py
@@ -4,6 +4,7 @@ Uses existing modules but with a simple, flat training loop.
 """
 
 import logging
+import math
 import time
 
 import chz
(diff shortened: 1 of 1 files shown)


YujiaBao:
Thanks for the PR! The floor division here is intentional — the comment just above explains it mimics PyTorch's `drop_last=True` to avoid the gradient-magnitude spike from a partial final batch (added in #415). Switching to `math.ceil` would re-introduce that issue and also make the `"Dropping last N examples"` log just below inaccurate. Closing, but thanks for taking the time to send this in!

PR #672, opened 2026-04-19
Add GSPO recipe

lokashrinav:
## Add GSPO recipe

GSPO ([arXiv:2507.18071](https://arxiv.org/abs/2507.18071)) implemented as a cookbook recipe using `forward_backward_custom`.

The Qwen team built GSPO to train Qwen3. GRPO computes importance ratios **per token**; GSPO computes **one ratio per sequence**—the **geometric mean** of token ratios. That tends to behave better on **long sequences** (less of the “ratio noise stacks with length” failure mode). It also avoids **routing replay** issues in **MoE** training, because sequence-level likelihood stays stable even when expert routing changes between gradient steps.

Built on `rl_loop.py`. Two changes:

- **`loss.py`**: GSPO loss, passed to `forward_backward_custom`. Old logprobs and advantages are captured in a **closure** because `loss_fn_inputs` can’t carry them directly.
- **`train.py`**: Advantages are normalized by **within-group standard deviation** (not only mean-centered). Uses `forward_backward_custom` instead of the built-in `"importance_sampling"` loss.

Tested on GSM8K.


YujiaBao (changes requested):
Thanks for putting this together! Walked through the math against the paper and compared structure against the other recipes in `recipes/`. The algorithm looks right — a few recipe-level things to address before merging.

**Math checks out.** Eq. 5/6/7 all map correctly onto `loss.py`, and the `[ob_len:]` slice correctly isolates response tokens (worth knowing: `forward_backward_custom` returns logprobs at full `model_input.length`, so masking is the caller's job — DPO does this with a `weights` tensor instead, but the slice is equivalent).

**The bigger asks are recipe conventions:**

- **Add a `README.md`.** Every other recipe has one (`sdft/README.md` is a good template). For a research recipe this really matters — paper link, algorithm summary, run command, param table, expected reward curve.
- **Add a unit test for `make_gspo_loss`.** The closure design is exactly where a silent regression could hide, and the loss is pure torch so it tests without an API key. `sdft/sdft_test.py` is the pattern to copy.
- **Timestamp the log path.** Right now runs overwrite each other's `metrics.jsonl`. See `sdft/train.py:104-113`.
- **Use `checkpoint_utils.resolve_renderer_name_from_checkpoint_or_default(...)`** instead of `model_info.get_recommended_renderer_name(...)` — otherwise resume-from-checkpoint breaks when model names don't match.
- **Expose `max_steps`, `wandb_project`, `wandb_name` in `Config`** — every other recipe does, and `max_steps` makes smoke-testing much easier.

Minor nits: per-datum `.item()` in `loss.py:71-72` forces D GPU→CPU syncs; `reward/mean` vs `rl_loop.py`'s `reward/total` makes overlays awkward; `if fwd_bwd_result.metrics:` on `train.py:231` is dead.

_(Review assisted by Claude Code.)_

lokashrinav:
Addressed the GSPO review feedback in 0f0051e:

- Added a recipe README with algorithm summary, run commands, parameter table, and expected metrics.
- Added pure unit tests for make_gspo_loss.
- Added timestamped log paths plus max_steps, wandb_project, and wandb_name config fields.
- Switched renderer resolution to checkpoint-aware checkpoint_utils handling and added checkpoint load/resume metadata checks.
- Cleaned up the GSPO loss metrics to avoid per-datum .item() syncs and aligned reward logging with rl_loop.

CI is green on the updated PR.

PR #673, opened 2026-04-20
Add kimi-cli file tools to Harbor environment

jeffrey-tml:
## Summary
- Add Glob, Grep, ReadFile, WriteFile, and StrReplaceFile tools to Harbor sandbox environments, mirroring kimi-cli's tool interfaces exactly
- Tools execute inside the sandbox via `SandboxInterface` (`run_command`, `read_file`, `write_file`)
- Grep is backed by ripgrep — the `rg` binary is resolved on the host and uploaded into the sandbox at setup time (no internet needed in sandbox)
- Gated behind `enable_file_tools` flag so existing bash-only configs are unaffected

## Test plan
- [ ] Unit tests with mock/local sandbox for each tool (Glob, Grep, ReadFile, WriteFile, StrReplaceFile)
- [ ] Verify tool schemas via `.to_spec()` match kimi-cli parameter names and defaults
- [ ] Integration test: run harbor_onpodi with `enable_file_tools=True` and verify tools appear in conversation prefix
- [ ] Verify `rg` binary upload works with Fishbowl sandboxes

### DIFF
--- a/tinker_cookbook/recipes/harbor_rl/eval.py
+++ b/tinker_cookbook/recipes/harbor_rl/eval.py
@@ -97,7 +97,7 @@ async def evaluate_task(
         env = build_agent_tool_env(
             renderer=renderer,
             tools=[bash_tool.bash],
-            initial_messages=_initial_messages(task, renderer, bash_tool),
+            initial_messages=_initial_messages(task, renderer, [bash_tool.bash]),
             reward_fn=reward_fn,
             max_turns=config.max_turns,
         )
--- a/tinker_cookbook/recipes/harbor_rl/harbor_env.py
+++ b/tinker_cookbook/recipes/harbor_rl/harbor_env.py
@@ -12,7 +12,15 @@ from typing import Any
 import chz
 
 from tinker_cookbook import model_info, tokenizer_utils
-from tinker_cookbook.recipes.harbor_rl.harbor_tools import HarborBashTool, HarborReward
+from tinker_cookbook.recipes.harbor_rl.harbor_tools import (
+    HarborBashTool,
+    HarborGlobTool,
+    HarborGrepTool,
+    HarborReadFileTool,
+    HarborReward,
+    HarborStrReplaceFileTool,
+    HarborWriteFileTool,
+)
 from tinker_cookbook.renderers import get_renderer
 from tinker_cookbook.renderers.base import Message, Renderer
 from tinker_cookbook.rl.types import Env, EnvGroupBuilder, RLDataset, RLDatasetBuilder
@@ -79,13 +87,14 @@ def load_harbor_tasks(dataset: str) -> list[HarborTask]:
 def _initial_messages(
     task: HarborTask,
     renderer: Renderer,
-    bash_tool: HarborBashTool,
+    tools: list,
+    system_prompt: str = HARBOR_SYSTEM_PROMPT,
 ) -> list[Message]:
     """Build initial messages with tool schemas and task instruction."""
-    tool_schemas = [bash_tool.bash.to_spec()]
+    tool_schemas = [t.to_spec() for t in tools]
     prefix = renderer.create_conversation_prefix_with_tools(
         tools=tool_schemas,
-        system_prompt=HARBOR_SYSTEM_PROMPT,
+        system_prompt=system_prompt,
     )
     return prefix + [{"role": "user", "content": task.instruction}]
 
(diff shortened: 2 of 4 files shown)


jeffrey-tml:
@codex review

YujiaBao (changes requested):
Thanks for this — the kimi-cli interface parity is nicely done and the feature-flag gating is clean. A few correctness issues and convention gaps I'd like addressed before merging. Full analysis below.

_Review assisted by Claude Code._

## Correctness

1. **Grep path-prefix stripping can corrupt match content** (`harbor_tools.py:467-469`)

   ```python
   if path not in (".", "/"):
       prefix = path.rstrip("/") + "/"
       output = output.replace(prefix, "")
   ```

   `str.replace` is global — any occurrence of the path inside a matched line's text (e.g. an import path, URL, log line) gets stripped. Please do this per-line and only at the start of the line (`line.removeprefix(prefix)`).

2. **StrReplaceFile replacement count is wrong for sequential edits** (`harbor_tools.py:869-875`)

   The tally iterates over `original_content` again, but edits are applied sequentially to the mutated string. Edits can double-count or miss matches that are created/destroyed by earlier edits. Either count during application or drop the count from the message.

3. **No validation that `rg` is shipped when file tools are enabled** (`harbor_env.py:178`)

   If `enable_file_tools=True` but `rg_host_path=None`, every Grep call silently fails because `/usr/local/bin/rg` doesn't exist in the sandbox. Please either raise in `__init__`/`_make_env_group_builders`, or skip Grep from the tool list when `rg_host_path` is absent.

4. **`HARBOR_SYSTEM_PROMPT` still advertises only the bash tool**

   When `enable_file_tools=True`, the model sees "You have access to a bash tool" and no mention of the 5 new tools. The `system_prompt` parameter was added to `_initial_messages` but never threaded through. Either update the default prompt to reflect the optional tools, or plumb a real override from `HarborEnvGroupBuilder`.

## Convention / CLAUDE.md

5. **`tools: list` is untyped** (`harbor_env.py:90, 147`) — please use `list[Tool]` to match the "explicit typing" convention.

6. **`# type: ignore[arg-type]` × 5 in `_build_tools`** — caused by `sandbox: SandboxInterface | None = None`, but every caller passes a real sandbox and none of the tool classes accept `None`. Drop the `| None = None` and the ignores disappear. CLAUDE.md: "avoid `type: ignore`".

7. **Unused parameter** — `_initial_messages(..., system_prompt=...)` has no non-default caller. Either wire it through (see #4) or drop it.

8. **Missing tests** — the existing `harbor_tools_test.py` already has a `FakeSandbox` fixture covering bash/reward/pickle. Please add colocated tests for each new tool: a `.to_spec()` shape check and at least one `.run(ToolInput(...))` happy-path + one error-path. The PR description's test plan lists these as unchecked; the repo convention is colocated `*_test.py`.

9. **No docs/example wiring** — no mention of `enable_file_tools` or `rg_host_path` in `README.md`, `train.py`, or `eval.py`. A short example config would help users adopt this.

## Smaller points (non-blocking)

- `harbor_tools.py:678` — `len(all_entries) > MAX_READ_LINES` is unreachable since the `line_offset < -MAX_READ_LINES` check at line 575 already bounds `tail_count`. Dead branch.
- `harbor_tools.py:455-457` — the `exit_code not in (0, 1)` check is reachable only when `exit_code != 0` (the `==1` case already returned). Simplify to `!= 0`.
- `harbor_tools.py:583` — `ReadFile` caps sandbox reads at `MAX_READ_BYTES * 2`. For files >~200KB, `total_lines` and the "End of file reached" message are based on truncated content. Either document the cap or read without cap and bound downstream.
- `harbor_tools.py:151-165` — Glob shells out to `python3 -c "..."`. Introduces a hidden dependency on `python3` in the sandbox image and is brittle to quote. Consider `find` or sandbox-native listing.
- `harbor_env.py:182-188` — the `rg` binary (~5 MB) is `write_file`d once per sandbox (group_size × tasks). The `read_bytes` is cached but the upload isn't batched. Worth a comment or a follow-up.
- `harbor_tools.py:366` — `type` parameter shadows the builtin (matches kimi-cli, so likely intentional — flagging for awareness).
- `harbor_tools.py:304` — `output_mode: str` could be `Literal["content", "files_with_matches", "count_matches"]` for validation.
- `harbor_tools.py:559-566` — the `ReadFile` docstring body is un-indented, unlike every other tool's docstring in the file. Affects how it renders in the tool schema.
- PR description mentions "Fishbowl sandboxes" — no such backend exists in this repo (Modal + SandboxFusion). Typo?

## Must-fix before merge
1, 2, 3, 4, 6, 8.

jeffrey-tml:
@codex review

jeffrey-tml:
@codex review

jeffrey-tml:
@codex review

jeffrey-tml:
@codex review

jeffrey-tml:
@codex review

jeffrey-tml:
@codex review

jeffrey-tml:
@codex review

jeffrey-tml:
@codex review

PR #677, opened 2026-04-21
sdft: reverse KL fix

likenneth:
Adds a `reverse` knob to SDFT (default False) that, when True with topk>0, trains against an analytical reverse-KL objective rather than the legacy single-sample importance-sampling approximation.

With p = student, q = teacher, 𝒯 = usable completion positions, and S_t = teacher's top-K at position t (both distributions renormalized over S_t):

    L_rev = Σ_{t∈𝒯} Σ_{k∈S_t} p_renorm(x_k|t) · sg[log p_renorm − log q_renorm]

Implemented via `forward_backward_custom_async` (renormalized teacher probs shipped in `weights` with 0 as the mask sentinel, matching the forward-KL CE encoding). Gradient flows only through the outer `p_renorm` weight — the bracketed advantage is detached.

Also marks the topk=0 branch as deprecated; it's superseded by `topk>0, reverse=True` and kept only for backward compat.

Verified on Qwen/Qwen3.5-35B-A3B, LoRA 64, lr=5e-4, 30 steps, ToolAlpaca:
- SDFT topk=0 (old single-sample IS): 58.76%
- SDFT reverse=True (new):             63.92%  (+5.2pp)
- Base model (published):              61.86%
- SDFT forward KL (published, full):   65.98%

### DIFF
--- a/tinker_cookbook/distillation/sdft.py
+++ b/tinker_cookbook/distillation/sdft.py
@@ -627,7 +627,8 @@ def reverse_kl_custom_loss(
     the per-slot validity mask via ``weights > 0`` and reconstruct
     ``log q_renorm = log(weights)`` for the KL computation.
     """
-    total_loss = torch.tensor(0.0)
+    device = logprobs_list[0].device if logprobs_list else torch.device("cpu")
+    total_loss = torch.zeros((), device=device)
     sum_kl = 0.0
     sum_student_entropy = 0.0
     sum_positions = 0.0
@@ -648,6 +649,7 @@ def reverse_kl_custom_loss(
         neg_inf = torch.full_like(student_logp_NK, float("-inf"))
         masked_logp = torch.where(slot_mask_NK, student_logp_NK, neg_inf)
         log_p_renorm = torch.log_softmax(masked_logp, dim=-1)
+        # Invalid slots land at 0 here (→ p_renorm=1), but are zeroed below via slot_mask_NK.
         log_p_renorm = torch.nan_to_num(log_p_renorm, nan=0.0, neginf=0.0)
         p_renorm = log_p_renorm.exp()
 
@@ -687,9 +689,9 @@ async def _train_step_reverse_kl(
 
     Mirrors :func:`tinker_cookbook.rl.train.train_step` but uses
     ``forward_backward_custom_async`` with :func:`reverse_kl_custom_loss`.
-    The datums produced by :func:`build_reverse_kl_datums` carry
-    ``position_mask`` / ``k_valid`` / ``teacher_logprobs`` instead of the
-    standard ``mask`` key, so no mask stripping is needed.
+    The datums produced by :func:`build_reverse_kl_datums` carry only
+    ``target_tokens`` and ``weights`` (with ``0`` in ``weights`` acting as
+    the mask sentinel), so no ``mask`` stripping is needed.
     """
     batches = split_list(data_D, min(num_substeps, len(data_D)))
     if not batches:
@@ -809,6 +811,12 @@ async def main(
             batches. Use :class:`~tinker_cookbook.recipes.sdft.datasets.SDFTDataset`.
         test_dataset: Optional test dataset for periodic evaluation.
     """
+    if cfg.reverse and cfg.topk == 0:
+        raise ValueError(
+            "reverse=True requires topk>0: the analytical reverse KL runs over "
+            "the teacher's top-K set, which doesn't exist when topk=0."
+        )
+
     ml_logger = ml_log.setup_logging(
         log_dir=cfg.log_path,
         wandb_project=cfg.wandb_project,
--- a/tinker_cookbook/distillation/sdft.py
+++ b/tinker_cookbook/distillation/sdft.py
@@ -70,6 +70,7 @@ from tinker_cookbook.rl.types import (
     TrajectoryGroup,
 )
 from tinker_cookbook.utils import ml_log, trace
+from tinker_cookbook.utils.misc_utils import split_list
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


tinker_cookbook/distillation/sdft.py
```diff
@@ -464,6 +465,269 @@ async def build_topk_distillation_datums(
...
+    ``[log p_renorm - log q_renorm]`` bracket, and take the mass-weighted
+    sum. Gradient flows only through the outer ``p_renorm`` weight, matching
+    Tinker's ``importance_sampling`` convention.
+
+    The per-slot ``weights`` carry renormalized teacher probability (``q_renorm``)
+    at valid slots and ``0`` at padding slots or masked positions. We recover
+    the per-slot validity mask via ``weights > 0`` and reconstruct
+    ``log q_renorm = log(weights)`` for the KL computation.
+    """
+    total_loss = torch.tensor(0.0)
```
YujiaBao: nit: `torch.tensor(0.0)` creates a CPU tensor. If `logprobs_list[i]` ever lands on a non-CPU device, `total_loss + loss_d` below triggers a host↔device copy each iteration. Safer to initialize from the incoming tensor's device:

```python
device = logprobs_list[0].device if logprobs_list else torch.device("cpu")
total_loss = torch.zeros((), device=device)
```

Likely latent today (Tinker custom-loss tensors are CPU), but worth pinning.

_Review assisted with Claude Code._

tinker_cookbook/distillation/sdft.py
```diff
@@ -464,6 +465,269 @@ async def build_topk_distillation_datums(
...
+        # Reconstruct log q_renorm safely; clamp away from 0 for the log op.
+        safe_weights = weights_NK.clamp(min=1e-30)
+        teacher_log_renorm_NK = torch.where(
+            slot_mask_NK, torch.log(safe_weights), torch.zeros_like(weights_NK)
+        )
+
+        neg_inf = torch.full_like(student_logp_NK, float("-inf"))
+        masked_logp = torch.where(slot_mask_NK, student_logp_NK, neg_inf)
+        log_p_renorm = torch.log_softmax(masked_logp, dim=-1)
+        log_p_renorm = torch.nan_to_num(log_p_renorm, nan=0.0, neginf=0.0)
```
YujiaBao, line 653: This line is doing subtle work and is worth a one-line comment. When a full row has no valid slots, `log_softmax(all -inf)` produces NaN; when a row has *some* invalid slots, those slots get `-inf` from `log_softmax`. `nan_to_num(..., nan=0, neginf=0)` maps both to 0, which means `p_renorm.exp() = 1` at invalid slots — non-zero, but harmless because `* slot_mask_NK.float()` on L655/661/664 zeros the contribution.

Suggest:
```python
# Invalid slots become 0 here (→ p_renorm=1), but are masked out below via slot_mask_NK.
log_p_renorm = torch.nan_to_num(log_p_renorm, nan=0.0, neginf=0.0)
```

Also: `TestReverseKLCustomLoss` doesn't cover the all-masked-row case — worth a test so this guardrail can't silently regress.

_Review assisted with Claude Code._

tinker_cookbook/distillation/sdft.py
```diff
@@ -464,6 +465,269 @@ async def build_topk_distillation_datums(
...
+    learning_rate: float,
+    num_substeps: int,
+    metrics: dict[str, Any] | None = None,
+) -> None:
+    """Substep-pipelined custom forward/backward for the reverse-KL loss.
+
+    Mirrors :func:`tinker_cookbook.rl.train.train_step` but uses
+    ``forward_backward_custom_async`` with :func:`reverse_kl_custom_loss`.
+    The datums produced by :func:`build_reverse_kl_datums` carry
+    ``position_mask`` / ``k_valid`` / ``teacher_logprobs`` instead of the
```
YujiaBao: Stale docstring — these keys don't exist in the actual datum. `build_reverse_kl_datums` packs everything into `target_tokens` + `weights` (with `weights > 0` as the validity sentinel), exactly as described in that function's own docstring.

Suggest rewriting to something like:

> The datums produced by `build_reverse_kl_datums` carry only `target_tokens` and `weights` (with `0` in `weights` as the mask sentinel), so no custom mask stripping is needed.

Will confuse the next reader otherwise.

_Review assisted with Claude Code._

tinker_cookbook/distillation/sdft.py
```diff
@@ -464,6 +465,269 @@ async def build_topk_distillation_datums(
...
+    standard ``mask`` key, so no mask stripping is needed.
+    """
+    batches = split_list(data_D, min(num_substeps, len(data_D)))
+    if not batches:
+        return
+
+    adam_params = tinker.AdamParams(learning_rate=learning_rate, beta1=0.9, beta2=0.95, eps=1e-8)
```
YujiaBao, line 700: `beta1=0.9, beta2=0.95, eps=1e-8` are hardcoded here. If the forward-KL / IS paths accept these from `cfg`, this path should too for parity; if they don't, ignore. Consistency > configurability.

_Review assisted with Claude Code._

tinker_cookbook/distillation/sdft.py
```diff
@@ -690,8 +955,30 @@ async def main(
                 for question, golden_answer in zip(questions_P, golden_answers_P)
             ]
 
-            if cfg.topk > 0:
-                # Top-K CE distillation
+            if cfg.topk > 0 and cfg.reverse:
```
YujiaBao, line 966: Silent fallthrough: the combination `cfg.topk == 0 and cfg.reverse == True` doesn't match either of the first two branches and lands in the deprecated IS block on L1004, ignoring the user's `reverse=True`. Worth an explicit guard up-front:

```python
if cfg.reverse and cfg.topk == 0:
    raise ValueError("reverse=True requires topk>0 (analytical reverse KL runs over the teacher's top-K).")
```

Fails loudly instead of silently running the wrong objective.

_Review assisted with Claude Code._

tinker_cookbook/recipes/sdft/sdft_test.py
```diff
@@ -381,3 +383,187 @@ async def test_no_completion_tokens(self):
...
+            topk=K,
+            skip_first_n_tokens=3,
+        )
+        weights = rev_datums[0].loss_fn_inputs["weights"].to_torch()
+        # Of model_input length 6, only 2 positions should have any positive weight.
+        active_positions = (weights > 0).any(dim=-1).sum().item()
+        assert active_positions == 2
+
+
+class TestReverseKLCustomLoss:
```
YujiaBao, line 493: Coverage gap: no test for a row where *all K slots* are invalid (e.g., full position mask off). That's the path most likely to surface NaN from `log_softmax(all -inf)`, and it's guarded in the loss by `nan_to_num` on `sdft.py:651` — a dedicated test would prevent that guardrail from silently regressing.

_Review assisted with Claude Code._

YujiaBao (approved):
Solid PR — math checks out (REINFORCE form on renormalized `p` is gradient-equivalent to analytical reverse KL since `Σ ∂p/∂θ = 0` over the valid K slots), empirical +5.2pp vs old IS, good test coverage, backward-compatible default.

Approving with non-blocking inline suggestions:
- Stale docstring on `_train_step_reverse_kl` (references keys that don't exist in the actual datum encoding) — worth fixing before merge.
- Silent fallthrough for `topk=0, reverse=True` — suggest a loud `ValueError`.
- CPU-pinned `total_loss` init — latent device issue.
- One missing edge-case test (all-masked row).

_Review assisted with Claude Code._

likenneth:
@codex review

likenneth:
@YujiaBao all fixed — review comments addressed in 1adde89 and formatter fallout in b693e35, all checks are green. I don't have push access on the upstream repo, so could you help merge when you get a chance? Thanks!

PR #683, opened 2026-04-26
Fix plugin.json skills field validation

benediktstroebl:
Plugin install fails with `skills: Invalid input` because the manifest validator requires relative paths to start with `./`. Changes `"skills": "skills/"` → `"skills": "./skills/"`.


benediktstroebl:
@codex review

PR #688, opened 2026-04-29
Replace parse_success bool with ParseTermination enum (fixes base-model eval regression)

### LINKED ISSUE #685: Base models silently score 0% on every single-turn benchmark
Evaluating any base model (e.g. Qwen/Qwen3-8B-Base) on a single-turn
 benchmark returns 0% even when the model produced clean \boxed{N} answers.

The reason is that the `role_colon` is the recommended renderer for every base model in `model_info.py`. `RoleColonRenderer.parse_response` (renderers/role_colon.py:105, added intentionally in 259500c for R1-Zero
 training) returns `parse_success=False` when the response terminates with
 EOS. EnvFromMessageEnv.step (rl/message_env.py:128-137) then returns `failed_parse_reward=0.0` with `parse_error=1.0` and never calls the benchmark grader.

 Repro: run any of gsm8k / math500 / aime_* / mmlu_pro / gpqa /
 ifeval / etc. against Qwen/Qwen3-8B-Base. We saw 0/30 on AIME 2024;
 re-grading the same trajectories gave 5/30 = 16.7%.

 Scope: all 9 base models × ~10 single-turn benchmarks. Training paths
 don't use EnvFromMessageEnv, so they're unaffected. No tests cover
 base-model evaluation.

dphuang2:
## Summary

- **Fixes #685**: base models silently scored 0% on every single-turn benchmark because `RoleColonRenderer.parse_response` returned `parse_success=False` for EOS-terminated responses, and `EnvFromMessageEnv.step` short-circuits with `failed_parse_reward=0` when `parse_success` is False — so the grader was never invoked on the most common base-model response shape.
- **Replaces the `parse_success: bool` return value with a `ParseTermination` StrEnum** (`STOP_SEQUENCE` / `EOS` / `MALFORMED`) so each consumer expresses its policy explicitly instead of leaning on a bool whose meaning silently shifted between callers (the root cause of how #339 broke evals).
- **Preserves R1-Zero / math RL training behavior** via an explicit `ProblemEnv.require_stop_sequence_for_format=True` default — EOS-terminated rollouts continue to lose the format reward at the RL layer exactly as they did pre-fix.
- Re-ran AIME 2025 against `Qwen/Qwen3-8B-Base` on this branch: **6/30 = 20.0%, 0 errors** (would have been 0/30 before this fix).

## The bug, in one paragraph

`RoleColonRenderer` formats prompts as `User:/Assistant:` and uses `\n\nUser:` as its stop sequence. On single-turn evals, base models commonly produce a clean answer and then emit EOS instead of `\n\nUser:` — there's no second turn to start. Pre-PR, that shape returned `parse_success=False`, which `EnvFromMessageEnv.step` reads as \"skip the grader, return `failed_parse_reward=0`\". Result: every base model × every single-turn benchmark scored 0% even when the answer was correct.

## Why a single bool wasn't enough

The original bool was overloaded to answer three distinct consumer questions:

1. Eval grading (`EnvFromMessageEnv`): \"is this gradeable?\" — wants any clean termination.
2. Strict format reward (`ProblemEnv`, used by R1-Zero / math RL): \"did this earn format reward?\" — wants stop-sequence specifically; EOS-only is a soft penalty.
3. OpenAI `finish_reason` / preference RL / etc: variations on the same theme.

PR #339 changed `RoleColonRenderer` to make EOS-terminated responses return `False` for R1-Zero's training needs — and silently broke every eval that read the same bool. This PR fixes the conflation by giving each call site its own explicit signal.

### `Renderer.parse_response` returns `ParseTermination` instead of `bool`

`tinker_cookbook/renderers/base.py` adds:

```python
class ParseTermination(StrEnum):
    STOP_SEQUENCE = \"stop_sequence\"  # the renderer's expected stop signal fired
    EOS = \"eos\"                       # model emitted EOS, accepted as a fallback by some renderers
    MALFORMED = \"malformed\"           # truncated / multiple stop signals / etc.

    @property
    def is_clean(self) -> bool: ...        # STOP_SEQUENCE or EOS
    @property
    def is_stop_sequence(self) -> bool: ...  # STOP_SEQUENCE only
```

All 8 renderer implementations (`role_colon`, `llama3`, `qwen3`, `qwen3_5`, `kimi_k2`, `deepseek_v3`, `gpt_oss`, plus the `parse_response_for_stop_token` helper) and ~17 call sites are updated. Chat-template renderers return `STOP_SEQUENCE` / `MALFORMED` only — the `EOS` state is meaningful only for `RoleColonRenderer`, which has a stop sequence distinct from EOS.

### `RoleColonRenderer.parse_response` semantics

| Termination state | Returns |
|---|---|
| Ends with `\n\nUser:` only | `STOP_SEQUENCE` |
| Ends with EOS, no `\n\nUser:` | `EOS` (was `False` pre-PR — this is the fix) |
| No terminator at all | `MALFORMED` |
| Both `\n\nUser:` and EOS | `MALFORMED` (sampling should have stopped at `User:`) |
| Multiple `\n\nUser:` | `MALFORMED` (model role-played the user turn) |

### `ProblemEnv.require_stop_sequence_for_format=True` default

`tinker_cookbook/rl/problem_env.py` adds an explicit knob so the R1-Zero / math RL format reward shaping is unchanged by the renderer fix:

- `True` (default): `correct_format` requires `termination.is_stop_sequence` — EOS-terminated rollouts lose the `format_coef * (0 - 1) = -0.1` reward, exactly as they did pre-PR.
- `False`: `correct_format` requires only `termination.is_clean`, opting into the eval-aligned semantics. Recommended only after an A/B against an existing recipe.

`EnvFromMessageEnv` (eval grading) uses `termination.is_clean` directly and is unaffected by this knob.

### Preference RL / verifiers / litellm

`rl/preference_envs.py` keeps its strict format check via `termination.is_stop_sequence` (matches `ProblemEnv`'s default). `recipes/verifiers_rl/tinker_openai.py` and `third_party/litellm/provider.py` use `termination.is_clean` for the OpenAI `finish_reason` (`\"stop\"` vs `\"length\"`).

## Tests

- `tinker_cookbook/renderers/role_colon_test.py` — 5 cases covering each `ParseTermination` state.
- `tinker_cookbook/rl/problem_env_test.py` — both `require_stop_sequence_for_format` modes × EOS-only / stop-sequence / truncated responses.
- All existing renderer / RL / litellm tests updated to the new return type and pass.

## Test plan

- [x] New unit tests in `tinker_cookbook/renderers/role_colon_test.py` pass on this branch and fail on `main` (regression coverage verified).
- [x] New unit tests in `tinker_cookbook/rl/problem_env_test.py` cover both `require_stop_sequence_for_format` modes against EOS-only, stop-sequence, and truncated responses.
- [x] Full local test suite green (1287 passed, excluding test files that require optional deps not installed locally — search_tool, verifiers_rl, harbor_rl, math_rl, text_arena, sandbox, trace). litellm tests pass separately (18/18).
- [x] Pyright clean on touched files.
- [x] Re-ran AIME 2025 (cookbook ships `aime_2025`/`aime_2026`; AIME 2024 is not in the registry) against `Qwen/Qwen3-8B-Base` on this branch: **6/30 = 20.0%, 0 errors, 5 truncated** in 199s. Non-zero score confirms the grader is now invoked on EOS-terminated responses; the value is in the expected range for a base 8B on AIME, ruling out a regression in the opposite direction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,17 @@ Each entry includes:
 
 ---
 
+### [cookbook] Fix base-model single-turn evals + replace `parse_success` bool with `ParseTermination` enum ([#688](https://github.com/thinking-machines-lab/tinker-cookbook/pull/688))
+**Date:** 2026-04-30
+**Type:** fix
+**Tags:** renderers, eval, rl
+
+Fixes #685: every base model × every single-turn benchmark scored 0% because `RoleColonRenderer.parse_response` reported `parse_success=False` on EOS-terminated responses, and `EnvFromMessageEnv.step` short-circuits with `failed_parse_reward=0` in that case — so the grader was never invoked. Re-running AIME 2025 against `Qwen/Qwen3-8B-Base` now gives 6/30 (20.0%) instead of 0/30.
+
+**Breaking API change.** `Renderer.parse_response` now returns `tuple[Message, ParseTermination]` instead of `tuple[Message, bool]`. `ParseTermination` is a `StrEnum` with `STOP_SEQUENCE` / `EOS` / `MALFORMED` values plus `is_clean` and `is_stop_sequence` properties. Note that `ParseTermination` is truthy in all three states (it's a `StrEnum`), so `if not parse_success:` patterns will silently always be False without raising `TypeError`. Migration: replace with `if not termination.is_clean:` (lenient, what eval grading reads) or `if not termination.is_stop_sequence:` (strict, what RL format-reward shaping reads). All in-tree call sites are updated. R1-Zero / math RL training behavior is preserved by default via a new `ProblemEnv.require_stop_sequence_for_format=True` knob.
+
+---
+
 ### [cookbook] Fix grading bugs in IFEval, MATH-500, and GPQA benchmarks ([#643](https://github.com/thinking-machines-lab/tinker-cookbook/pull/643))
 **Date:** 2026-04-08
 **Type:** fix
--- a/tinker_cookbook/completers.py
+++ b/tinker_cookbook/completers.py
@@ -190,7 +190,7 @@ class TinkerMessageCompleter(MessageCompleter):
         )
 
         # Decode the response
-        parsed_message, _success = self.renderer.parse_response(response.sequences[0].tokens)
+        parsed_message, _termination = self.renderer.parse_response(response.sequences[0].tokens)
 
         result: renderers.Message = {"role": "assistant", "content": parsed_message["content"]}
         if "tool_calls" in parsed_message:
--- a/tinker_cookbook/distillation/datasets.py
+++ b/tinker_cookbook/distillation/datasets.py
@@ -112,7 +112,7 @@ class PromptOnlyEnv(ProblemEnv):
 
     async def step(self, action: Action, *, extra: ActionExtra | None = None) -> StepResult:
         """Return zero reward always."""
-        message, parse_success = self.renderer.parse_response(action)
+        _message, _termination = self.renderer.parse_response(action)
         return StepResult(
             reward=0.0,
             episode_done=True,
--- a/tinker_cookbook/distillation/datasets.py
+++ b/tinker_cookbook/distillation/datasets.py
@@ -112,7 +112,6 @@ class PromptOnlyEnv(ProblemEnv):
 
     async def step(self, action: Action, *, extra: ActionExtra | None = None) -> StepResult:
         """Return zero reward always."""
-        _message, _termination = self.renderer.parse_response(action)
         return StepResult(
             reward=0.0,
             episode_done=True,
--- a/tinker_cookbook/recipes/math_rl/arithmetic_env.py
+++ b/tinker_cookbook/recipes/math_rl/arithmetic_env.py
@@ -21,8 +21,15 @@ class ArithmeticEnv(ProblemEnv):
         y: int,
         renderer: renderers.Renderer,
         convo_prefix: list[renderers.Message] | None = None,
+        format_coef: float = 0.1,
+        require_stop_sequence_for_format: bool = True,
     ):
-        super().__init__(renderer, convo_prefix)
+        super().__init__(
+            renderer,
+            convo_prefix,
+            format_coef=format_coef,
+            require_stop_sequence_for_format=require_stop_sequence_for_format,
+        )
         self.x = x
         self.y = y
 
--- a/tinker_cookbook/recipes/math_rl/math_env.py
+++ b/tinker_cookbook/recipes/math_rl/math_env.py
@@ -28,8 +28,15 @@ class MathEnv(ProblemEnv):
         convo_prefix: list[renderers.Message] | None = None,
         grader: Literal["sympy", "math_verify"] = "sympy",
         timeout: float = 1.0,
+        format_coef: float = 0.1,
+        require_stop_sequence_for_format: bool = True,
     ):
-        super().__init__(renderer, convo_prefix)
+        super().__init__(
+            renderer,
+            convo_prefix,
+            format_coef=format_coef,
+            require_stop_sequence_for_format=require_stop_sequence_for_format,
+        )
         self.problem = problem
         self.answer = answer
         self.grader = grader
--- a/tinker_cookbook/recipes/multiplayer_rl/guess_number/env.py
+++ b/tinker_cookbook/recipes/multiplayer_rl/guess_number/env.py
@@ -72,7 +72,7 @@ class GuessNumberEnv(Env):
     async def step(self, action: Action, *, extra: ActionExtra | None = None) -> StepResult:
         # step 1: parse the action tokens into a message
         # this step is specific to our library, but usually templated, so you can just copy it.
-        (action_message, _parse_success) = self.renderer.parse_response(action)
+        (action_message, _termination) = self.renderer.parse_response(action)
 
         # step 2: based on the string answer, we compute the reward and the user turn.
         # This part is NOT templated, so you need to implement it. But it is plain python without using special libraries.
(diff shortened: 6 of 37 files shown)


dphuang2:
@codex review

YujiaBao (commented):
Nice diagnosis — the bool was overloaded across consumers, and a one-line "treat EOS as clean in RoleColon" fix would have just kicked the next #339 down the road. The enum is the right call. Empirical AIME validation looks solid and the strict-mode preservation traces bit-identical to pre-PR.

A few things before merge, then some nits.

**Add a regression test at the actual bug surface.** The bug is `EnvFromMessageEnv.step` × RoleColon × EOS → grader silently skipped. The new tests cover the renderer and `ProblemEnv`, but not the path that #685 actually flagged. The mock in `message_env_test.py:_make_renderer` maps `parse_success: bool → STOP_SEQUENCE/MALFORMED` and can't even synthesize `EOS`. ~10 lines that mock `parse_response` to return `(msg, ParseTermination.EOS)` and assert `MessageEnv.step` was invoked would lock this in.

**Tutorial 201 still teaches the old API.** `tutorials/201_rendering.py:145-148`:
```python
parsed_message, parse_success = renderer.parse_response(fake_tokens)
print(f"Parse success: {parse_success}")
```
Now prints `Parse success: ParseTermination.STOP_SEQUENCE` and the prose around it (line 132-136) calls it a bool. Worth updating since 201 is the canonical entry point.

**Three sites changed strictness for RoleColon beyond what #685 needs:**

| Site | Pre-PR for RoleColon EOS | Post-PR |
|---|---|---|
| `recipes/rubric/env.py:97` | format_score=0 | format_score=1 |
| `recipes/verifiers_rl/tinker_openai.py:105` | finish_reason="length" | "stop" |
| `third_party/litellm/provider.py:175` | finish_reason="length" | "stop" |

Each could've used `is_stop_sequence` to preserve. Going lenient is defensible but isn't flagged in the description as a deliberate change. The rubric one is most off — its comment at `env.py:142` says "Apply format penalty similar to ProblemEnv" but ProblemEnv now defaults to strict. Mild preference: tighten these to `is_stop_sequence` so the PR stays purely about #685. The variable name `parse_success_bool` and metric keys `format_parse_success` / `parse_success` in rubric/env.py are stale either way.

**`require_stop_sequence_for_format` is unreachable from `MathEnv`.** `recipes/math_rl/math_env.py:32` calls `super().__init__(renderer, convo_prefix)`, so for the most common ProblemEnv consumer the new knob can't be flipped without subclassing. Either forward it through `MathEnv.__init__`, or drop the kwarg and hardcode strict.

**Stale comment at `problem_env.py:65`** — "renderer's lenient parse_success" → `termination.is_clean`.

Nits:
- `role_colon.py:91-93` has `import logging; logger = ...` inside `parse_response`. Module level is cheaper since you're touching it.
- `role_colon.py:106-110` (the truncated `len==1, no EOS` branch) has no inline comment while the symmetric `len==2 + EOS` branch at 113-118 does.
- `distillation/datasets.py:115`: both unpacked names are `_`-prefixed, return is zero reward. Either drop the call or note why it's kept.
- Mock helpers in `message_env_test.py` and `litellm/provider_test.py` still take `parse_success: bool` and translate. Taking `termination: ParseTermination` directly would let tests drive `EOS` without going through RoleColon.
- Worth a CHANGELOG "Breaking" entry: `ParseTermination` is a `StrEnum`, so `bool(ParseTermination.MALFORMED) == True`. External code with `if not parse_success:` will silently always be False — no `TypeError`. Migration: `if not termination.is_clean:`.

Verified the inheritance chain (Nemotron3 → Qwen3_5 → Qwen3VL → Qwen3, KimiK25/K26 → KimiK2, DeepSeekV3Thinking/Disable → _DeepSeekV3BaseRenderer) — all covered transitively. Pyright clean on touched files.

*Review assisted by Claude Code.*

dphuang2:
Thanks @YujiaBao — addressed in 4199bf5. Per-point:

**Required**

- ✅ **Regression test at the bug surface.** Added `test_eos_termination_invokes_grader` and `test_malformed_termination_skips_grader` in `message_env_test.py`. The mock helper now also accepts `termination: ParseTermination` directly so tests can drive `EOS` without going through RoleColon.
- ✅ **Tutorial 201.** Updated the snippet and the surrounding prose to teach `ParseTermination` with `is_clean` / `is_stop_sequence` instead of a bool.
- ✅ **Three sites tightened to `is_stop_sequence`.** You're right — pre-PR for RoleColon EOS, all three (`rubric/env.py`, `verifiers_rl/tinker_openai.py`, `litellm/provider.py`) reported strict failure / `finish_reason="length"`, and the enum refactor accidentally went lenient. All three now use `is_stop_sequence`. Also renamed `parse_success_bool` → `format_valid` and dropped the stale `format_parse_success` / `parse_success` log keys in rubric. Added a `test_eos_termination_gives_length_finish` to litellm to lock the strictness in.
- ✅ **`require_stop_sequence_for_format` forwarding.** Added to `MathEnv.__init__` and `ArithmeticEnv.__init__` (the two `ProblemEnv` subclasses with non-trivial __init__ — `PromptOnlyEnv` has `format_coef=0` so the knob is a no-op there).
- ✅ **Stale comment at `problem_env.py:65`** — replaced with `termination.is_clean`.

**Nits**

- ✅ Hoisted `import logging` to module scope in `role_colon.py`.
- ✅ Added inline comment on the truncated/no-EOS branch.
- ✅ Dropped the unused `parse_response` call in `distillation/PromptOnlyEnv.step`.
- ✅ Mock helpers (`_make_renderer`, `_make_sampling_result`) now accept `termination: ParseTermination` directly; old `parse_success: bool` kept for back-compat.
- ✅ CHANGELOG entry flagging the breaking API change, with migration guidance pointing at `is_clean` / `is_stop_sequence`. Specifically called out the StrEnum truthiness footgun (`bool(ParseTermination.MALFORMED) == True`).

Test sweep: 1290 passed (+3 new tests), pyright clean, ruff clean. Waiting on remote CI.

YujiaBao (approved):

PR #692, opened 2026-05-06
Add cross-account checkpoint copy recipe

derek-tml:
## Summary
- Add a `copy_checkpoint` script that copies trainable `weights/...` checkpoints across accounts using a source checkpoint access token and destination API key.
- Document usage, security model, supported output kinds, and the sampler-only source checkpoint limitation.
- Link the recipe from the main recipes README.

## Test plan
- `uv run --no-sync python -m py_compile tinker_cookbook/scripts/copy_checkpoint.py`
- `uv run --no-sync python -m tinker_cookbook.scripts.copy_checkpoint --help`
- Manually verified cross-account copy with a `weights/...` source to destination `sampler_weights/...`.

Made with [Cursor](https://cursor.com)


tinker_cookbook/recipes/cross_account_checkpoint/README.md
```diff
@@ -0,0 +1,58 @@
+# Cross-Account Checkpoint Copy
```
dphuang2: I think recipes should be reserved for post-training recipes

tinker_cookbook/scripts/copy_checkpoint.py
```diff
@@ -0,0 +1,130 @@
...
+This creates a fresh LoRA training run under the destination account, loads the
+source checkpoint using the source account's API key as `weights_access_token`,
+and saves a new destination-owned checkpoint.
+
+Note: this only works for trainable `weights/...` checkpoints, not sampler-only
+`sampler_weights/...` checkpoints.
+
+Usage:
+    export SRC_TINKER_API_KEY=...   # owns the source checkpoint
+    export DST_TINKER_API_KEY=...   # owns the copied checkpoint
```
andriigrynenko: I would remove the destination api key and rely on existing tinker SDK auth flow. This way is we have more auth mechanisms (e.g. not just api key) you won't need to update this script

tinker_cookbook/scripts/copy_checkpoint.py
```diff
@@ -0,0 +1,130 @@
...
+) -> None:
+    if not TRAINING_WEIGHTS_PATH_RE.search(source_path):
+        raise SystemExit(
+            f"Source path must be a tinker://.../weights/<name> checkpoint, "
+            f"got {source_path!r}. Sampler checkpoints cannot be copied "
+            "directly; use the matching weights/... path."
+        )
+    name = output_name or checkpoint_name_from_path(source_path)
+
+    # TODO: once token-scoped metadata lookup works in the SDK, replace this
```
andriigrynenko: make this conditional on SDK version?

tinker_cookbook/scripts/copy_checkpoint.py
```diff
@@ -0,0 +1,130 @@
...
+
+This creates a fresh LoRA training run under the destination account, loads the
+source checkpoint using the source account's API key as `weights_access_token`,
+and saves a new destination-owned checkpoint.
+
+Note: this only works for trainable `weights/...` checkpoints, not sampler-only
+`sampler_weights/...` checkpoints.
+
+Usage:
+    export SRC_TINKER_API_KEY=...   # owns the source checkpoint
```
andriigrynenko: nit: maybe call this access token and say that api key can be used for now - maybe we'll have a more granular access tokens here later which are more secure (by having narrower scope)

tinker_cookbook/scripts/copy_checkpoint.py
```diff
@@ -0,0 +1,130 @@
...
+
+    source_client = tinker.ServiceClient(api_key=source_api_key)
+    weights_info = (
+        source_client.create_rest_client().get_weights_info_by_tinker_path(source_path).result()
+    )
+    if not weights_info.is_lora or weights_info.lora_rank is None:
+        raise SystemExit(
+            f"Source checkpoint {source_path!r} is not a LoRA run; "
+            "non-LoRA copies are not supported."
+        )
```
andriigrynenko: I don't think this is limited to lora in any way

andriigrynenko (changes requested):

dphuang2 (approved):

PR #693, opened 2026-05-07
HarborReward._upload_tests: recurse into nested fixture dirs

danqi:
Path.iterdir silently skips directories, so test fixtures placed in subdirs of `tests_dir` (e.g. `tests/repo_overlay/`) never get uploaded into the sandbox. `test.sh` scripts that do `cp -r /tests/repo_overlay/. /repo/` then become no-ops and grading fails for reasons that are not visible in per-task logs.

Switch to `rglob` and preserve the relative path. `modal_sandbox.write_file` already `mkdir -p`s the parent directory, so nested target paths just work. Test updated to assert recursive upload and that nested files are not marked executable.


ZitongYang (approved):

nealwu (approved):
approving to unblock @ZitongYang's review!

PR #695, opened 2026-05-07
Fix `get_lora_param_count` with model-specific per-component table

nealwu:
Removes the previous implementation, which was off by 10x or more for many newer models, plus required network access and HF auth on gated repos and was often extremely slow.

Now mirrors `create_lora_training_client`'s signature so callers can predict the exact trainable parameter count for any Tinker base model:

```
get_lora_param_count(model_name, lora_rank=32,
                     train_mlp=True, train_attn=True, train_unembed=True)
```

```
total = lora_rank * (mlp * train_mlp +
                     attn * train_attn +
                     unembed * train_unembed)
```

Removed kwargs: `detailed`, `include_experts`, `shared_expert_outer_loras`.

Tests verify the function reproduces an independently-maintained reference table for all 31 Tinker base models across 7 ranks and 7 valid flag combinations (217 parametrizations × 7 ranks).

Next up: updating docs to reflect the new signature.


opherlieber:
Maybe add a test to validate that all models in model_info support get_lora_param_count?

nealwu:
> Maybe add a test to validate that all models in model_info support get_lora_param_count?

hmm I like this idea! but the problem is it has a bunch of models that aren't actually on Tinker:

```
8 models live in model_info but not in our params dict — none of them appear in models.json (Tinker's served-models list), so they likely exist in model_info for renderer/HF-metadata purposes only:

Qwen/Qwen3-0.6B
Qwen/Qwen3-1.7B
Qwen/Qwen3-4B
Qwen/Qwen3-4B-Base
Qwen/Qwen3-14B
Qwen/Qwen3-14B-Base
meta-llama/Llama-3.2-1B-Instruct
meta-llama/Llama-3.2-3B-Instruct
```

erikwijmans (approved):

simonguozirui:
LGTM after @nealwu walked me through his measurement methodology (I also ran some to validate) and mirroring of the class signature is quite neat. So we need to update this list for every model release? Also that `mlp_attn_extra` abstraction hold up for the future (right now just for ssm part of NemotronH) 

Do note if developers and researchers want to compute the number of params themselves and cross reference, we could let them know LoRA counts for MoEs are using shared-outer-LoRA for experts.

tests/downstream_compat/test_cli_and_hyperparam.py
```diff
@@ -30,11 +323,55 @@ def test_get_lora_lr_over_full_finetune_lr_signature(self):
...
+        "flag_combo",
+        sorted(
+            {combo for params in _REFERENCE_PARAMS_PER_RANK.values() for combo in params},
+            reverse=True,
+        ),
+    )
+    @pytest.mark.parametrize("model_name", sorted(_REFERENCE_PARAMS_PER_RANK.keys()))
+    def test_get_lora_param_count_matches_measurements(
+        self, model_name: str, flag_combo: tuple[bool, bool, bool]
+    ) -> None:
```
simonguozirui, line 358: nit: this test and _LORA_PARAMS_PER_RANK_BY_COMPONENT seem to come from same measurement, so mostly testing adding up the components, is that the goal of the test? that won't cover any actual lora count measurement issues.

simonguozirui (approved):

tests/downstream_compat/test_cli_and_hyperparam.py
nealwu: the difference is the test doesn't assume addition of components and actually measures each configuration, so if there are any inconsistencies (like we saw here) the test would help pick it up

nealwu:
> LGTM after @nealwu walked me through his measurement methodology (I also ran some to validate) and mirroring of the class signature is quite neat. So we need to update this list for every model release? Also that `mlp_attn_extra` abstraction hold up for the future (right now just for ssm part of NemotronH)
> 
> Do note if developers and researchers want to compute the number of params themselves and cross reference, we could let them know LoRA counts for MoEs are using shared-outer-LoRA for experts.

(the `mlp_attn_extra` abstraction is removed now FYI, was not correct)

and on "shared-outer-LoRA for experts" sounds good, I'll update the docs to explain this and have you help double check it!

PR #699, opened 2026-05-11
Add RL rollout sample concurrency cap

derek-tml:
## Summary
- Add an optional `max_concurrent_samples` knob to the shared RL trainer config.
- Thread a shared in-process semaphore through sync, stream-minibatch, and async rollout paths so concurrent `SamplingClient` requests can be capped.
- Add unit coverage for the semaphore cap and incompatible executor configuration.

## Why
Some RL workloads issue many rollout sampling requests concurrently. This option gives users a way to apply backpressure to sampling without changing environment or rollout strategy code.

The guard against `rollout_executor` is intentional: `asyncio.Semaphore` is local to an event loop/process and would not provide a correct global cap once rollouts are dispatched through a separate executor.

## Test plan
- `uv run pytest tinker_cookbook/rl/rollout_error_resilience_test.py -q`


derek-tml:
@codex please review

derek-tml:
@codex review again

derek-tml:
@codex review again

PR #711, opened 2026-05-15
Fix flaky `test_quantize_dequantize_roundtrip`

nealwu:
The INT4 group-quant roundtrip test used `atol=0.5` on unseeded random input, but for `std=2` inputs over groups of 32 the theoretical max per-value error (~group_max/14) routinely lands around 0.55-0.65 on the tail. Observed flake rate on main: ~17% across 10k seeds (in particular, it failed for me on https://github.com/thinking-machines-lab/tinker-cookbook/pull/710 twice in a row).

Loosen atol to 0.75. Empirical max-error across 10k seeds was 0.7461, so flakes drop to < 1 in 10,000. Test intent (quant/dequant is its own inverse to bounded error) is preserved; the precision floor is checked separately by `test_dequantize_known_values`.

Also assert that the 95th percentile is less than 0.5, which empirically has a < 1 in 10,000 chance of flakes as well.

### DIFF
--- a/tinker_cookbook/weights/merge_test.py
+++ b/tinker_cookbook/weights/merge_test.py
@@ -1764,6 +1764,10 @@ class TestPackedInt4:
         # empirical max-error across 10k random seeds was 0.7461; atol=0.75 keeps
         # the structural-correctness check while making flakes < 1 in 10,000.
         assert torch.allclose(tensor.float(), dequantized.float(), atol=0.75)
+        # Tighter body-of-distribution check: 95% of the 512 values should be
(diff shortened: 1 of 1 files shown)


nealwu:
oof, it just failed again on `main`: https://github.com/thinking-machines-lab/tinker-cookbook/actions/runs/25929209504/job/76218477452

simonguozirui (approved):
Seems reasonable assumption for int4 roundtrip (for the Kimi MLPs), especially after consulting with @jackcook. Let's do note distribution and any assumptions we made on this.

PR #722, opened 2026-05-19
Test: `get_lora_param_count` covers every `model_info` entry

nealwu:
Now that `model_info` is cleaned up, can finally implement the idea from this comment: https://github.com/thinking-machines-lab/tinker-cookbook/pull/695#issuecomment-4400411038

Adds a parametric test over every model registered in ``tinker_cookbook.model_info`` that asserts ``get_lora_param_count`` returns a positive value. Catches drift when a new model is added to the registry without measuring and registering its LoRA component counts in ``_LORA_PARAMS_PER_RANK_BY_COMPONENT``.


opherlieber (approved):

simonguozirui (approved):
check makes sense to me

PR #727, opened 2026-05-20
Update checkpoint copy script

derek-tml:
## Summary

- Make `--source-access-token` optional for same-org checkpoint copies, defaulting from `SRC_TINKER_ACCESS_TOKEN` only when provided.
- Add `--destination-project-id` / `--target-project-id` so copied checkpoints can be saved into a target project.
- Remove the old manual metadata/loading fallback path while keeping an explicit `tinker>=0.19.0` check that tells users to upgrade old SDK installs.

## Validation

- `uv run --with pre-commit pre-commit run ruff-format --files tinker_cookbook/scripts/copy_checkpoint.py`
- `uv run --with pre-commit pre-commit run ruff-check --files tinker_cookbook/scripts/copy_checkpoint.py`
- `uv run ruff check tinker_cookbook/scripts/copy_checkpoint.py`
- `uv run python -m py_compile tinker_cookbook/scripts/copy_checkpoint.py`
- `uv run python -m tinker_cookbook.scripts.copy_checkpoint --help`
- Manually exercised the old-SDK branch with simulated `0.18.0`.

### DIFF
--- a/tinker_cookbook/scripts/copy_checkpoint.py
+++ b/tinker_cookbook/scripts/copy_checkpoint.py
@@ -1,19 +1,27 @@
 """Copy trainable Tinker weights into the currently authenticated account.
 
-This copies a trainable Tinker checkpoint by loading the source weights with an
-access token and saving them as a new destination-owned checkpoint.
+This copies a trainable Tinker checkpoint by loading the source weights and
+saving them as a new destination-owned checkpoint.
 
 Note: this only works for trainable `weights/...` checkpoints, not sampler-only
 `sampler_weights/...` checkpoints.
 
 Usage:
-    export SRC_TINKER_ACCESS_TOKEN=...  # grants access to the source checkpoint;
-                                        # a Tinker API key works for now
+    python -m tinker_cookbook.scripts.copy_checkpoint \\
+        --source-path tinker://<run-id>:train:0/weights/<name>
+    # Prints: sampler_path: tinker://<new-run-id>:train:0/sampler_weights/<name>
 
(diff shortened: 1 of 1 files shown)


tinker_cookbook/scripts/copy_checkpoint.py
```diff
@@ -43,39 +52,12 @@ def copy_checkpoint(
...
-            else True,
-            user_metadata={"copied_from_path": source_path},
-        )
-        training_client.load_state(source_path, weights_access_token=source_access_token).result()
+    destination_client = tinker.ServiceClient(project_id=destination_project_id)
+    training_client = destination_client.create_training_client_from_state(
+        source_path,
+        weights_access_token=source_access_token,
+        user_metadata={"copied_from_path": source_path},
+    )
```
dphuang2: oh, is this just better

tinker_cookbook/scripts/copy_checkpoint.py
```diff
@@ -43,39 +52,12 @@ def copy_checkpoint(
         )
     name = output_name or checkpoint_name_from_path(source_path)
 
-    destination_client = tinker.ServiceClient()
-    tinker_version = importlib.metadata.version("tinker").split(".")
-
-    # SDK >=0.19.0 can fetch source checkpoint metadata with weights_access_token and is preferred.
```
andriigrynenko, line 49: maybe assert on SDK version so that if someone happens to use this with an old SDK they know to upgrade?

andriigrynenko (approved):

PR #735, opened 2026-05-26
Add `lora_init_seed` config for reproducible LoRA initialization

Butanium:
## Summary

- Add an optional `lora_init_seed: int | None = None` field to the SFT (`tinker_cookbook/supervised/train.py`), RL (`tinker_cookbook/rl/train.py`), and DPO (`tinker_cookbook/preference/train_dpo.py`) `Config` dataclasses.
- Thread the field into `service_client.create_lora_training_client[_async](..., seed=...)` in all three trainers.
- Default `None` preserves current behavior (SDK picks a random seed per `tinker.ServiceClient.create_lora_training_client_async(seed=None)`); setting it makes the LoRA adapter initialization reproducible across runs with the same config.

## Motivation

I want to study sensitivity to LoRA adapter initialization (running the same training config across seeds with everything else held fixed). Today the trainers don't expose the SDK's `seed=` knob, so the adapter init is non-deterministic and not controllable from the cookbook layer.

🤖 Drafted with [Claude Code](https://claude.com/claude-code) — Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>


Butanium (commented):
that was pretty useful for my experiments

PR #737, opened 2026-05-27
[checkpoint] add options for saving every N tokens or M seconds

klshuster:
Adds the ability to specify `save_every_tokens` or `save_every_seconds`, which do as they describe: save checkpoints after a certain amount of tokens or seconds. can be combined with `save_every` to save as frequently as one desires

### DIFF
--- a/tinker_cookbook/checkpoint_utils.py
+++ b/tinker_cookbook/checkpoint_utils.py
@@ -2,6 +2,7 @@ import asyncio
 import dataclasses
 import json
 import logging
+import time
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal
@@ -566,6 +567,8 @@ class CheckpointManager:
         service_client: tinker.ServiceClient,
         log_path: str,
         save_every: int = 0,
+        save_every_tokens: int = 0,
+        save_every_seconds: float = 0.0,
         ttl_seconds: int | None = 604800,
         rolling_save_every: int = 0,
         rolling_ttl_seconds: int = 7200,
@@ -576,6 +579,8 @@ class CheckpointManager:
         self._service_client = service_client
         self._log_path = log_path
         self._save_every = save_every
+        self._save_every_tokens = save_every_tokens
+        self._save_every_seconds = save_every_seconds
         self._ttl_seconds = ttl_seconds
         self._rolling_save_every = rolling_save_every
         self._rolling_ttl_seconds = rolling_ttl_seconds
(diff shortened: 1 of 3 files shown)


tinker_cookbook/checkpoint_utils.py
```diff
@@ -586,13 +591,30 @@ def __init__(
...
+
     # ------------------------------------------------------------------
     # Periodic checkpoints
     # ------------------------------------------------------------------
 
-    def should_save_periodic(self, step: int) -> bool:
-        """Return True if *step* warrants a periodic checkpoint."""
-        return self._save_every > 0 and step > 0 and step % self._save_every == 0
+    def restore_last_saved_tokens(self, last_saved_tokens: int) -> None:
+        self._last_saved_tokens = last_saved_tokens
```
derek-tml, line 602: should we have a `_last_saved_wall_time` equivalent??

tinker_cookbook/checkpoint_utils.py
```diff
@@ -586,13 +591,30 @@ def __init__(
...
+
+    def should_save_periodic(self, step: int, *, elapsed_tokens: int = 0) -> bool:
+        """Return True if *step*, *elapsed_tokens*, or *elapsed_seconds* warrants a periodic checkpoint."""
```
derek-tml: nit: `elapsed_seconds` is not a term, perhaps confusing.

tinker_cookbook/supervised/train.py
```diff
@@ -125,7 +125,12 @@ class Config:
     # Checkpointing and evaluation (0 = disabled for *_every fields)
     evaluator_builders: list[EvaluatorBuilder] = chz.field(default_factory=list)
     infrequent_evaluator_builders: list[EvaluatorBuilder] = chz.field(default_factory=list)
+    # Step-based periodic checkpoint cadence
     save_every: int = 20
+    # Token-based periodic checkpoint cadence
+    save_every_tokens: int = 0
+    # Wall-clock periodic checkpoint cadence in seconds
+    save_every_seconds: float = 0.0
```
derek-tml, line 133: just wondering, do we want to expose these knobs in RL too?

derek-tml (approved):

tinker_cookbook/checkpoint_utils.py
klshuster: hmm i'm a little worried about that because wall time between runs can be quite variable. I think within a training run it makes sense but across... not sure

tinker_cookbook/supervised/train.py
klshuster: we can! I think the `save_every_seconds` might be nice, `save_every_tokens` might be tricky (do we mean trainer tokens or sampler tokens?). in any case, i think that would be a good follow-up PR

PR #740, opened 2026-05-28
Update recipe rerun results and assets

derek-tml:
## Summary
- refresh recipe README results for chat SL, code RL, math RL, rubric, SDFT, and true-thinking-score
- add rubric reward plot assets and rubric grader parsing fixes
- add the Harbor eval helper and update Harbor recipe docs/scripts


ZitongYang (approved):
`harbor_rl` looks good to me!

tinker_cookbook/recipes/chat_sl/README.md
```diff
@@ -30,7 +30,7 @@ python -m tinker_cookbook.recipes.chat_sl.train \
     wandb_project=cookbook_sl
 ```
 
-After 1740 steps of training, `test/nll` decreases to 0.50.
+After 7000 steps of training, `test/nll` decreases to 0.641.
```
nealwu: 👀 why did we increase training steps by so much? (how long does this take to run?)

tinker_cookbook/recipes/code_rl/README.md
```diff
@@ -49,22 +49,23 @@ Optional environment variables for Modal:
...
-    model_name="Qwen/Qwen3-4B-Instruct-2507" \
+    model_name="Qwen/Qwen3.5-4B" \
     group_size=8 groups_per_batch=128 \
     learning_rate=4e-5 \
     lora_rank=32 \
     max_tokens=24576
 ```
 
-After 100 steps of training, you can expect the following performance on **LiveCodeBench v6 (2025.02–2025.05)**:
+After 190 steps of training, you can expect the following performance on **LiveCodeBench v6 (2025.02–2025.05)**:
```
nealwu, line 63: similar question here

tinker_cookbook/recipes/code_rl/README.md
```diff
@@ -49,22 +49,23 @@ Optional environment variables for Modal:
...
 
 | Model | Pass@1 | Pass@8 |
 |-------|--------|--------|
-| Qwen3-4B-Instruct-2507 (before training) | 33.8% | 44.3% |
-| Qwen3-4B-Instruct-2507 (after 100 steps) | 42.7% | 55.0% |
+| Qwen3.5-4B (before training) | 23.6% | 34.3% |
+| Qwen3.5-4B (after 190 steps) | 52.4% | 76.0% |
```
nealwu, line 68: interesting, so it was actually worse before training but then even better afterward

tinker_cookbook/recipes/harbor_rl/README.md
```diff
@@ -111,22 +118,27 @@ uvx harbor datasets download swebench-verified@1.0 -o ~/.cache/harbor/tasks/sweb
...
 
 | Benchmark | Total | PASS | FAIL | ERROR | Pass Rate |
 |-----------|-------|------|------|-------|-----------|
-| SWE-Bench Verified 1.0 | 500 | 46 (9.2%) | 52 (10.4%) | 402 (80.4%) | 9.2% |
-| Terminal-Bench 2.0 | 89 | 18 (20.2%) | 36 (40.4%) | 35 (39.3%) | 20.2% |
+| SWE-Bench Verified 1.0 | 500 | 145 (29.0%) | 52 (10.4%) | 303 (60.6%) | 29.0% |
+| Terminal-Bench 2.0 | 89 | 14 (15.7%) | 31 (34.8%) | 44 (49.4%) | 15.7% |
 
-**Config**: `max_turns=200, max_tokens=8192, temperature=0.1, sandbox_timeout=3600s`
+**Config**: `max_turns=200, max_tokens=8192, temperature=1.0, sandbox_timeout=3600s`
```
nealwu, line 141: why the major temperature change?

tinker_cookbook/recipes/math_rl/README.md
```diff
@@ -13,65 +13,39 @@ uv pip install 'tinker-cookbook[math-rl] @ git+https://github.com/thinking-machi
...
-
-- 3+5+13=21
-
-- 3+7+11=21
-
-- 5+7+11=23
-
-So the best possible is sum1=20 and sum2=21, product 420. So the maximum sum is 420. \boxed{420}</think>
+The point (0,3): x=0, y=3. r = sqrt(0^2+3^2)=3. \theta = arctan(y/x) but x=0, y>0 so \theta=\pi/2. So (3, \pi/2). Format: \boxed{(3,\frac{\pi}{2})}.
+</think>
```
nealwu, line 37: where did we get these new examples from?

tinker_cookbook/recipes/rubric/env.py
```diff
@@ -229,11 +228,14 @@ class RubricGradedDatasetBuilder(RLDatasetBuilder):
...
-    grader_llm_name: str = "Qwen/Qwen3-30B-A3B-Instruct-2507"
+    grader_llm_name: str = "Qwen/Qwen3.6-35B-A3B"
+    grader_renderer_name: str | None = None
 
     def _get_grader_llm(self) -> MessageCompleter:
         tokenizer = get_tokenizer(self.grader_llm_name)
-        renderer_name = model_info.get_recommended_renderer_name(self.grader_llm_name)
+        renderer_name = self.grader_renderer_name or model_info.get_recommended_renderer_name(
+            self.grader_llm_name
+        )
```
nealwu, line 238: as a heads up this returns the thinking renderer by default. not saying that's good or bad, just double checking that's what you wanted

nealwu (approved):

tinker_cookbook/recipes/rubric/env.py
nealwu: based on https://github.com/thinking-machines-lab/tinker-cookbook/blob/b9a494708cae396460d7ed5f842fb6eba723ea7e/tinker_cookbook/model_info.py#L24 (first one is the default)

tinker_cookbook/recipes/harbor_rl/README.md
derek-tml: the reason is because training happens at temperature=1.0, so i changed the eval to match. i believe it should always have been this.

tinker_cookbook/recipes/math_rl/README.md
derek-tml: this polar coordinate is item 0 of the MATH-500 dataset, just pulled out first one for the new renderer

tinker_cookbook/recipes/rubric/env.py
derek-tml: yeah this is fine, i explicitly pass in self.grader_renderer_name = qwen3_5_disable_thinking as default anyways.

tinker_cookbook/recipes/code_rl/README.md
derek-tml: 190 is also by default if you use batch size 128. maybe they used 256 before, or just used an intermediate checkpt. either way i think not a huge deal for this one.

tinker_cookbook/recipes/chat_sl/README.md
derek-tml: the script goes to 7000 by default if we use batch_size=128. if they used batch_size=512 before it would be 1740. let me rerun it with that batch size since its quite off

PR #748, opened 2026-06-01
Write rollout traces to summaries

derek-tml:
## Summary
- Adds a `RolloutTrace` dataclass for structured per-step transcript data.
- Carries `StepResult.trace` into `Transition.trace`.
- Writes `steps[].trace` and `steps[].stop_reason` into rollout summary JSONL records.

Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #749
* __->__ #748

Co-authored-by: Cursor <cursoragent@cursor.com>

PR #749, opened 2026-06-01
Add rollout traces to built-in envs

derek-tml:
## Summary
- Adds `RolloutTrace` payloads to built-in rollout envs that can decode prompt/response text.
- Covers problem, prompt-only distillation, rubric, preference, message-env, and multiplayer rollout paths.
- Keeps reward and state details in structured fields where useful.

Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* __->__ #749
* #748

Co-authored-by: Cursor <cursoragent@cursor.com>

### DIFF
--- a/tinker_cookbook/recipes/rubric/env.py
+++ b/tinker_cookbook/recipes/rubric/env.py
@@ -90,7 +90,8 @@ class RubricGradedEnv(Env):
 
     async def step(self, action: Action, *, extra: ActionExtra | None = None) -> StepResult:
         with logtree.scope_header("Prompt"):
-            logtree.log_formatter(ConversationFormatter(messages=self.convo))
+            prompt_formatter = ConversationFormatter(messages=self.convo)
+            logtree.log_formatter(prompt_formatter)
 
         # obtain the policy action message
         (policy_action_message, termination) = self.renderer.parse_response(action)
@@ -114,7 +115,8 @@ class RubricGradedEnv(Env):
             print(colored(f"Termination: {termination}", "green") + "\n")
 
         with logtree.scope_header("Policy Response"):
-            logtree.log_formatter(ConversationFormatter(messages=[policy_action_message]))
+            response_formatter = ConversationFormatter(messages=[policy_action_message])
+            logtree.log_formatter(response_formatter)
             logtree.log_text(f"Termination: {termination}")
 
         convo = self.convo + [policy_action_message]
@@ -126,6 +128,7 @@ class RubricGradedEnv(Env):
 
         with logtree.scope_header("Rubric Grades"):
             rows = []
+            rubric_grades = []
             for idx, (rubric_item, (score, grader_response)) in enumerate(
                 zip(self.rubric_items, results, strict=True),
                 start=1,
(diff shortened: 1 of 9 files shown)


derek-tml:
@codex  review

derek-tml:
@claude review

tinker_cookbook/rl/message_env.py
```diff
@@ -125,6 +147,9 @@ async def step(
                 metrics={"max_tokens_reached": 1.0},
             )
 
+        if self._last_messages is None:
+            self._last_messages = await self.message_env.initial_observation()
+        prompt_messages = deepcopy(self._last_messages)
```
dphuang2, line 152: The pre-step `deepcopy` is the right call since `MessageEnv.step()` mutates history in place. `test_trace_prompt_snapshots_messages_before_step_mutation` only covers the happy path though. The parse-error early return and the context-overflow return both emit a trace too, can we get the snapshot assertion on those paths as well? Those are the ones most likely to regress quietly.

tinker_cookbook/rl/preference_envs.py
```diff
@@ -340,6 +342,18 @@ async def compute_group_rewards(
             matchup_count[i] += 1
         format_coef = 1.0
 
+        prompt_trace = prompt_formatter.to_data()
+        for trajectory, messages, is_valid in safezip(
+            trajectory_group,
+            response_messages,
+            is_valid_list,
+        ):
+            trajectory.transitions[0].trace = RolloutTrace(
```
dphuang2, line 351: This assumes every preference trajectory has exactly one transition, and that `trajectory_group`/`response_messages`/`is_valid_list` stay aligned. `safezip` catches a length mismatch but nothing asserts the right response lands on the right trajectory. Worth a small test, this is the one spot here with real back-fill logic.

tinker_cookbook/recipes/rubric/env.py
```diff
@@ -172,6 +181,12 @@ async def step(self, action: Action, *, extra: ActionExtra | None = None) -> Ste
                 "termination": str(termination),
                 "num_rubrics": len(self.rubric_items),
             },
+            trace=RolloutTrace(
+                prompt=prompt_formatter.to_data(),
+                policy_response=response_formatter.to_data(),
+                reward_data=reward_terms,
```
dphuang2, line 187: Since `reward_terms` feeds both `table_from_dict` and `reward_data`, the trace ends up with display strings (`f"{...:.3f}"`) rather than numbers. For a durable record I think we want raw floats so downstream can compute on them. Cheap fix: keep a numeric dict and format only for the table. Same thing in `problem_env.py` with `reward_table`.

dphuang2 (commented):
Nice, the structured trace is much easier to consume than detokenizing `ob`/`ac`. Few things below, mostly test coverage. Non-blocking except maybe the `message_env` error-path coverage.

Two broader questions on the trace being unconditional (the field lives in #748 but it's this PR that populates it everywhere):

1. It's built on every step and always serialized into the summaries, plus the per-step `deepcopy` in `message_env`, so long multi-turn rollouts pay CPU + a bigger summaries file. Is always-on intended or should it be gated?
2. `log_formatter` already stashes `to_data()` into `*_logtree.json`, so for the `log_formatter` envs the prompt/response is somewhat duplicated. Fine if it's deliberate (summaries as the machine-readable per-traj record vs the HTML logtree), just want to confirm it's a choice.

PR #751, opened 2026-06-01
Update Basics tutorials (1xx) for model deprecations

nealwu:
Note: branched from https://github.com/thinking-machines-lab/tinker-cookbook/pull/732 to split up into smaller PRs.

Refreshes the Basics tutorials (101–104) off models slated for retirement on 2026-06-12, plus a few fixes found while running each notebook. First in a stack of five PRs splitting the tutorial refresh by section (Basics → Core Concepts → Cookbook Abstractions → Advanced → Deployment); this one is based on `main`.

## Tutorials
- **101 Hello Tinker** — switch to `Qwen/Qwen3.5-9B-Base`. A base model is the right fit for the raw `tokenizer.encode → sample` demo (a chat-tuned model drops into thinking mode on raw prompts). Tidy the sampling params.
- **102 First SFT** — `Qwen3-4B-Instruct-2507 → Qwen3.5-4B`; `Kimi-K2.5 → Kimi-K2.6` (with the `kimi_k26_disable_thinking` renderer); drop the deprecated `name=` arg from `save_weights_and_get_sampling_client`.
- **103 Async Patterns** — align the narration with the `asyncio.gather` API the code actually uses (it previously described the synchronous `sample()`/`.result()` futures API).
- **104 First RL** — `meta-llama/Llama-3.1-8B → Qwen/Qwen3.5-9B-Base` with the `role_colon` renderer (the recommended renderer for this base model).

## Tooling
- Add a `[tutorials]` optional-dependencies extra (`marimo`, `matplotlib`) so a plain `pip install tinker-cookbook` stays lean; update `tutorials/README.md` to install via `tinker-cookbook[tutorials]`.

## Testing
Ran the notebooks end-to-end in marimo against the live Tinker service.

_Note: diffs include some marimo serialization churn (the `__generated_with` version bump and trimmed cell-return tuples) from running the notebooks._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #732, opened 2026-05-22
Update tutorials off models slated for retirement on 2026-06-12

nealwu:
Swaps each tutorial's model (and any hardcoded paired renderer) to the recommended replacement from the deprecation page:
- Qwen3-4B-Instruct-2507 -> Qwen3.5-4B (qwen3_5_disable_thinking)
- Llama-3.x base -> Qwen3.5-9B-Base (role_colon)
- Qwen3-30B-A3B -> Qwen3.6-35B-A3B (qwen3_5)
- Qwen3-VL-235B-A22B-Instruct -> Qwen3.5-397B-A17B (vision-capable)
- Kimi-K2.5 -> Kimi-K2.6 (kimi_k26_disable_thinking)

Also refreshes the renderer table and supported-models prose in 201 and 102 so they don't reference retired families.

PR #752, opened 2026-06-01
Update Core Concepts tutorials (2xx) for model deprecations and SDK API drift

nealwu:
Note: branched from https://github.com/thinking-machines-lab/tinker-cookbook/pull/732 to split up into smaller PRs.

Refreshes the Core Concepts tutorials (201–205). Beyond the model swaps, these notebooks had accumulated breakage from SDK changes (results now come back as `TensorData` wrappers, some calls must be `await`ed, etc.) — all found and fixed by running each notebook end-to-end.

Second in the five-PR stack. **Based on #751 (Basics, 1xx) — review/merge that first.** The diff against `main` will be clean once #751 lands.

## Tutorials
- **201 Rendering** — `Qwen3-30B-A3B → Qwen3.6-35B-A3B` and `Qwen3-VL-235B-A22B-Instruct → Qwen3.5-397B-A17B`, using the `qwen3_5` renderers; modernize the vision section (Qwen3.5/3.6 are natively vision-capable — same renderer plus an image processor); rebuild the `parse_response` example (its hardcoded token IDs were tied to the old tokenizer); make the renderer-switching cell actually switch (Qwen + Kimi); note that `qwen3_5` is a thinking renderer.
- **202 Loss Functions** — `Qwen3.5-4B`; unwrap `TensorData` logprobs with `.tolist()` before slicing.
- **203 Completers** — use top-level `await` instead of `asyncio.run` (which can't run inside marimo's event loop); drop now-unused imports.
- **204 Weights** — fix the checkpoint-lifecycle cells: `await` the async RestClient/save APIs (sync `.result()` calls warn and risk deadlock in marimo's loop), pass the full `model_id` to `list_checkpoints` (truncating it caused `404 Model not found`), and run the blocking `weights.download` via `asyncio.to_thread`.
- **205 Evaluations** — unwrap `TensorData` with `.to_torch()` in the NLL evaluator; fix the dead Inspect AI link (`inspect.ai → inspect.aisi.org.uk`).

## Testing
Ran each notebook end-to-end in marimo against the live Tinker service.

**Requires tinker ≥ 0.22.3** — 204's checkpoint download fails on older SDKs due to a server-side bug fixed in 0.22.3.

_Note: diffs include marimo serialization churn (version-string bump, trimmed cell-return tuples)._

### DIFF
--- a/tutorials/201_rendering.py
+++ b/tutorials/201_rendering.py
@@ -102,7 +102,7 @@ def _(mo):
     mo.md(r"""
     The output is a `ModelInput` object containing the tokenized chat template. Notice how each message is wrapped in special tokens like `<|im_start|>` and `<|im_end|>`, and the final `<|im_start|>assistant` is left open for the model to fill in.
 
-    Because `qwen3_5` is a thinking renderer, the prompt also ends with an open `<think>` tag that primes the model to reason before answering. If you'd prefer non-thinking mode instead, the `qwen3_5_disable_thinking` variant instead inserts `<think></think>` instead so the model replies directly.
+    Because `qwen3_5` is a thinking renderer, the prompt also ends with an open `<think>` tag that primes the model to reason before answering. If you'd prefer non-thinking mode instead, the `qwen3_5_disable_thinking` variant inserts a closed `<think></think>` so the model replies directly.
     """)
     return
 
@@ -300,7 +300,7 @@ def _(mo):
     mo.md(r"""
     ## Vision inputs with `ImagePart`
 
-    For vision-language models (Qwen3.5 and Qwen3.6 are all vision-capable), message content can include images alongside text. Use `ImagePart` for images and `TextPart` for text within the same message.
+    For vision-language models (Qwen3.5 and Qwen3.6 models are all vision-capable), message content can include images alongside text. Use `ImagePart` for images and `TextPart` for text within the same message.
     """)
     return
 
--- a/tutorials/201_rendering.py
+++ b/tutorials/201_rendering.py
@@ -14,7 +14,7 @@ def _():
 @app.cell(hide_code=True)
 def _(mo):
     mo.md(r"""
-    # Tutorial: Rendering
+    # Tutorial 201: Rendering
 
     Rendering converts a list of messages into a token sequence that a model can consume. While similar to HuggingFace chat templates, Tinker's rendering system handles the full training lifecycle: supervised learning, reinforcement learning, and deployment.
 
--- a/tutorials/201_rendering.py
+++ b/tutorials/201_rendering.py
@@ -1,6 +1,6 @@
 import marimo
 
-__generated_with = "0.21.1"
+__generated_with = "0.23.8"
 app = marimo.App()
 
 
@@ -34,7 +34,7 @@ def _(mo):
     mo.md(r"""
     ## Setup
 
-    We need a tokenizer (to map between text and token IDs) and a renderer (to apply the model's chat format).
+    We need a tokenizer (to map between text and token IDs) and a renderer (to apply the model's chat format). Note for this example that both Qwen3.5 and Qwen3.6 models use the same `qwen3_5` renderer.
     """)
     return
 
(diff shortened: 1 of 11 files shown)


tutorials/204_weights.py
```diff
@@ -242,23 +243,23 @@ def _(mo):
 
 
 @app.cell
-def _(rest_client, run_id):
+async def _(model_id, rest_client):
     # List checkpoints for this training run
-    checkpoints_response = rest_client.list_checkpoints(run_id).result()
+    checkpoints_response = await rest_client.list_checkpoints_async(model_id)
```
derek-tml, line 249: how come we changed `list_checkpoints_async` to take in `model_id` instead of `run_id`? was this intentional?

tutorials/204_weights.py
```diff
@@ -125,35 +125,35 @@ def _(mo):
 @app.cell
 async def _(training_client):
     # Save weights for inference (sampler checkpoint)
-    sampler_result = await training_client.save_weights_for_sampler_async("tutorial-sampler")
+    sampler_result = await training_client.save_weights_for_sampler("tutorial-sampler")
```
derek-tml: we are awaiting a synchronous function here

tutorials/201_rendering.py
```diff
@@ -94,13 +94,15 @@ def _(messages, renderer, tokenizer):
...
-    return (prompt,)
+    return
 
 
 @app.cell(hide_code=True)
 def _(mo):
     mo.md(r"""
     The output is a `ModelInput` object containing the tokenized chat template. Notice how each message is wrapped in special tokens like `<|im_start|>` and `<|im_end|>`, and the final `<|im_start|>assistant` is left open for the model to fill in.
+
+    Because `qwen3_5` is a thinking renderer, the prompt also ends with an open `<think>` tag that primes the model to reason before answering. If you'd prefer non-thinking mode instead, the `qwen3_5_disable_thinking` variant instead inserts `<think></think>` instead so the model replies directly.
```
derek-tml: small nit: grammar
```suggestion
    Because `qwen3_5` is a thinking renderer, the prompt also ends with an open `<think>` tag that primes the model to reason before answering. If you'd prefer non-thinking mode instead, the `qwen3_5_disable_thinking` variant inserts a closed `<think></think>` so the model replies directly.
```

tutorials/201_rendering.py
```diff
@@ -272,29 +279,34 @@ def _(mo):
...
+    return
 
 
 @app.cell(hide_code=True)
 def _(mo):
     mo.md(r"""
     ## Vision inputs with `ImagePart`
 
-    For vision-language models (like Qwen3-VL), message content can include images alongside text. Use `ImagePart` for images and `TextPart` for text within the same message.
+    For vision-language models (Qwen3.5 and Qwen3.6 are all vision-capable), message content can include images alongside text. Use `ImagePart` for images and `TextPart` for text within the same message.
```
derek-tml: small nit: grammar
```suggestion
    For vision-language models (Qwen3.5 and Qwen3.6 are both vision-capable), message content can include images alongside text. Use `ImagePart` for images and `TextPart` for text within the same message.
```

derek-tml (approved):
confirm first comment before merging

tutorials/204_weights.py
nealwu: this was intentional; `list_checkpoints` / `list_checkpoints_async` takes a `model_id` as input ([see here](https://tinker-docs.thinkingmachines.ai/tinker/api-reference/restclient/); it's called `training_run_id` confusingly but the type is `ModelID`). it needs to be the full `<run_id>:train:<seq>` string that `get_info().model_id` returns. previously this was just the bare `run_id` and that resulted in `404 Model not found`

tutorials/204_weights.py
nealwu: this one is kinda confusing but it's correct. it's a synchronous function that returns an `APIFuture` that we need to `await`: https://tinker-docs.thinkingmachines.ai/tinker/api-reference/trainingclient/#save_weights_for_sampler

I can update this to save the future to a variable so that it's easier to understand

tutorials/201_rendering.py
nealwu: thanks for catching!

tutorials/201_rendering.py
nealwu: thanks for catching!

nealwu: changed to "Qwen3.5 and Qwen3.6 models are all vision-capable"

PR #755, opened 2026-06-02
Remove redundant skills field from plugin manifest

jasminepoon:
## Problem

`/plugin install tinker@...` fails with:

```
Failed to install: ... has an invalid manifest file
Validation errors: skills: Invalid input
```

`.claude-plugin/plugin.json` sets `"skills": "skills/"`. Per the [Claude Code plugin spec](https://code.claude.com/docs/en/plugins-reference), the `skills` field is for **additional** skill directories *beyond* the auto-discovered default `skills/`, and when present it must be a `./`-prefixed relative path (e.g. `"./custom/skills/"`).

This repo's skills already live in the default `skills/` directory, so the field is:
- **redundant** — Claude Code auto-discovers `skills/` without it, and
- **malformed** — the bare `"skills/"` value fails manifest validation, blocking install entirely.

## Fix

Remove the `skills` field and rely on default auto-discovery. Verified that the plugin installs and both `/tinker:research` and `/tinker:debug` load once the field is gone.

## Test change

`tests/downstream_compat/test_install_skills.py` previously asserted `data["skills"] == "skills/"` — i.e. it pinned the exact value that breaks real installation. Updated to stop requiring the redundant field and instead guard against it being reintroduced with an invalid (non-`./`-prefixed, or default-pointing) value.

## Verification

- `plugin.json` parses as valid JSON.
- `pytest tests/downstream_compat/test_install_skills.py` → 8 passed.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -9,6 +9,5 @@
   "homepage": "https://thinkingmachines.ai/tinker/",
   "repository": "https://github.com/thinking-machines-lab/tinker-cookbook",
   "license": "MIT",
-  "keywords": ["tinker", "fine-tuning", "rl", "sft", "dpo", "llm"],
-  "skills": "skills/"
+  "keywords": ["tinker", "fine-tuning", "rl", "sft", "dpo", "llm"]
 }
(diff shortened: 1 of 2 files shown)


jasminepoon:
Supersedes https://github.com/thinking-machines-lab/tinker-cookbook/pull/683, which fixes the same root cause by reformatting the value to "./skills/" but leaves the redundant field in place and does not update the downstream-compat test that pins "skills/" — so that PR currently fails CI. This PR removes the field entirely and updates the test, with all checks green.

jacobfulano:
I encountered this issue as well. Thanks for the fix

PR #770, opened 2026-06-06
Update Advanced tutorials (4xx) for model deprecations and SDK API drift

nealwu:
Branched from #732 to split the tutorial refresh into smaller PRs.

Refreshes the Advanced tutorials (401–407). Fourth in the five-PR stack — **based on the 3xx PR (#771); review/merge that first.** Issues found by running each notebook end-to-end.

## Changes
- **401 SL Hyperparameters** — `Qwen3.5-4B`; note that the notebook shows the sweep API + a pre-computed table (it doesn't launch a live sweep, which would run full SFT jobs).
- **402 RL Hyperparameters** — `Qwen3.5-4B`.
- **403 DPO** — `Qwen3.5-4B` (`qwen3_5_disable_thinking`); note explaining the empty `<think></think>` markers in the rendered output.
- **404 Sequence Extension** — clarify that the multi-turn demo uses placeholder token IDs.
- **405 Multi-Agent** — `Qwen3.5-4B` (`qwen3_5_disable_thinking`).
- **406 Prompt Distillation** — `Qwen3.5-4B` (`qwen3_5_disable_thinking`); SDK fixes; evaluate the student on a held-out test set so the result shows generalization, not memorization; reorder the results table
so variable-width multilingual text doesn't break alignment.
- **407 Full RLHF Pipeline** — `Qwen3.5-9B-Base` base; import `TrainOnWhat` from `renderers` (it moved); `await main()` instead of `asyncio.run` in all three stages; add the now-required `recipe_name` to each `Config`; and render the final eval prompt with the model's renderer + its stop sequences (the old raw `tokenizer.encode` + `stop=["\n\n"]` made both models stop immediately and print nothing).

Ran in marimo against the live Tinker service. Diffs include marimo churn (version-string bump, trimmed cell-return tuples, imports left unused after the edits).

### DIFF
--- a/tutorials/403_dpo_preferences.py
+++ b/tutorials/403_dpo_preferences.py
@@ -1,6 +1,6 @@
 import marimo
 
-__generated_with = "0.21.1"
+__generated_with = "0.23.8"
 app = marimo.App()
 
 
@@ -27,21 +27,11 @@ def _(mo):
 def _():
     from tinker_cookbook.preference.types import (
         Comparison,
-        ComparisonRenderer,
         ComparisonRendererFromChatRenderer,
         LabeledComparison,
-        PreferenceModel,
-        PreferenceModelFromChatRenderer,
     )
 
-    return (
-        Comparison,
-        ComparisonRenderer,
-        ComparisonRendererFromChatRenderer,
-        LabeledComparison,
-        PreferenceModel,
-        PreferenceModelFromChatRenderer,
-    )
+    return Comparison, ComparisonRendererFromChatRenderer, LabeledComparison
 
 
 @app.cell(hide_code=True)
@@ -82,8 +72,9 @@ def _(Comparison, LabeledComparison):
     swapped = labeled.swap()
     print("\nAfter swap:")
     print(f"Completion A: {swapped.comparison.completion_A[0]['content']}")
+    print(f"Completion B: {swapped.comparison.completion_B[0]['content']}")
     print(f"Preferred:    {swapped.label}")
-    return (comparison, labeled, swapped)
+    return (comparison,)
 
 
 @app.cell(hide_code=True)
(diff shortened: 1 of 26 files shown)


tutorials/406_prompt_distillation.py
```diff
@@ -61,42 +61,42 @@ def _(mo):
...
-
-4) When in doubt, choose 'ot' rather than guessing.
-
-Output format:
-- Respond with EXACTLY one line: "Final Answer: xx"
-- Where xx is one of {ar, de, el, en, es, fr, hi, ru, tr, ur, vi, zh, ot} and nothing else.
-
-Text to classify:
-{text}"""
+    Goal: Classify the language of the provided text into exactly one of these labels:
```
derek-tml, line 67: nit: the indentation here matters, i think the old one was more correct.

tutorials/403_dpo_preferences.py
```diff
@@ -14,7 +14,7 @@ def _():
 @app.cell(hide_code=True)
 def _(mo):
     mo.md(r"""
-    # Tutorial 11: DPO and Preference Learning
+    # Tutorial 403: DPO and Preference Learning
 
     Build preference data, train with DPO, and evaluate with a PreferenceModel.
```
derek-tml: looks like we removed PreferenceModel

derek-tml (approved):

tutorials/406_prompt_distillation.py
nealwu: this is tricky because marimo did this automatically. even if we fix it, users running the marimo notebook will end up with the indented version of the prompt

tutorials/403_dpo_preferences.py
nealwu: thanks, updated this description to avoid overclaiming

tutorials/406_prompt_distillation.py
nealwu: updating to use `textwrap.dedent()` which is robust even with marimo automatically adding indentation

PR #771, opened 2026-06-06
Update Cookbook Abstractions tutorials (3xx) for model deprecations, SDK drift, and a GRPO grouping fix

nealwu:
Branched from #732 to split the tutorial refresh into smaller PRs.

Refreshes the Cookbook Abstractions tutorials (301–304). Third in the five-PR stack — **based on #752 (2xx); review/merge that first.** Issues found by running each notebook end-to-end.

## Changes
- **301** — renumber cross-references; clarify why the hand-encoded example scores reward 0.9 not 1.0 (missing stop token), and why `_remove_mask` drops the mask key.
- **302** — renumber cross-references; drop the deprecated `name=` arg from `save_weights_and_get_sampling_client_async()`; comment on the deliberate thinking renderer (lenient `FormatEnv` → constant rewards →
no GRPO signal otherwise).
- **303** — `Qwen3-4B-Instruct-2507 → Qwen3.5-4B` (`qwen3_5_disable_thinking`); build the datum via `datum_from_model_input_weights` (old code called a removed API and omitted `target_tokens`); add required
`recipe_name`; `await` instead of `asyncio.run`.
- **304 RL with Config**
  - **GRPO grouping fix (correctness).** `ArithmeticDataset.get_batch` passed `env_thunk=self._make_env`, and `_make_env` drew a *fresh random problem on every call*. `ProblemGroupBuilder` invokes the thunk `group_size` times, so each "group" was `group_size` **different** problems. GRPO centers advantages within a group (`advantage = reward − mean(group)`), which is only meaningful when the group is the *same* problem sampled `group_size` times — otherwise the baseline averages unrelated problems and the advantage reflects problem difficulty rather than completion quality. Fixed to sample one problem per group and bind it with `functools.partial`, matching tutorial 301 and the production `math_rl` recipe.
  - Model/renderer swap (`Qwen3.5-4B`, `qwen3_5_disable_thinking`); widen the operand range (1–50 → 1–300) so the task isn't trivially easy; fix the metric names in the "inspect reward curves" section (`reward/mean`/`reward/correct` → `env/all/reward/total`/`env/all/correct`).
- **Tooling** — require `tinker>=0.22.3`; add `tinker_cookbook[math-rl]` to the `[tutorials]` extra (301 imports `MathEnv`).

Ran end-to-end in marimo against the live Tinker service. Diffs include marimo churn (version-string bump, trimmed cell-return tuples).

### DIFF
--- a/tutorials/304_rl_with_config.py
+++ b/tutorials/304_rl_with_config.py
@@ -1,6 +1,6 @@
 import marimo
 
-__generated_with = "0.21.1"
+__generated_with = "0.23.8"
 app = marimo.App()
 
 
@@ -29,8 +29,6 @@ def _(mo):
 
 @app.cell
 def _():
-    import asyncio
-    import re
     from collections.abc import Sequence
 
     import chz
@@ -47,10 +45,8 @@ def _():
         RLDataset,
         RLDatasetBuilder,
         Sequence,
-        asyncio,
         chz,
         get_tokenizer,
-        re,
         renderers,
     )
 
@@ -66,7 +62,7 @@ def _(mo):
 
 
 @app.cell
-def _(ProblemEnv, renderers):
+def _(ProblemEnv):
     import random
 
     class ArithmeticEnv(ProblemEnv):
@@ -92,7 +88,7 @@ def _(ProblemEnv, renderers):
         def get_reference_answer(self):
             return self.answer
 
-    return (ArithmeticEnv, random)
+    return ArithmeticEnv, random
 
 
 @app.cell(hide_code=True)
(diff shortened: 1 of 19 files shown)


tutorials/304_rl_with_config.py
```diff
@@ -249,7 +249,7 @@ def _(mo):
 def _():
     from pathlib import Path
 
-    log_dir = Path("~/logs/tutorial-rl-config").expanduser()
+    log_dir = Path("/tmp/tinker-tutorials/rl-config").expanduser()
```
derek-tml: `.expanduser()` is only useful if Path starts with `~`, which is no longer the case, so you could get rid of `.expanduser()`.

derek-tml (approved):
awesome

tutorials/304_rl_with_config.py
nealwu: nice

PR #780, opened 2026-06-19
Raise inspect/Tinker bridge default max_tokens from 128 to 2048

Butanium:
## Problem

`InspectAPIFromTinkerSampling` (in `tinker_cookbook/eval/inspect_utils.py`) bridges a Tinker sampling client into `inspect_ai` as a target `Model`. When an eval doesn't explicitly set `max_tokens`, inspect passes `generate()` a `GenerateConfig` with `max_tokens=None`. The bridge then did:

```python
max_tokens=config.max_tokens or 128,
```

so `None or 128` silently capped **every** target response at 128 tokens. In multi-turn / conversational evals this truncates target responses mid-sentence, and any judge/scorer downstream ends up grading partial answers.

Repro: point a Tinker checkpoint at a multi-turn inspect eval without setting `max_tokens` on the model/eval — target responses are all cut off at exactly 128 tokens.

## Fix

`tinker.SamplingParams` requires an integer `max_tokens`, so a numeric fallback is unavoidable here — we can't omit the field the way the OpenAI/Google inspect providers do. But 128 is far too low.

This raises the fallback to **2048**, matching inspect's own shared default — [`DEFAULT_MAX_TOKENS = 2048`](https://github.com/UKGovernmentBEIS/inspect_ai/blob/c15905788/src/inspect_ai/_util/constants.py#L11) (`inspect_ai._util.constants`), the fallback inspect injects for its other providers (groq, together, hf, bedrock, etc.). A Tinker target now behaves like every other inspect target by default instead of silently truncating.

The value is pulled out into a module-level `DEFAULT_MAX_TOKENS` constant with a comment explaining why the fallback exists and where the number comes from.

## Diff

One line changed at the call site plus a named constant; nothing else touched.

Co-Authored-By: Claude <noreply@anthropic.com>


wuisabel-gif:
I reproduced the reasoning: config.max_tokens or 128 means any inspect eval that doesn't explicitly set max_tokens (it arrives as None) gets silently capped at 128 tokens, which truncates longer/multi-turn target responses so scorers end up grading partial answers. The fix looks right to me, and I confirmed the constant it mirrors: from inspect_ai._util.constants import DEFAULT_MAX_TOKENS gives 2048, so this keeps the Tinker bridge consistent with inspect's other providers. It's also a no-op when max_tokens is set explicitly. Would be great to see this merged.

dphuang2 (approved):

tinker_cookbook/eval/inspect_utils.py
```diff
@@ -36,6 +36,13 @@
 
 logger = logging.getLogger(__name__)
 
+# Fallback when an inspect eval doesn't set `max_tokens` (it arrives as None).
+# tinker.SamplingParams requires an integer, so we can't omit it the way the
+# OpenAI/Google inspect providers do; we mirror inspect's own default
+# (inspect_ai._util.constants.DEFAULT_MAX_TOKENS) to stay consistent with every
+# other inspect target instead of silently truncating generations.
+DEFAULT_MAX_TOKENS = 2048
```
dphuang2, line 44: is it worth pulling that constant in by importing it instead of hard-coding it here?

Butanium: yeah that makes sense let me change to that

Butanium: I mean the tradeoff is that it's a `_util` private module, so in thoery they could just change it without notice. Lmk what you prefer

PR #794, opened 2026-06-30
Modal Inference Integration

harmya:
Add support for serving Tinker fine-tunes on Modal

Takes a tinker:// LoRA checkpoint to a live OpenAI-compatible endpoint in two steps: prepare merges the checkpoint into the base model and writes it to a Volume, serve runs it on SGLang behind a @app.server endpoint that scales to zero

Lives in tinker_cookbook/inference/modal/ (common.py, prepare.py, serve.py, README.md).

```bash
modal run    -m tinker_cookbook.inference.modal.prepare --tinker-path tinker://… --base-model Qwen/Qwen3-8B --name my-ft
modal deploy -m tinker_cookbook.inference.modal.serve   # FINETUNE=my-ft MODEL=Qwen/Qwen3-8B
```

Merge-only for now and it runs on a GPU so it's quick

Tested by comparing the Modal endpoint against the Tinker sampling client on the same checkpoint.


tinker_cookbook/inference/modal/common.py
```diff
@@ -0,0 +1,88 @@
...
+# Adding support for a model we have tested
+MODEL_REGISTRY: dict[str, ModelConfig] = {
+    cfg.base_model: cfg
+    for cfg in (
+        ModelConfig("Qwen/Qwen3-8B", gpu="H100:1", tp=1),
+        ModelConfig("Qwen/Qwen3.5-4B", gpu="H100:1", tp=1),
+        ModelConfig("Qwen/Qwen3.6-35B-A3B", gpu="H100:2", tp=2),
+        ModelConfig("nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16", gpu="H100:2", tp=2),
+    )
+}
```
derek-tml, line 39: curious how come only these set of models? is it possible to do larger models as well or no?

tinker_cookbook/inference/modal/serve.py
```diff
@@ -0,0 +1,135 @@
...
+    gpu=CONFIG.gpu,
+    volumes={ARTIFACTS_PATH: artifacts, HF_CACHE_PATH: hf_cache},
+    secrets=[hf_secret],
+    port=PORT,
+    target_concurrency=TARGET_INPUTS,
+    scaledown_window=10 * MINUTES,
+    startup_timeout=20 * MINUTES,
+    exit_grace_period=15,
+    routing_region=ROUTING_REGION,
+    unauthenticated=True,
```
derek-tml: intentional to have no authentication by default? maybe better to require api key as default.

tinker_cookbook/inference/modal/prepare.py
```diff
@@ -0,0 +1,56 @@
...
+    MINUTES,
+    app,
+    artifact_dir,
+    artifacts,
+    hf_cache,
+    prepare_image,
+)
+
+tinker_secret = modal.Secret.from_name("tinker")  # TINKER_API_KEY
+hf_secret = modal.Secret.from_name("huggingface")  # HF_TOKEN, for gated bases
```
derek-tml: is this always required? what happens if this secret is not set?

tinker_cookbook/inference/modal/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+```bash
+FINETUNE=my-finetune MODEL=Qwen/Qwen3-8B modal deploy -m tinker_cookbook.inference.modal.serve
+```
+
+`modal deploy` gives a persistent URL; `modal run` spins up a replica and runs a
+smoke test. Hit it with any OpenAI client:
+
+```python
+from openai import OpenAI
```
derek-tml: might require openai under pyproject.toml tinker-cookbook[modal]

tinker_cookbook/inference/modal/README.md
```diff
@@ -0,0 +1,66 @@
...
+client = OpenAI(base_url="https://<...>.modal.direct/v1", api_key="x")
+client.chat.completions.create(
+    model="my-finetune",
+    messages=[{"role": "user", "content": "hello"}],
+)
+```
+
+## Tested models
+
+Parity is mean per-token `|Δlogprob|` between Tinker's sampling client and the
```
derek-tml: can you give more detail on how this logprob is calculated? this number might be adding more confusion for users. im not sure we should include this either since it can change over time.

derek-tml (changes requested):

tinker_cookbook/inference/modal/common.py
harmya: I can test larger models! Will update as test more and add them. Mostly these models as they are easy to iterate on

tinker_cookbook/inference/modal/serve.py
harmya: This fails if the user hasn't created a proxy token associated with their modal account. To keep the scope simple for this PR, I kept it unauthenticated. But on second thought, I will think the easiest path for the user to set this up and add instructions

tinker_cookbook/inference/modal/prepare.py
harmya: Not always required but helpful for two reasons:
- unauthenticated downloads are slow
- user wants to merge the LoRA on some private model

I will verify the failure mode. But for this to be a noop, I will add "create_if_missing" so it doesn't fail outright

tinker_cookbook/inference/modal/README.md
harmya: Will resolve

tinker_cookbook/inference/modal/README.md
harmya: These are per-token log-probabilities each assigns to the tokens they agree on, I think its best to remove these outright. Will do that.

tinker_cookbook/inference/modal/prepare.py
derek-tml: i think tinker cookbook in most places assumes no huggingface token, which is why i ask

harmya: Cool! Will make it so that its optional and make it clear

tinker_cookbook/inference/modal/README.md
kunal-patil-glean: Just throwing in my two cents here but it would be great to know the behavior difference between the Tinker sampling client and the Modal deployment. If not through logprobs/KLD then maybe through a script where you can pass a json with some requests, it samples from both, and gets you % of requests they disagree on outputs

derek-tml (approved):

Issue #764, opened 2026-06-04
Tinker skill installation error


kevinfeng-hs:
When installing the Tinker skill for Claude Code, I get the following error:
<img width="1084" height="609" alt="Image" src="https://github.com/user-attachments/assets/9484f28c-8b1d-434d-aa37-27236e6595d0" />
Installation was last attempted June 4th 2026 using Claude Code v2.1.162.
Perhaps @derek-tml might know what's going on?

confiscate:
here is a 1-line fix, please take a look @dphuang2 ! Thank you very much.
https://github.com/thinking-machines-lab/tinker-cookbook/pull/795

confiscate:
this should be fixed @kevinfeng-hs , can  you check and close the ticket?

kevinfeng-hs:
Confirming that it's fixed, thanks!

PR #795, opened 2026-06-30
Fix installation error for Tinker skill for Claude Code

confiscate:
fix for [tinker-cookbook issue #764]( https://github.com/thinking-machines-lab/tinker-cookbook/issues/764)

in the tinker-cookbook repo, this file:
```bash
.claude-plugin/plugin.json
```
had this field value
```javascript
"skills": "skills/"
```
which was rejected by Claude Code's manifest validator with error
```javascript
skills: Invalid input
```
The field specified a relative path without the required `./` prefix.

The fix was simply removing the `"skills": "skills/"` field.
Claude Code already scans `skills/` by default, so the explicit path was both redundant and malformed.


dphuang2 (approved):

dphuang2:
thanks!

Issue #796, opened 2026-07-01
Could not reproduce on policy distillation on DeepMath


1stprinciple:
We tried to use exactly the same command as https://github.com/thinking-machines-lab/tinker-cookbook/blob/main/tinker_cookbook/recipes/distillation/README.md

python -m tinker_cookbook.recipes.distillation.on_policy_distillation \
    model_name=Qwen/Qwen3.5-9B-Base \
    teacher_model=Qwen/Qwen3.5-9B \
    load_checkpoint_path=tinker://b89f6676-49c4-53ed-9788-2e0c81a9eabf:train:0/weights/final \
    dataset=deepmath \
    learning_rate=1e-4 \
    groups_per_batch=512 \
    lora_rank=128 \
    max_tokens=16384 \
    wandb_project=cookbook_distillation

However, within the first 20 steps, the curve fluctuates.

Also the AIME 24 benchmark becomes worse than the SFT model. 
Qwen3.5-9B DeepMath OPD step 20  scores 43.3%, with lots of truncations (17 out of 30)

The checkpoint I used for evaluation: tinker://0e964ae8-63af-5887-b4d2-afe9a717c982:train:0/sampler_weights/000020
Could you take a look? 

Thank you very much!

<img width="1554" height="906" alt="Image" src="https://github.com/user-attachments/assets/2344a8ab-38f2-4501-ac68-4abf722afd48" />

confiscate:
the root cause is probably because there is a mismatch in your student/teacher renderers.

student (`Qwen3.5-9B-Base`) = `role_colon`  renderer
```
User: What is 2+2?

Assistant: The answer is 4.

User:
```

teacher (`Qwen3.5-9B`) = `qwen3_5` renderer
```
<|im_start|>user
What is 2+2?<|im_end|>
<|im_start|>assistant
<think>

</think>

The answer is 4.<|im_end|>
```

teacher then computes logprobs on `role_colon`-formatted tokens instead of its native `qwen3_5` format.
this means the KL target is biased, and the teacher's logprobs don't reflect its true distribution, they reflect its confused response to seeing wrong formatting. The student then tries to match this distorted teacher.

I don't have access to your checkpoint but I did try to reproduce your run, with `groups_per_batch=16` instead of your 512 to save on costs 😄 

<img width="1784" height="1035" alt="Image" src="https://github.com/user-attachments/assets/dbae428e-c968-4047-9c4b-8acded0de01a" />

[...]

PR #797, opened 2026-07-04
Account for renderer mismatches in distillation recipe

confiscate:
Fix for issue https://github.com/thinking-machines-lab/tinker-cookbook/issues/796

#### 1. on_policy_distillation.py

- Added teacher_renderer_name: str | None = None parameter to CLIConfig
- Defaults to the teacher model's recommended renderer (e.g., qwen3_5 for Qwen3.5-9B)
- Passed through to the training config

#### 2. train_on_policy.py

- Added teacher_renderer and tokenizer parameters to incorporate_kl_penalty
- When teacher_renderer is provided, the prompt is reformatted using the teacher's native renderer before computing teacher logprobs
- Tracks teacher prompt length to correctly slice teacher logprobs for alignment with student's response tokens

#### 3. README.md

- Updated example commands to include teacher_renderer_name=qwen3_5
- Added "Renderer mismatch" section explaining the issue and fix

## How it works

Before (biased):
```
Student generates: "User: ...\n\nAssistant: 4.\n\nUser:"
Teacher sees:      "User: ...\n\nAssistant: 4.\n\nUser:"  ← wrong format!
Teacher logprobs:  biased (+0.287 nats/token)
```

After (fixed):
```
Student generates: "User: ...\n\nAssistant: 4.\n\nUser:"
Teacher sees:      "<|im_start|>user\n...<|im_end|>
```

### DIFF
--- a/tinker_cookbook/distillation/train_on_policy.py
+++ b/tinker_cookbook/distillation/train_on_policy.py
@@ -19,7 +19,7 @@ import tinker
 import torch
 from tinker.types import LossFnType
 
-from tinker_cookbook import checkpoint_utils, model_info
+from tinker_cookbook import checkpoint_utils, model_info, renderers
 from tinker_cookbook.display import colorize_example
 from tinker_cookbook.distillation.datasets import (
     CompositeDataset,
@@ -57,6 +57,8 @@ async def incorporate_kl_penalty(
     dataset_indices_D: list[int],
     kl_penalty_coef: float,
     kl_discount_factor: float,
+    teacher_renderer: renderers.Renderer | None = None,
+    tokenizer: Tokenizer | None = None,
 ) -> dict[str, float]:
     """
     Compute reverse KL between the student (log p) and the teacher model (log q), computed as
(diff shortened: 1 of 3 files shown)


Abhishek21g:
Looks like the test suite itself is passing; the blockers are formatting + pyright. From the failing jobs:

- `pre-commit`: `ruff format` wants to reformat `tinker_cookbook/distillation/train_on_policy.py`
- `pyright`: `hasattr(chunk, "tokens")` does not narrow the chunk union enough, so pyright still sees `ImageChunk` / `ImageAssetPointerChunk` in the comprehensions at lines 86, 102, and 113
- `pyright`: `prompt_text` appears to be inferred as a non-string/list-ish value at line 93, so `.split(...)` is rejected
- `pyright`: the inline `messages` list passed to `teacher_renderer.build_generation_prompt(...)` needs to be typed as the cookbook `Message` shape, not inferred as `list[dict[str, ...]]`

A likely path is to run ruff format, then use the cookbook typed chunk/message helpers (or an explicit helper that extracts only encoded text token chunks) so pyright gets a real narrowed type rather than `hasattr` narrowing.

confiscate:
formatting fixed!

derek-tml:
hey, thanks for your pr! if you take a look at `resolve_renderer_name_from_checkpoint_or_default_async`, it prefers:
1. explicit renderer_name
2. renderer_name stored in the checkpoint metadata
3. recommended renderer for model_name

since in `on_policy_distillation` we are kicking it off with:

```
python -m tinker_cookbook.recipes.distillation.on_policy_distillation \
    model_name=Qwen/Qwen3.5-9B-Base \
    teacher_model=Qwen/Qwen3.5-9B \
    load_checkpoint_path=tinker://b89f6676-49c4-53ed-9788-2e0c81a9eabf:train:0/weights/final \
    dataset=deepmath \
    learning_rate=1e-4 \
    groups_per_batch=512 \
    lora_rank=128 \
    max_tokens=16384 \
    wandb_project=cookbook_distillation
```

it uses the loaded checkpoint, which has the proper `qwen3_5` renderer. hope this answers the question.

confiscate:
hey @derek-tml thank you for taking the time to look at this PR, especially on July 4th! 🇺🇸 

what you said makes sense.

I tried to retrace your steps by running locally `resolve_renderer_name_from_checkpoint_or_default_async()`.

It looks like I get a 404 when accessing that checkpoint, leading to the `role_colon` renderer being used as opposed to what you saw (proper `qwen3_5` renderer being used). Maybe this has to do with permissions that only a TML staff has.


## My local test run

I made a small simple script to test `resolve_renderer_name_from_checkpoint_or_default_async()`


```python
import asyncio
from tinker_cookbook import checkpoint_utils

CHECKPOINT = "tinker://b89f6676-49c4-53ed-9788-2e0c81a9eabf:train:0/weights/final"
MODEL_NAME = "Qwen/Qwen3.5-9B-Base"

async def main():
    # resolve_renderer_name_from_checkpoint_or_default_async() prefers
    # 1. explicit renderer_name (set to None below) -> skip to #2
    # 2. renderer_name stored in the checkpoint metadata (set to CHECKPOINT constant)
    # 3. recommended renderer for model_name (fallback if #2 fails)

    renderer = await checkpoint_utils.resolve_renderer_name_from_checkpoint_or_default_async(
        explicit_renderer_name=None,
        load_checkpoint_path=CHECKPOINT,
        model_name=MODEL_NAME,
    )
    print(f"Resolved renderer for {MODEL_NAME} with checkpoint: {renderer}")


asyncio.run(main())
```

the script tries to trace the preference chain you mentioned

> if you take a look at resolve_renderer_name_from_checkpoint_or_default_async, it prefers:
> 
> 1. explicit renderer_name
> 2. renderer_name stored in the checkpoint metadata
> 3. recommended renderer for model_name

the result I got was the renderer was still `role colon` format:
```bash 
uv run check_renderer.py
Could not fetch renderer metadata for checkpoint tinker://b89f6676-49c4-53ed-9788-2e0c81a9eabf:train:0/weights/final: Error code: 404 - {'detail': 'Model not found.'}
Resolved renderer for Qwen/Qwen3.5-9B-Base with checkpoint: role_colon
```
Screenshot 2026-07-05 at 12 39 11 AM

from what I can tell using the script, it looks like the preference chain ran like this

1. explicit_renderer_name is None → skipped 
2. load_checkpoint_path=CHECKPOINT → makes API call → gets 404 → returns None → falls through
3. model_info.get_recommended_renderer_name("Qwen/Qwen3.5-9B-Base") → "role_colon"

so I think the discrepancy of what you saw vs what I saw, was that when we both access the same Checkpoint, I got a 404, but you got a Checkpoint with metadata pointing to use the `qwen3_5` renderer.

- Perhaps this is due to some permissions that only TML staff has? 
- When you run `resolve_renderer_name_from_checkpoint_or_default_async` do you get the `qwen3_5` renderer?
-  Would the original user in https://github.com/thinking-machines-lab/tinker-cookbook/issues/796 get a 404 when accessing the Checkpoint?

Thank you again for taking the time to look at this! Especially during the long weekend! 👍 🙏 🙇‍♂️

confiscate:
hey @derek-tml 👋 
if I run that same small script with a Checkpoint that I own/created, I get the result you mentioned (renderer  `qwen3_5` derived from Checkpoint).

So maybe it was a Checkpoint permission issue ❓ 
In any case, have a nice weekend! 👍 

```bash
henry@Henrys-MacBook-Air:~/Development/tinker-tutorial-test % uv run check_renderer.py
Resolved renderer for empty with checkpoint: qwen3_5
```

<br>
Screenshot 2026-07-05 at 1 51 42 AM

<br><br>


```python
import asyncio
from tinker_cookbook import checkpoint_utils

# CHECKPOINT = "tinker://b89f6676-49c4-53ed-9788-2e0c81a9eabf:train:0/weights/final"
CHECKPOINT = "tinker://32179d76-4b80-54c4-9bef-3439943a41e7:train:0/weights/final"
# MODEL_NAME = "Qwen/Qwen3.5-9B-Base"
# MODEL_NAME = "Qwen/Qwen3.5-9"
MODEL_NAME = "empty"


async def main():
    # resolve_renderer_name_from_checkpoint_or_default_async() prefers
    # 1. explicit renderer_name (set to None below) -> skip to #2
    # 2. renderer_name stored in the checkpoint metadata (set to CHECKPOINT constant)
    # 3. recommended renderer for model_name (fallback if #2 fails)

    renderer = await checkpoint_utils.resolve_renderer_name_from_checkpoint_or_default_async(
        explicit_renderer_name=None,
        load_checkpoint_path=CHECKPOINT,
        model_name=MODEL_NAME,
    )
    print(f"Resolved renderer for {MODEL_NAME} with checkpoint: {renderer}")


asyncio.run(main())
```

confiscate:
marked pull request as draft so it cannot be merged. 
Will close the PR after the thread is completed, as the code changes in the PR seem not to be the real fix anymore

PR #823, opened 2026-07-13
Add bits-per-byte (BPB) alongside NLL for tokenizer-independent comparison

YujiaBao:
## Summary

The chat_sl sweep optimizes `train_mean_nll` (and `test/nll`), which are **per-token** cross-entropy. Per-token NLL is **not comparable across tokenizers**: a coarser tokenizer packs more text into each token (higher nats/token), while a finer one spreads the same text over more tokens (lower nats/token) — even at equal modeling quality. This makes any cross-model comparison in `results/sft_sweep.md` apples-to-oranges.

This PR adds **bits-per-byte (BPB)** as a tokenizer-independent counterpart, logged alongside NLL. BPB divides the total log-loss (in bits) over the trained tokens by the number of UTF-8 bytes of the target text:

```
bpb = -sum(logprob_i for i where weight_i > 0) / (ln(2) * total_target_bytes)
```

The numerator sums the raw per-token NLL over the loss-weighted tokens (weights used only as a `> 0` **mask**, so it stays correct under the SFT default `reduction="mean"`, where per-example weights sum to 1.0 rather than being 0/1). The denominator counts the UTF-8 bytes of those same decoded tokens — fixed by the raw text, not the tokenization. Same span in numerator and denominator ⟹ a proper compression rate that's comparable across tokenizers.

Within a single model, BPB and NLL differ only by a constant, so `argmin` over LR / LoRA rank is unchanged — this only affects cross-tokenizer comparisons.

## Details

- **`supervised/common.py`** — new `compute_bpb(logprobs, weights, target_tokens, tokenizer)` + `_weighted_target_byte_count` (decodes each contiguous run of loss-weighted target tokens, skipping special tokens, and sums UTF-8 bytes). Uses `loss_fn_inputs["target_tokens"]`, already on every SFT datum. Weights are treated as a mask, so BPB is correct regardless of `reduction`.
- **`supervised/nll_evaluator.py`** — `NLLEvaluator` gains an optional `tokenizer`; when supplied it also reports `{name}/bpb`. Backward compatible (defaults to `None`; the DPO caller is unaffected).
- **`supervised/train.py`** — passes the tokenizer to the eval (adds `test/bpb`) and logs `train_mean_bpb` next to `train_mean_nll`.
- **`recipes/sl_loop.py`** — logs `train_mean_bpb` too.
- **`recipes/chat_sl/sweep/analyze.py`** — `RunResult` carries optional `test_bpb`/`train_bpb` (read from W&B + local `metrics.jsonl`); the per-model tables gain `Test BPB`/`Train BPB` columns when present (legacy runs keep the old layout; missing cells render as an em dash), and the doc header explains the NLL-vs-BPB distinction for cross-model reading.
- **`recipes/chat_sl/sweep/README.md`** + **`results/sft_sweep.md`** — document selecting `metric=test/bpb` when comparing across tokenizers, and the cross-model caveat.

Sweeping across models is then just:

```bash
python -m tinker_cookbook.recipes.chat_sl.sweep recipe=sft metric=test/bpb
```

## Test

- `supervised/common_test.py` — `compute_bpb`: hand-computed value, zero-weight (prompt) masking, **weight-magnitude invariance (0/1 vs mean-reduction)**, UTF-8 byte counting, special-token handling, batch aggregation, zero-byte → `nan`, contiguous-run splitting.
- `recipes/chat_sl/sweep/analyze_test.py` — table generation with/without BPB and the mixed-grid em-dash case.
- Verified end-to-end against a real `gpt2` tokenizer: `_weighted_target_byte_count` reconstructs the exact raw-text byte length, and BPB is invariant to mean-reduction weights.
- `uv run --extra dev pytest tinker_cookbook/supervised/common_test.py tinker_cookbook/recipes/chat_sl/sweep/analyze_test.py` → **21 passed**; `ruff check` / `ruff format --check` clean; `pyright` **0 errors**.

https://claude.ai/code/session_019F6rpcmF91w8f8LUnVBcRk

### DIFF
--- a/tinker_cookbook/supervised/common.py
+++ b/tinker_cookbook/supervised/common.py
@@ -107,13 +107,16 @@ def compute_bpb(
     removes this dependence by dividing the *total* log-loss (converted to bits)
     by the number of UTF-8 bytes of the target text::
 
-        bpb = -sum(logprobs * weights) / (ln(2) * total_target_bytes)
+        bpb = -sum(logprob_i for i where weight_i > 0) / (ln(2) * total_target_bytes)
 
-    The numerator is the same weighted sum used by :func:`compute_mean_nll`
-    (only without the per-token averaging); the denominator counts the bytes of
-    the decoded loss-weighted tokens, which is fixed by the raw text rather than
-    by the tokenization. Both cover the same weighted span, so the ratio is a
-    proper compression rate that can be compared across tokenizers.
+    The numerator sums the raw per-token NLL over the loss-weighted tokens,
+    using the weights only as a ``> 0`` mask rather than as multipliers. This
+    keeps the total nats correct even when the weights are normalized per
+    example (``reduction="mean"``, the SFT default, where a datum's weights sum
+    to 1.0 instead of being 0/1). The denominator counts the UTF-8 bytes of
+    those same decoded tokens, which is fixed by the raw text rather than by the
+    tokenization. Because numerator and denominator cover the same span, the
+    ratio is a proper compression rate that can be compared across tokenizers.
 
     Args:
         logprobs_list (list[tinker.TensorData]): Per-token log-probabilities,
@@ -128,7 +131,7 @@ def compute_bpb(
     Returns:
         float: Bits per byte, or ``nan`` if the weighted target has zero bytes.
     """
-    total_weighted_logprobs = 0.0
+    total_nll_nats = 0.0
     total_bytes = 0
 
     for logprobs, weights, target_tokens in zip(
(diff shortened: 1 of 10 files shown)


tinker_cookbook/supervised/common.py
```diff
@@ -45,6 +48,113 @@ def compute_mean_nll(
...
+            total_bytes += _decoded_byte_length(run, tokenizer)
+            run = []
+    if run:
+        total_bytes += _decoded_byte_length(run, tokenizer)
+    return total_bytes
+
+
+def _decoded_byte_length(token_ids: list[int], tokenizer: Tokenizer) -> int:
+    """UTF-8 byte length of ``token_ids`` decoded to natural-language text."""
+    text = tokenizer.decode(token_ids, skip_special_tokens=True)
```
derek-tml, line 88: my bot is complaining that numerator and denominator doesn't fully match, could you help confirm?

> This sums loss for every weighted token, but the byte denominator below decodes with skip_special_tokens=True. Since our renderers include end-of-turn tokens like <|im_end|> / <|eot_id|> in the trainable output, BPB currently includes their NLL in the numerator while excluding their bytes from the denominator. That biases BPB, especially for short completions and across different chat templates. Can we make the numerator use the same non-special-token span as the denominator?

derek-tml (approved):

tinker_cookbook/supervised/common.py
YujiaBao: thanks! check this followup pr https://github.com/thinking-machines-lab/tinker-cookbook/pull/824

PR #826, opened 2026-07-13
Filter method-only special tokens in BPB

derek-tml:
## Summary

- Make `compute_bpb` filter tokenizer special/control tokens exposed through `is_special_token(token_id)`, not only `all_special_ids`.
- Keep the BPB numerator and byte denominator on the same filtered target span.
- Add a regression test for tokenizers that expose special IDs by method only.

## Validation

- `uv run --with pytest python -m pytest tinker_cookbook/supervised/common_test.py`
- `uv run --with ruff ruff check tinker_cookbook/supervised/common.py tinker_cookbook/supervised/common_test.py`
- `uv run --with ruff ruff format --check tinker_cookbook/supervised/common.py tinker_cookbook/supervised/common_test.py`

### DIFF
--- a/tinker_cookbook/supervised/common.py
+++ b/tinker_cookbook/supervised/common.py
@@ -91,6 +91,22 @@ def _decoded_byte_length(token_ids: list[int], tokenizer: Tokenizer) -> int:
     return len(text.encode("utf-8"))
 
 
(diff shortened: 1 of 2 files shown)


YujiaBao:
@derek-tml actually can you check https://github.com/thinking-machines-lab/tinker-cookbook/pull/825

YujiaBao (approved):

PR #829, opened 2026-07-15
sample_reasoning: scale default max_tokens with effort

dphuang2:
Running `sample_reasoning.py` with its defaults, the `xhigh` (0.99) sample generates past the fixed `max_tokens=8192` and terminates `malformed` with no boxed answer, so the script's own default sweep produces a truncated comparison point. This is the documented higher-effort-needs-a-larger-budget behavior, but the script should account for it.

This scales the default generation budget with effort (4096 below 0.3, 8192 up to 0.9, 16384 at 0.9 and above). `max_tokens=...` now pins a single budget for every effort instead of being the default. The output also prints the budget next to the generated-token count and warns explicitly when a response hits it.

Verified live against prod: `xhigh` on the default AIME prompt now completes at 11,352 tokens with `stop_sequence` termination (previously truncated at 8192), and an explicit `max_tokens=64` override pins the budget and prints the truncation warning.

### DIFF
--- a/tinker_cookbook/scripts/inkling/sample_reasoning.py
+++ b/tinker_cookbook/scripts/inkling/sample_reasoning.py
@@ -7,7 +7,9 @@ problem 26, for which it also prints the expected answer.
 
     uv run python -m tinker_cookbook.scripts.inkling.sample_reasoning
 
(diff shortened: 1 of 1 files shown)


dphuang2:
**[AI-generated]**

@codex review

PR #838, opened 2026-07-23
Add preserve-thinking Nemotron-3 renderers (HF truncate_history_thinking=False)

### LINKED ISSUE #837: Mismatch between nemotron renderer and hf template for previous turn thinking
the tinker renderer for nemotron drops previous thinking turn, and unlike the hf template doesn't expose an option to keep them.

Butanium:
Fixes #837.

## Problem

The Nemotron-3 HF chat template allow to NOT truncate the thinking from previous turns behind the `truncate_history_thinking` flag, but the cookbook renderers hardcoded `truncate_history_thinking=True`, so there was no way to keep previous-turn reasoning.

From the HF chat template:
```jinja
{%- set truncate_history_thinking = truncate_history_thinking if truncate_history_thinking is defined else True %}
...
{%- if not (truncate_history_thinking and loop.index0 < ns.last_user_idx) %}
    {{- '<|im_start|>assistant\n' ~ (content | ... | trim) ~ '<|im_end|>\n' }}   {# keep thinking #}
{%- else %}
    ... {%- set c = "<think></think>" ~ c.split('</think>')[-1] %} ...            {# collapse thinking #}
```

With the default `truncate_history_thinking=True`, each historical assistant turn's `<think>…</think>` block collapses to `<think></think>`. Setting it `False` keeps the full block. All three Nemotron-3 templates (Nano, Super, Ultra) declare this flag identically.
(small note: Qwen3.5's template has no equivalent knob (its history thinking is unconditionally dropped), so no Qwen change is needed.)

## Change

- Added **`nemotron3_preserve_thinking`** (Nano/Super) and **`nemotron3_ultra_preserve_thinking`** (Ultra): thin renderer variants that forward `strip_thinking_from_history=False`. This mirrors the existing `kimi_k26_preserve_thinking` precedent; HF's `truncate_history_thinking=False` is byte-equivalent to the cookbook's `strip_thinking_from_history=False`.
- The base machinery already skips stripping when `strip_thinking_from_history=False`, but Nemotron's `_assistant_header_suffix` injected `<think></think>` for historical messages **unconditionally**, which would double-render preserved thinking (`<think></think>\n<think>real reasoning</think>`). Gated that injection on the flag — a one-condition change (`not is_historical` → `not is_historical or not self.strip_thinking_from_history`) on both `Nemotron3Renderer` and `Nemotron3UltraRenderer`. The default (strip) path is untouched.
- Registered both names in the renderer factory and in `model_info` recommended-renderer lists.
- These variants report `has_extension_property = True` (inherited), unlike the default renderers.

## Tests

`nemotron3_test.py` compares the preserve renderers against `apply_chat_template(..., truncate_history_thinking=False)` and asserts **token-id equality** (Nano + Ultra, generation + supervised, plus a historical thinking+text+tool_calls turn). The `_hf_*_tokens` helpers now thread `truncate_history_thinking` through, defaulting to `True` so every pre-existing test is unchanged. Added sanity-contrast tests (default collapses historical thinking, preserve keeps it verbatim), renderer-type, extension-property, and `model_info` registration tests. Full `nemotron3_test.py` + `model_info_test.py` pass (70).

## Notes for review

- Only added preserve variants for the full-thinking modes (base + Ultra), where preserving reasoning is meaningful; the disable/low/medium renderers can still opt in via the constructor kwarg:
  ```python
  # reasoning-off mode that still keeps previous-turn <think> blocks
  renderer = Nemotron3DisableThinkingRenderer(tokenizer, strip_thinking_from_history=False)
  ```
  (verified byte-identical to `apply_chat_template(..., enable_thinking=False, truncate_history_thinking=False)`; same holds for `Nemotron3LowThinkingRenderer` / `Nemotron3UltraMediumThinkingRenderer` with their mode flags.)

🤖 Drafted by Claude (Opus 4.8) in Claude Code, reviewed and edited by @Butanium.

Co-Authored-By: Claude <noreply@anthropic.com>

https://claude.ai/code/session_01HTuq4hYbEseWwRa38npcjg

### DIFF
--- a/tinker_cookbook/model_info.py
+++ b/tinker_cookbook/model_info.py
@@ -27,12 +27,13 @@ _GPT_OSS = ("gpt_oss_no_sysprompt", "gpt_oss_medium_reasoning")
 _KIMI_K2 = ("kimi_k2",)
 _KIMI_K25 = ("kimi_k25", "kimi_k25_disable_thinking")
 _KIMI_K26 = ("kimi_k26", "kimi_k26_disable_thinking", "kimi_k26_preserve_thinking")
-_NEMOTRON3 = ("nemotron3", "nemotron3_disable_thinking")
+_NEMOTRON3 = ("nemotron3", "nemotron3_disable_thinking", "nemotron3_preserve_thinking")
 _NEMOTRON3_SUPER = _NEMOTRON3 + ("nemotron3_low_thinking",)
 _NEMOTRON3_ULTRA = (
     "nemotron3_ultra",
     "nemotron3_ultra_disable_thinking",
     "nemotron3_ultra_medium_thinking",
+    "nemotron3_ultra_preserve_thinking",
 )
 _TML_V0 = ("tml_v0",)
 
--- a/tinker_cookbook/model_info_test.py
+++ b/tinker_cookbook/model_info_test.py
@@ -56,6 +56,7 @@ class TestNemotron3:
             "nemotron3_ultra",
             "nemotron3_ultra_disable_thinking",
             "nemotron3_ultra_medium_thinking",
+            "nemotron3_ultra_preserve_thinking",
         ]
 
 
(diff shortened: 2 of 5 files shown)


Butanium:
cc @dphuang2

dphuang2 (approved):
thanks!

PR #840, opened 2026-07-24
Search recipe: make the worked example follow the 1-5 word answer rule

### LINKED ISSUE #835: Search recipe: worked example answer contradicts the "one to five words" instruction
In `tinker_cookbook/recipes/search_tool/search_env.py`, the task instructions tell the model:

> 5. Include your final answer after the "Answer:" prefix. The answer should be between one to five words.

But the worked example in the same prompt ends with a full-sentence answer:

> 4. Answer: The population of New York City grew the most in 2024, and the population of San Francisco changed by XXXX in 2024.

So the single in-context example demonstrates the opposite of the stated rule, which likely works against the "1–5 words" instruction during rollouts.

Note the example question itself is compound — "which year did NYC see the most growth **and** how did SF population change in that year" — so it inherently can't be answered in 1–5 words. Fixing just the `Answer:` line isn't enough; the example and the instruction need to be brought into agreement.

Two possible directions:
1. Swap in a simple, single-part example whose answer genuinely fits in 1–5 words.
2. Relax the "1–5 words" instruction to allow a short sentence, matching the current example.

Happy to send a PR once there's a preferred direction.

RupeshKeesaram:
Fixes #835 (nice catch by @wuisabel-gif).

The worked example in `SEARCH_TASK_INSTRUCTIONS` ended with a full-sentence answer, contradicting instruction 5 ("between one to five words"). Since the reward in `tools.py` does normalized exact-match against short gold answers, the instruction is the right side to keep, so I fixed the example rather than relaxing the rule.

Rather than swapping in a single-hop question, I reworded the compound question into a bridge question: "What was the population of San Francisco in the year that New York City saw its most population growth between 2020 and 2025?" The example still demonstrates the multi-hop flow (find the year first, then use it in the second query), but the final answer is now a single number, consistent with the length rule. Also tightened the step 3 narration so it matches the query it actually issues.

No code changes outside the prompt string; `offline_eval.py` imports the constant so it picks this up automatically.

### DIFF
--- a/tinker_cookbook/recipes/search_tool/search_env.py
+++ b/tinker_cookbook/recipes/search_tool/search_env.py
@@ -31,13 +31,13 @@ Here are instructions for how to solve a problem:
 5. Include your final answer after the "Answer:" prefix. The answer should be between one to five words.
 
 Here is an example of solving a real question:
-"Between 2020 and 2025, which year did New York City see the most population growth and how did San Francisco population change in that year?"
+"What was the population of San Francisco in the year that New York City saw its most population growth between 2020 and 2025?"
 
-1. Think step by step: In order to answer this question, I need to know the population of New York City and San Francisco between 2020 and 2025. I will search for the population of New York City in each year
+1. Think step by step: In order to answer this question, I first need to know which year between 2020 and 2025 New York City saw the most population growth. I will search for the population of New York City in each year.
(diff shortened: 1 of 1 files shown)


nealwu (approved):
Thanks @RupeshKeesaram! Your fix had the right structure (keep the 1–5 word rule and keep the multi-hop example). I added one more commit on top to make things more precise / stable, and I fixed some grammar as well.

PR #846, opened 2026-07-31
Make deprecation deadlines visible in CI

rgbkrk:
CI uses shallow checkouts without tags, so `hatch-vcs` reports the package as `0.1.dev*`. Version-gated deprecations therefore never expire in CI. This hid rollout logging helpers scheduled for removal in v0.4.0 that still ship in v0.5.3.

The way I found this was:

* Installed tinker cookbook as an editable install (for me to use in a separate repo)
* Run the tests from there (which meant the version was 0.5.3)
* Tripwire from the deprecation lines actually trigger at this point

## Summary

- fetch full git history so CI tests against the real package version
- remove the expired rollout logging helpers and port their NumPy scalar coverage to the supported serializer

### DIFF
--- a/.github/workflows/downstream-compat.yaml
+++ b/.github/workflows/downstream-compat.yaml
@@ -13,6 +13,8 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # hatch-vcs needs full history for version
 
       - name: install-uv
         uses: astral-sh/setup-uv@v6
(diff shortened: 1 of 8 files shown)


IgorGanapolsky:
Thanks for the clear root-cause writeup — shallow CI checkouts hiding version-gated deprecations is a classic hatch-vcs footgun.

From agent-harness CI elsewhere: we also force `fetch-depth: 0` (or at least tags) whenever packaging version gates behavior. Worth ensuring `pytest` and smoke-recipe workflows share the same checkout policy so deprecation tripwires cannot diverge between unit and recipe CI.

Not affiliated with TML; reviewing as an external consumer of cookbook releases.

PR #848, opened 2026-07-31
Handle single-token completions when building RL datums

rgbkrk:
A completion can be a single token: the model emits `<|endoftext|>` as its first token. `tokens[:-1]` is then empty, and the server rejects the empty chunk:

```
Empty tokens list in input for field input_sequence[28][7]
```

`forward_backward` takes the whole batch in one call, so one such datum fails the step after it was fully sampled. On base weights this hits roughly 22% of full tutorial runs (math in the comments below).

Fix: append the shifted input chunk only when the completion has more than one token. The lone token still trains via `target_tokens`. Skipping these completions (this PR's first version) was wrong: a lone EOS scores 0 in a non-degenerate group, and its negative advantage is the signal that discourages stopping immediately. Mirrors `create_rightshifted_model_input_and_leftshifted_targets`.

Applied in `tutorials/104_first_rl.py` and `recipes/rl_loop.py`, which had the same construction.

### DIFF
--- a/tinker_cookbook/recipes/rl_loop.py
+++ b/tinker_cookbook/recipes/rl_loop.py
@@ -215,7 +215,11 @@ def main(config: Config):
                 sampled_tokens_G_T, logprobs_G_T, advantages_G
             ):
                 ob_len = prompt.length - 1
-                model_input = prompt.append(types.EncodedTextChunk(tokens=sampled_tokens[:-1]))
+                # A one-token completion has an empty shift; keep it as a target and
+                # skip the empty input chunk (the server rejects empty chunks).
+                model_input = prompt
+                if len(sampled_tokens) > 1:
+                    model_input = prompt.append(types.EncodedTextChunk(tokens=sampled_tokens[:-1]))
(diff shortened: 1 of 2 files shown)


dphuang2:
so the model stopped immediately? thats a bit surprising

do you have the trajectory? in some sense i'd rather people actually run into this error so we learn not to provide users with an example that hits a crazy edge case like this

rgbkrk:
Most likely, yeah. I don't have the raw trajectory anymore but the failure showed that completion contained at most one token. Since `\n\nUser:` is three tokens, a one-token completion here would be `<|endoftext|>`.

On the base weights, `P(EOS first)` is about 0.02% per sample. Across 1,280 samples though that's around a 22% chance of at least one immediate EOS

$$P(\text{at least one})
=1-(1-0.000195)^{1280}
\approx 1-0.7791
\approx 0.2209
$$

rgbkrk:
` fc801711-f4be-5388-b0a4-95ef27599175:train:0:11 ` is the rejected step 3 forward_backward.

PR #849, opened 2026-07-31
Frame the produced Qwen3 turn the way the chat template does

rllin:
## What

`Qwen3Renderer` documents token-identity with `apply_chat_template(enable_thinking=True)`. On the turn being produced, it isn't:

```diff
-<|im_start|>assistant\n4<|im_end|>                                 # no reasoning: no block
+<|im_start|>assistant\n<think>\n\n</think>\n\n4<|im_end|>

-<|im_start|>assistant\n<think>2+2 is 4</think>4<|im_end|>          # reasoning: unpadded
+<|im_start|>assistant\n<think>\n2+2 is 4\n</think>\n\n4<|im_end|>
```

21 tokens against the template's 25 on a one-exchange conversation.

## Why

The chat template (`Qwen/Qwen3-8B`, assistant branch) writes the block in one line:

```jinja
{%- if loop.index0 > ns.last_query_index %}
    {%- if loop.last or (not loop.last and reasoning_content) %}
        {{- '<|im_start|>' + message.role + '\n<think>\n'
            + reasoning_content.strip('\n') + '\n</think>\n\n' + content.lstrip('\n') }}
```

Padded, and present even when `reasoning_content` is empty — so a produced turn that did not reason still opens the block. This renderer wrote `<think>{reasoning}</think>` and nothing at all when there was no reasoning.

Existing HF-parity tests miss it because every one of them ends on a **user** message, so `add_generation_prompt=True` exercises the generation prefix — which this renderer already gets right — and never a produced assistant turn.

Two things follow from that same line, in the second commit:

- **String content is normalized too.** The template splits an inline block out of a string and `lstrip`s the remainder (lines 37-39), so `<think>\n\n</think>\nx` and `<think>\n\n</think>\n\nx` both render as the latter. Passing the string through as written reproduced whichever spelling the caller used.
- **Instruct variants are excluded.** `Qwen3InstructRenderer` and `Qwen3VLInstructRenderer` subclass `Qwen3Renderer` and override only `has_extension_property`, so they inherited the block and were handed a `<think>` they document as never using. `frames_thinking` gates it.

## Test Plan

`test_qwen3_produced_assistant_turn_matches_hf`, parametrized with and without reasoning. Red on `main`, green with the change:

```
2 failed   test_qwen3_produced_assistant_turn_matches_hf[None]
           test_qwen3_produced_assistant_turn_matches_hf[2+2 is 4]
```

`qwen3_test.py` + `qwen3_tool_declaration_test.py`: **79 passed**, all pre-existing tests unmodified.

One trap the test documents: the template reads reasoning from a top-level `reasoning_content` and ignores content parts it doesn't know. Handed `[ThinkingPart, TextPart]` it renders the turn **empty**, so the reference side must be fed the wire shape or the comparison passes for the wrong reason.

## Impact

Rendered tokens move by 4-5 per produced assistant turn for anyone training on one. That is the intent — it is what a public inference stack serves the same model — but it is a behaviour change, not a pure bugfix.

## Not included

`qwen3_5.py`, a separate renderer I haven't measured for the same gap.

### DIFF
--- a/tinker_cookbook/renderers/qwen3.py
+++ b/tinker_cookbook/renderers/qwen3.py
@@ -37,6 +37,17 @@ from tinker_cookbook.renderers.base import (
 from tinker_cookbook.tokenizer_utils import Tokenizer
 
 
+def _frame_thinking(reasoning: str) -> str:
+    """Frame reasoning the way Qwen3's chat template does.
+
+    The template writes `<think>\\n{reasoning}\\n</think>\\n\\n` around the turn being
+    produced -- padded, and present even when the reasoning is empty. Rendering
+    `<think>{reasoning}</think>` instead differs from `apply_chat_template` by 3 tokens
+    with reasoning and 4 without, on the turn the model is about to continue.
+    """
+    return "<think>\n" + reasoning.strip("\n") + "\n</think>\n\n"
+
+
 def _merge_consecutive_text_parts(
     chunks: list[ImagePart | TextPart],
 ) -> list[ImagePart | TextPart]:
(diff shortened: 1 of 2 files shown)


dphuang2 (approved):
oh good catch

PR #856, opened 2026-08-07
feat: optional GSAR groundedness scoring for tau2-bench trajectories

fede-kamel:
## Summary

tau2-bench's grading (`_check_actions`) is action-matching: did the agent call the right tools with the right arguments. It doesn't check whether the agent's *spoken reply to the customer* actually reflects what those tools returned. An agent that calls `get_reservation_details` correctly and then fabricates the cabin class in its summary scores identically to one that reports it correctly.

This adds an optional groundedness check, using [GSAR](https://tulipagents.ai/concepts/gsar/) (typed claim/evidence partitioning — [arXiv:2604.23366](https://arxiv.org/abs/2604.23366)) from [tulipagents.ai](https://tulipagents.ai)'s open-source `tulip-agents` SDK.

## How it works

```mermaid
sequenceDiagram
    participant Env as Tau2MessageEnv
    participant Backend as ToolBackend
    participant Log as trajectories.jsonl
    participant Score as tulip_grounding
    participant Judge as GSAR judge

    Env->>Backend: execute tool_name, args
    Backend-->>Env: real result
    Note over Backend: call_log now includes the result, not just name and arguments
    Env->>Log: write agent_final_message and predicted_actions

    Note over Log,Score: everything above ships unconditionally
    Note over Score,Judge: nothing below runs unless tulip_grounding is imported

    Score->>Log: read trajectories.jsonl
    Score->>Judge: judge final_message against tool evidence
    Judge-->>Score: grounded, ungrounded, contradicted, or complementary
```

Two changes:

1. **`ToolBackend.execute` now logs the real result** alongside each call, and `Tau2MessageEnv._finalize` logs the agent's final message to the customer. Both already flow into `trajectories.jsonl` via the existing `StoredTrajectory.logs` field — no new persistence mechanism, no behavior change for anyone not consuming the two new keys.
2. **New, fully optional `tinker_cookbook/eval/tulip_grounding.py`** — scores a trajectory's final message against its real tool-call evidence via `tulip.reasoning.gsar_judge`. Nothing else in this repository imports it or depends on `tulip-agents`.

## Verified, not asserted

Given the *identical* real tool call and *identical* real tool result, an honest summary and a hallucinated one (fabricated cabin, fabricated baggage count) score **identical `action_score` (1.0)** but diverge under GSAR:

| | action_score | grounding_score | decision |
|---|---|---|---|
| Honest summary | 1.0 | 1.00 | resolved |
| Hallucinated summary (same tool call, same result) | 1.0 | 0.40 | abstain (contradicted claims flagged) |

Full reproduction in `tulip_grounding_test.py`, using tau2-bench's own real `ToolBackend`/`_check_actions` — imported, not reimplemented.

## Honest design note

Deliberately **not** routed through Tinker's own `SamplingClient` — that's a low-level, token-based sampling API with no native structured-output mode, and a reliable JSON-extraction adapter on top of it is unproven. Using any `tulip`-compatible chat model (OpenAI, Anthropic, or an OpenAI-compatible endpoint) is the honest choice for a first version, not a limitation hidden from the reader — same note is in the module's own docstring.

## Verified

- 78 tests pass in `tinker_cookbook/eval/` (4 skipped without a live judge key — real, billed API calls, not part of the default suite for that reason).
- `ruff check` / `ruff format --diff` / `pyright` all clean.
- Live-tested against two independent providers during development (OpenAI, Together).

## Testing plan

```
pip install tulip-agents  # optional, only if you import tulip_grounding
pytest tinker_cookbook/eval/tulip_grounding_test.py -v
pytest tinker_cookbook/eval/ -q   # full regression, no key needed
```

### DIFF
--- a/tinker_cookbook/eval/benchmarks/_tau2_bench.py
+++ b/tinker_cookbook/eval/benchmarks/_tau2_bench.py
@@ -192,8 +192,15 @@ class ToolBackend:
 
     def execute(self, tool_name: str, arguments: dict) -> str:
         """Execute a tool call and return the result as a string."""
-        self.call_log.append({"name": tool_name, "arguments": arguments})
+        result_str = self._execute_and_serialize(tool_name, arguments)
+        # Logged with the result, not just name/arguments -- action-matching
+        # grading (_check_actions) only needs the call, but anything scoring
+        # whether the agent's later claims are *grounded* needs to know what
+        # the tool actually returned.
+        self.call_log.append({"name": tool_name, "arguments": arguments, "result": result_str})
+        return result_str
 
+    def _execute_and_serialize(self, tool_name: str, arguments: dict) -> str:
         if tool_name not in self._tools_by_name:
             logger.warning(
                 "Tool %r not found in schema. Available: %s",
@@ -683,8 +690,12 @@ class Tau2MessageEnv(MessageEnv):
 
         logs["example_id"] = self.example_id
         logs["num_turns"] = self._turn_count
-        logs["predicted_actions"] = json.dumps(all_predicted_calls)[:500]
+        logs["predicted_actions"] = json.dumps(all_predicted_calls)[:2000]
         logs["expected_actions"] = json.dumps(self.expected_actions)[:500]
+        # The agent's final turn to the customer -- what a grounding check
+        # (e.g. does this message actually match what the tools returned,
+        # not just whether the right tools were called) would score.
+        logs["agent_final_message"] = get_text_content(last_message)[:2000]
 
         return MessageStepResult(
             reward=score,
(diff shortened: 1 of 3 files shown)


fede-kamel:
@YujiaBao — would you be open to a design review of this tau2-bench extension before it is considered for merge? You authored the benchmark framework, so I want to validate the boundary rather than presume it.

The narrow claim is not that tau2-bench action scoring is wrong: it correctly scores action behavior. This adds an optional second question for agentic trajectories: does the final customer-facing message faithfully reflect the results of the tools the episode actually executed? The included reproduction keeps tool calls/results identical and changes only the final message, showing why those are distinct signals.

I would especially value your view on three choices:

1. Is retaining the real serialized tool result and final customer message in trajectory logs useful benchmark instrumentation on its own?
2. Is an optional groundedness adapter a good cookbook example, or should the generic logging land separately and model/judge-specific adapters live elsewhere? tulip-agents is deliberately not an install dependency.
3. Is the final customer-facing message the right artifact to score for this benchmark, versus a different trajectory artifact?

One implementation point I have already identified before asking for merge review: the score_trajectories_file helper must score trajectories concurrently rather than awaiting each judge call serially, in line with the cookbook evaluation guidance. I will address that only if the overall direction is welcome. All current repository CI checks pass. Thank you for taking a look.

PR #859, opened 2026-08-08
[renderers] a tool call with nothing before it takes no separator

rllin:
```diff
  {"role": "assistant", "content": "", "tool_calls": [get_weather(city="SF")]}
- <|im_start|>assistant\n\n<tool_call>\n{...}\n</tool_call><|im_end|>
+ <|im_start|>assistant\n<tool_call>\n{...}\n</tool_call><|im_end|>
```

That newline separates the calls from the text before them, and the calls from each other.
A turn that is only a tool call has nothing to separate, and the template writes nothing:

```jinja
{%- if (loop.first and content) or (not loop.first) %}{{- '\n' }}{%- endif %}
```

We wrote it unconditionally. The comment said it matched HF, which is true for the case the
guard exists for -- a turn that also has text -- and wrong for a turn that does not.

`pytest tinker_cookbook/renderers/` -- 532 passed, 46 skipped.

`test_qwen3_tool_call_separator_follows_the_template` covers both sides of the guard, and
against `main` exactly the wrong one is red:

| turn | expects | on `main` |
|---|---|---|
| tool call only | `assistant\n<tool_call>` | **red** |
| text, then tool call | `Let me check.\n<tool_call>` | green |

The second is what stops the newline being dropped altogether.

Found while diffing this renderer against ours in an internal conformance harness that
compares both against the published chat template.

---

**Stack** (base → top): #866 → #864 → #858 → **#859** → #865 → #862 → #867

### DIFF
--- a/tinker_cookbook/renderers/qwen3.py
+++ b/tinker_cookbook/renderers/qwen3.py
@@ -255,6 +255,11 @@ class Qwen3Renderer(Renderer):
             # with ThinkingPart separated from text (as returned by parse_response).
             output_content = content
 
+        # The template's `content` is the message's own text, and the separator below asks
+        # about that -- not about the empty block, which is framing the template writes
+        # itself. Captured before the block is prepended.
+        has_text = bool(output_content)
+
         # The template opens the block on the turn being produced whether or not it
         # reasoned, so a turn that did not reason still gets an empty one -- whatever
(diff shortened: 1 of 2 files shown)


rllin:
Converted to ghstack; this is now #875.

Issue #860, opened 2026-08-09
An SFT example's observation should be the prompt its turn is sampled after (nemotron3, qwen3_5, kimi_k26)


rllin:
Three renderers build a supervised example whose observation ends differently from the prompt the same turn is sampled after. The model is trained on a context it is never given.

## The check

An SFT example's observation should equal `build_generation_prompt` for its prefix — which `build_supervised_example` already documents:

> The supervised example tokens should match what `build_generation_prompt` would produce for the same conversation prefix, so the model trains on the same distribution it sees at inference time.

Nothing asserts it. It needs no chat template and no second implementation, so it runs on every renderer:

```python
mi, w = renderer.build_supervised_example(messages, train_on_what=TrainOnWhat.ALL_ASSISTANT_MESSAGES)
ids, w = mi.to_ints(), list(w)
stop = len(w)
while stop and w[stop - 1] <= 0:
    stop -= 1
start = stop
while start and w[start - 1] > 0:
    start -= 1

observation = ids[:start]
sampled_after = renderer.build_generation_prompt(messages[:-1]).to_ints()
assert observation == sampled_after
```

## What it finds

Every case, on three renderers:

| renderer | trains after | sampled after |
|---|---|---|
| `nemotron3` | `<think></think>` | `<think>\n` |
| `qwen3_5` | `<think>\n\n</think>\n\n` | `<think>\n` |
| `kimi_k26` | `<\|im_middle\|>` | `<\|im_middle\|><think>` |

All three are thinking renderers, and in each the prefill is handled differently on the two paths. `deepseekv3`, `gpt_oss_medium_reasoning` and `qwen3_vl_instruct` are consistent.

Reproduction for `nemotron3`, using the renderer's own defaults:

```python
tk = AutoTokenizer.from_pretrained("nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16", trust_remote_code=True)
r = get_renderer("nemotron3", tk)
msgs = [{"role": "user", "content": "What is 2+2?"}, {"role": "assistant", "content": "4"}]

# SFT observation ends: '<|im_start|>assistant\n<think></think>'
# generation prompt   : '<|im_start|>assistant\n<think>\n'
# the template agrees with the generation prompt
```

## They are not the same problem

I checked whether each renderer's own document form is a strict extension of its own generation prompt, which decides whether the boundary can be moved at all:

[...]

PR #861, opened 2026-08-09
[renderers] the kimi think prefill is observation, not something the turn produced

rllin:
## What

```diff
  build_supervised_example([user, assistant "4"]), kimi_k25 and kimi_k26
- observation: …<|im_assistant|>assistant<|im_middle|>            trains '<think></think>4<|im_end|>'
+ observation: …<|im_assistant|>assistant<|im_middle|><think>     trains '</think>4<|im_end|>'
  generation prompt: …<|im_assistant|>assistant<|im_middle|><think>
```

## Why

`build_generation_prompt` prefills `<think>`, so the model is handed that token rather than
sampling it. Left in the message's output it is trained anyway, and the example's observation
stops one token short of the prompt the same turn is sampled after -- which is what
`build_supervised_example` documents it must match:

> The supervised example tokens should match what `build_generation_prompt` would produce for
> the same conversation prefix, so the model trains on the same distribution it sees at
> inference time.

Moving it to the header is what `Qwen3DisableThinkingRenderer` already does with its own empty
block, for the same reason and with the same comment.

The document form stays a strict extension of the generation prompt either way, so nothing
about the rendered tokens changes -- only which side of the observation boundary the prefill
falls on, and therefore whether it is trained.

## Test

`pytest tinker_cookbook/renderers/` -- 532 passed, 46 skipped, from 530 on `main`.

`test_the_think_prefill_is_observation_not_something_the_turn_produced`, over `kimi_k25` and
`kimi_k26`, verified red against `main`:

    trains after  '…<|im_middle|>'
    sampled after '…<|im_middle|><think>'

Found by a check from an internal conformance harness -- an example's observation must equal
the generation prompt for its prefix. It needs no chat template and no second implementation.
Two other renderers fail it for related reasons; those are #860, since one of them cannot be
fixed here.

### DIFF
--- a/tinker_cookbook/renderers/kimi_k25.py
+++ b/tinker_cookbook/renderers/kimi_k25.py
@@ -10,6 +10,8 @@ from tinker_cookbook.renderers.base import (
     ContentPart,
     ImageProcessorProtocol,
     Message,
+    RenderContext,
+    RenderedMessage,
     Role,
     ToolSpec,
     image_to_chunk,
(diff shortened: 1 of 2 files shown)


rllin:
Subsumed by #864, which fixes this for `kimi_k2` and `kimi_k25` together by moving the loss boundary rather than moving the prefill into the header.

The measurement that decided it: kimi's supervised render already contains `build_generation_prompt` as an exact **token** prefix — only the weights fell in the wrong place. So no tokens need to change at all, which this PR's approach did require.

It also closes the coverage hole underneath: `kimi_k25` was missing from `_CONSISTENCY_RENDERERS` while `kimi_k2` was present, and k25 is the one that adds the `<think>` prefill — so the bug was untested.

PR #862, opened 2026-08-09
[renderers] the qwen3_5 boundary falls inside a token merge

rllin:
The PR below this one puts the observation/action boundary at `build_generation_prompt` for
every renderer by re-weighting -- it compares the prompt as a **token** prefix of the render,
so it changes no tokens. `qwen3_5` is the one case that cannot reach:

```
prompt ends    [248068, 198]      = '<think>', '\n'
render has     [… 248068, 271 …]  = '<think>', '\n\n'
```

Token `271` (`\n\n`) **straddles the boundary**. No slice of the render is what the model is
handed, because the prompt stops mid-pair and sampling must start a fresh token there.

So the split has to happen before the tokenizer sees it. The header is assembled as a string
and the generation suffix is a string prefix of it, so the renderer cuts the string it just
built and encodes the two halves separately:

```python
prefix_str = self._produced_turn_prefix_str(message, ctx)
if prefix_str and header_str.startswith(prefix_str):
    output_chunks_encoded.insert(0, encode(header_str[len(prefix_str):]))
    header_str = prefix_str
```

Both sides come out tokenised the way the model will see them, by construction. Nothing is
decoded, no chunk is rebuilt, and image chunks pass through untouched -- an earlier version
of this did the same split by decoding the whole message back to text and re-encoding it,
which had to bail on any non-text chunk.

`_generation_suffix_str` is now the single place the prompt's tail is written.
`Qwen3_5DisableThinkingRenderer` overrides that string rather than `_get_generation_suffix`,
so the prompt it emits and the split it takes cannot drift apart -- overriding only the
encoding is exactly how they did drift while this was being written.

`qwen3_5` leaves `_RENDERERS_WITH_DIFFERENT_SUPERVISED_GEN_HEADERS`, which drops to
`{nemotron3}`.

## Test Plan

`test_supervised_generation_parse_consistency` asserts
`observation == build_generation_prompt(messages[:-1])` and exempted `qwen3_5` by name.
Removing the exemption runs its cases; they fail without the renderer change and pass with it.

`pytest tinker_cookbook/renderers/` -- 554 passed / 34 skipped, unchanged by the rewrite from
the decode-and-re-encode version to the string split.

---

### DIFF
--- a/tinker_cookbook/renderers/qwen3.py
+++ b/tinker_cookbook/renderers/qwen3.py
@@ -739,10 +739,32 @@ class Qwen3VLRenderer(Qwen3Renderer):
                     )
                 )
 
+        # The generation prompt may end partway through this header. Splitting the string
+        # before encoding puts the boundary where sampling starts and keeps both sides
+        # tokenised the way the model will see them.
+        prefix_str = self._produced_turn_prefix_str(message, ctx)
+        if prefix_str and header_str.startswith(prefix_str):
+            rest = header_str[len(prefix_str) :]
+            if rest:
+                output_chunks_encoded.insert(
+                    0,
+                    tinker.EncodedTextChunk(
+                        tokens=self.tokenizer.encode(rest, add_special_tokens=False)
+                    ),
+                )
+            header_str = prefix_str
+
         header = tinker.types.EncodedTextChunk(
             tokens=self.tokenizer.encode(header_str, add_special_tokens=False)
         )
         return RenderedMessage(header=header, output=output_chunks_encoded)
+
+    def _produced_turn_prefix_str(self, message: Message, ctx: RenderContext) -> str:
+        """How much of this message's header the generation prompt already wrote.
+
+        Empty when the prompt stops before the header does, which is the usual case.
+        """
+        return ""
 
 
 class Qwen3VLInstructRenderer(Qwen3VLRenderer):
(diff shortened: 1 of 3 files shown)


rllin:
Superseded by #863, which fixes this structurally in `build_supervised_example` rather than in `qwen3_5`'s `_get_generation_suffix`.

Measuring against the model this repo's tests actually pair with each renderer showed the boundary was wrong in `deepseekv3_thinking` and `nemotron3` too, in the same way — so the per-renderer version here was one of three restatements of a single rule, each phrased in whatever terms that renderer uses to say where the prompt ends. #863 covers all three and removes `qwen3_5` from `_RENDERERS_WITH_DIFFERENT_SUPERVISED_GEN_HEADERS` outright.

Leaving this open rather than closing it, in case you'd rather take the narrow fix — happy to close if #863 is the preferred shape.

rllin:
Correcting my earlier comment: I said this was superseded by #863, but #863 is now closed in favour of #864, and **#864 does not cover this case**.

#864 moves the loss boundary by comparing the prompt as a token prefix, which is exact and changes no tokens — but `qwen3_5` with a no-ThinkingPart target is precisely where that fails. The render has `<think>`,`\n\n` where the prompt has `<think>`,`\n`: token `271` straddles the boundary, so no slice of the render equals what the model is actually given.

That leaves this PR as the only thing covering it. Reopening the question rather than closing: it needs re-encoding, and #864 deliberately avoids that everywhere else.

rllin:
Converted to ghstack; this is now #877.

PR #863, opened 2026-08-09
Train the turn after the prompt it is sampled from

rllin:
## Why is this change needed?

`build_supervised_example` should produce the tokens `build_generation_prompt` produces for the same prefix, so training and sampling see the same context. Renderers whose prompt ends in a prefill — an open `<think>` — left it on the wrong side, training the model to produce a token it is always given:

```diff
  deepseekv3_thinking, a turn that reasons
  sampled from:  <｜User｜>q<｜Assistant｜><think>
- observation:   <｜User｜>q<｜Assistant｜>          action: <think>R</think>4<eos>
+ observation:   <｜User｜>q<｜Assistant｜><think>   action: R</think>4<eos>
```

| renderer | effect |
|---|---|
| `qwen3_5` | fixed; leaves `_RENDERERS_WITH_DIFFERENT_SUPERVISED_GEN_HEADERS` |
| `deepseekv3_thinking`, `nemotron3` | fixed for turns that reason |
| `kimi_k25` | unchanged — overrides `build_supervised_example`, so #861 is still needed |
| everything else | untouched; the guard returns early |

Normalised in the assembly rather than per renderer because the fix needs the prefix's generation prompt, and `render_message` — where #862 and #861 put it — only ever sees one message. Supersedes #862.

The split is on text, since a boundary can fall inside a token merge, so `test_supervised_example_against_hf_chat_templates` compares text. A supervised target with no ThinkingPart stays exempt for `deepseekv3_thinking` and `nemotron3`: it keeps a header of its own where `build_generation_prompt` ends with an open `<think>`, which is the case `_RENDERERS_WITH_DIFFERENT_SUPERVISED_GEN_HEADERS` is named for.

## Test Plan

`pytest tinker_cookbook/renderers/` — **536 passed / 40 skipped**, from 530 / 46 on `main`. The six are `qwen3_5` cases that previously skipped `test_supervised_generation_parse_consistency`, which asserts `observation == build_generation_prompt(messages[:-1])`; they fail without this change.

`ALL_TOKENS` / `ALL_MESSAGES` / `ALL_ASSISTANT_MESSAGES` verified untouched — they weight the prompt region deliberately, so the normalisation is gated to the two last-turn modes.

Related: #858, #859, #860, #861, #862.

### DIFF
--- a/tinker_cookbook/renderers/base.py
+++ b/tinker_cookbook/renderers/base.py
@@ -22,6 +22,7 @@ from typing import (
     Protocol,
     TypedDict,
     Union,
+    cast,
 )
 
 import pydantic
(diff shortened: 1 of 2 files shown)


rllin:
Superseded by #864, which does the same job by re-weighting instead of re-encoding.

This PR split the render as **text** and re-encoded the action, which changed token sequences and forced `test_supervised_example_against_hf_chat_templates` to compare text instead of tokens. #864 compares the prompt as a token prefix and moves only the weights, so no token sequence changes and that test is untouched.

#864 also reaches `kimi_k2`/`kimi_k25`, which this PR could not — their `build_supervised_example` override meant the base normalisation never ran.

What #864 does **not** carry over is the one cell this PR fixed by re-encoding: `qwen3_5` with a no-ThinkingPart target, where the boundary falls inside a `\n\n` merge. #862 still covers that case and stays open.

PR #864, opened 2026-08-09
Withhold loss from what the generation prompt already supplied

rllin:
A renderer whose generation prompt ends in a prefill -- kimi's open `<think>` -- left it
on whichever side its own header logic put it, so the model was trained to produce a token
it is always given. Only the weights move; **the tokens are untouched**:

```diff
  kimi_k25, sampled from …<|im_assistant|>assistant<|im_middle|><think>
- observation …<|im_assistant|>assistant<|im_middle|>          action <think>R</think>4
+ observation …<|im_assistant|>assistant<|im_middle|><think>   action R</think>4
```

The observation is `build_generation_prompt(messages[:boundary])`, compared as tokens and
used only to place the boundary, so no chunk is rebuilt and nothing is re-encoded. When the
render does not begin with the prompt the weights are returned unchanged -- that means the
two disagree about more than the boundary, which is a renderer bug and not something a
re-weight can fix.

| renderer | before | after |
|---|---|---|
| `kimi_k2`, `kimi_k25` | prefill trained | **prefill observed** |
| `qwen3_5`, `nemotron3`, `deepseekv3_thinking`, turn with reasoning | prefill trained | **prefill observed** |
| same three, turn with no reasoning | wrong | unchanged; still exempt |
| `qwen3`, `llama3`, `gpt_oss` | correct | untouched |

`KimiK2Renderer` reimplements `build_supervised_example` rather than delegating, so it calls
the same helper at its own return.

## Test Plan

`kimi_k25` was absent from `_CONSISTENCY_RENDERERS` while `kimi_k2` was present -- and it is
`kimi_k25` that adds the `<think>` prefill, so the bug was untested. Adding it:
**6 failed before this change, 9 passed after.**

`pytest tinker_cookbook/renderers/` -- 536 passed / 46 skipped, from 530 / 46 on main. No
token sequence changes, so the HF chat-template comparisons are untouched.

---

**Stack** (base → top): #866 → **#864** → #858 → #859 → #865 → #862 → #867

> **Note:** #858 ("the empty think block has one writer") was squash-merged into this branch on 2026-08-09 rather than into `main`, because its base was a feature branch and so had no required checks to wait on. Its commit is now part of this PR, which is why the diff is larger than the description above covers. The stack above has been rebased to drop the duplicate.

### DIFF
--- a/tinker_cookbook/renderers/base.py
+++ b/tinker_cookbook/renderers/base.py
@@ -1825,7 +1825,63 @@ class Renderer(ABC):
         weights_tensor = torch.tensor(weights_data)
 
         model_input_chunks = [chunk for chunk, _ in model_input_chunks_weights]
-        return tinker.ModelInput(chunks=model_input_chunks), weights_tensor
+        model_input = tinker.ModelInput(chunks=model_input_chunks)
+        return model_input, self._train_after_the_generation_prompt(
+            messages, train_on_what, model_input, weights_tensor
+        )
+
+    def _train_after_the_generation_prompt(
+        self,
+        messages: list[Message],
+        train_on_what: TrainOnWhat,
+        model_input: tinker.ModelInput,
+        weights: torch.Tensor,
+    ) -> torch.Tensor:
+        """Withhold loss from everything build_generation_prompt would have supplied.
+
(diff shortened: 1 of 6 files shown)


rllin:
Converted to ghstack; this is now #873.

PR #865, opened 2026-08-09
[renderers] a deepseek supervised target is rendered the way sampling makes it

rllin:
`build_generation_prompt` prefills `<think>`, so a turn is sampled as `R</think>answer` --
or, when it did not reason, as `</think>answer`. The renderer wrote neither. A target with
no ThinkingPart came out with no think tokens at all:

```diff
  sampled after:  <｜User｜>q<｜Assistant｜><think>
- observation …<｜Assistant｜>          action 4<｜end▁of▁sentence｜>
+ observation …<｜Assistant｜><think>   action </think>4<｜end▁of▁sentence｜>
```

so the model was trained to answer straight after `<｜Assistant｜>` while inference always
hands it `<think>` first. Both shapes now open the block and let the boundary land inside
it, which is what the PR below this one places.

**This deliberately does not match `apply_chat_template`.** HF renders a finished assistant
turn the way it renders *history*: a bare `</think>answer`, with no opening tag anywhere --
the opening `<think>` exists only in the generation prompt. That form is unmatched by
construction and no sampling can produce it, so it is the wrong target to train on even
though it is the right thing for the endpoint to write when the turn later becomes history.

`_RENDERERS_WITH_DIFFERENT_SUPERVISED_GEN_HEADERS` was answering two questions at once --
"diverges from the HF document" and "observation != generation prompt" -- which is why
`deepseekv3_thinking` could not leave it. Split, and deepseek leaves the second:

| set | members | governs |
|---|---|---|
| `_RENDERERS_THAT_DIVERGE_FROM_HF_FOR_THE_TARGET` | deepseek, qwen3_5, nemotron3 | the document comparison |
| `_RENDERERS_WITH_DIFFERENT_SUPERVISED_GEN_HEADERS` | qwen3_5, nemotron3 | the boundary comparison |

## Test Plan

`test_supervised_generation_parse_consistency` asserts
`observation == build_generation_prompt(messages[:-1])` and skipped `deepseekv3_thinking`
entirely. It no longer does: **its cases fail on main for both turn shapes and pass here.**

`pytest tinker_cookbook/renderers/` -- 548 passed / 40 skipped, from 545 / 43.

---

**Stack** (base → top): #866 → #864 → ~~#858 merged~~ → #859 → #865 → #862 → #867 → #868

### DIFF
--- a/tinker_cookbook/renderers/deepseek_v3.py
+++ b/tinker_cookbook/renderers/deepseek_v3.py
@@ -25,6 +25,7 @@ from tinker_cookbook.renderers.base import (
     UnparsedToolCall,
     detect_unterminated_tool_block,
     ensure_text,
+    has_thinking,
     parse_response_for_stop_token,
     parse_think_blocks,
 )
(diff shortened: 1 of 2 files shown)


rllin:
Converted to ghstack; this is now #876.

PR #866, opened 2026-08-09
[renderers] one writer for "does this content carry reasoning"

rllin:
Four renderers each spelled the same predicate inline, and a fifth spelled a different
one under the same name:

| site | list | string |
|---|---|---|
| `qwen3.Qwen3DisableThinkingRenderer` | any ThinkingPart | `"<think>" in content` |
| `qwen3_5._assistant_header_suffix` | any ThinkingPart | `"<think>" in content` |
| `nemotron3._assistant_header_suffix` (x2) | any ThinkingPart | `"<think>" in content` |
| `nemotron3._format_tool_calls_chunks` | any ThinkingPart | **always False** |

The first four are byte-identical and become `has_thinking(content)`, next to its siblings
`remove_thinking` / `ensure_list` / `parse_think_blocks` in `base.py`.

The fifth is deliberately different and stays inline, renamed `has_thinking_part` so it no
longer shadows the helper: it asks whether `_format_thinking_text` ran, and that only runs
on a ThinkingPart. An inline `<think>` in a string never went through it, so it never wrote
the trailing newline that code is compensating for. Answering it with the general predicate
would be a behaviour change, not a refactor.

## Test Plan

`pytest tinker_cookbook/renderers/` -- 530 passed / 46 skipped, identical to main. This is a
pure extraction: no call site changes which question it asks.

---

**Stack** (base → top): **#866** → #864 → #858 → #859 → #865 → #862 → #867

### DIFF
--- a/tinker_cookbook/renderers/base.py
+++ b/tinker_cookbook/renderers/base.py
@@ -851,6 +851,18 @@ def message_to_jsonable(message: Message) -> dict[str, Any]:
     return result
 
 
+def has_thinking(content: Content) -> bool:
+    """Whether content carries reasoning, in either shape it can arrive in.
+
+    Structured content carries a ThinkingPart; a string carries an opening ``<think>``,
+    closed or not. Renderers ask this to decide whether to write the empty block their
+    template puts on a turn that did not reason.
+    """
+    if isinstance(content, list):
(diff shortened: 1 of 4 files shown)


rllin:
Converted to ghstack; this is now #872.

PR #867, opened 2026-08-09
[renderers] the produced turn is a concept, not a lie about is_last

rllin:
`ctx.is_last` means "the final message". Four renderers were using it to mean "part of the
turn being produced", and two had to falsify it to get the right answer:

```python
kimi_k2       is_last=is_last_assistant_turn            # not the last message
nemotron3     dataclasses.replace(ctx, is_last=True)    # to defeat the base's check
```

A turn can span several messages -- an assistant message, a tool response, the assistant's
follow-up -- so "is this the last message" cannot answer "should this keep its reasoning".
`RenderContext.in_produced_turn` answers it, from one overridable
`_produced_turn_start_index`.

**`KimiK2Renderer.build_supervised_example` goes away.** It was 100 lines reimplementing the
base loop to change two things: where the turn starts, and that `ctx` should report it. Both
are the base's job now, so what is left is the boundary:

```python
def _produced_turn_start_index(self, messages):
    for idx in range(len(messages) - 2, -1, -1):
        if messages[idx]["role"] == "assistant" and not messages[idx].get("tool_calls"):
            return idx + 1
    return 0
```

`kimi_k2.py` **-100/+21**; `RendererError` became unused, its only `raise` having been in the
duplicated `train_on_what` dispatch.

Where the turn starts is **not** where the prompt ends. Kimi puts the first at 0 when no
earlier assistant qualifies, while the prompt still ends after the last user message, so
`_first_trained_message_index` keeps computing its own boundary. Conflating them retrained
kimi's `<think>` prefill -- the exact thing the PR below this one fixes.

`nemotron3`'s patch stays. Rewriting `qwen3` to read `in_produced_turn` so it could go was
tried and reverted: it fails `test_generation_against_hf_chat_templates[Qwen3-8B-tool_call_gen]`,
because Qwen3's template strips reasoning from every message but the literal last while
Nemotron-3's keeps the whole turn. That patch encodes a template difference.

## Test Plan

`test_last_assistant_turn_observes_the_prefill_too` is new, and is what should have existed
already: `LAST_ASSISTANT_TURN` weights a different amount than `LAST_ASSISTANT_MESSAGE` but
is sampled from the same prompt, so it must observe the same prefill. Only the latter was
covered. **6 failed / 6 passed against the conflated boundary, 12 pass now.**

Verified by differential rather than by test count: 40 (renderer x conversation x
`train_on_what`) cases dumped as tokens and weights before and after, **0 differ**. The
conflation bug above changed 5 of them while all 551 tests stayed green, which is why the
new test exists.

`pytest tinker_cookbook/renderers/` -- 563 passed / 37 skipped.

---

### DIFF
--- a/tinker_cookbook/renderers/base.py
+++ b/tinker_cookbook/renderers/base.py
@@ -705,6 +705,15 @@ class RenderContext:
     differently based on whether they come before or after the last user message.
     """
 
+    in_produced_turn: bool = False
+    """Whether this message belongs to the turn the model is being asked to produce.
+
+    Distinct from ``is_last``, which is literally the final message. A turn can span several
+    messages -- an assistant message, a tool response, the assistant's follow-up -- so a
+    renderer that keeps reasoning in the produced turn while stripping it from history has to
+    ask about the turn, not the message. ``_produced_turn_start_index`` says where it starts.
+    """
+
 
 class ToolSpec(TypedDict):
     """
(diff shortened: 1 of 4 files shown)


rllin:
Converted to ghstack; this is now #878.

PR #868, opened 2026-08-09
[renderers] build_supervised_examples splits a conversation instead of refusing it

rllin:
A renderer without the extension property writes an assistant message one way as history
and another as the turn being produced, so one sequence cannot show every turn the context
it was sampled from. `build_supervised_examples` exists for exactly that -- and raised for
every renderer but `kimi_k2` and `tml_v0`:

    NotImplementedError: build_supervised_examples has not been implemented for this renderer.

So a caller with a multi-turn conversation had no correct option. The singular form silently
renders the earlier turns against a prefix they were never sampled from -- the failure its
own `ALL_ASSISTANT_MESSAGES` warning describes -- and the plural form refuses.

The default now does what that warning tells callers to do by hand:

```python
return [
    self.build_supervised_example(messages[: idx + 1], train_on_what=LAST_ASSISTANT_MESSAGE)
    for idx in candidates
    if messages[idx]["role"] == "assistant"
]
```

`LAST_ASSISTANT_MESSAGE` per example is the point: it is the mode whose boundary is already
covered by `test_supervised_generation_parse_consistency`, so each split example inherits an
invariant the whole conversation cannot have.

Modes that weight non-assistant messages still raise, with a reason rather than a TODO:
there is no per-turn prompt to end those examples at.

The `# TODO: ... merges examples with shared prefixes` is deliberately not done. That is a
compute optimisation, not correctness, and bundling it would widen the claim.

## Test Plan

`test_every_example_of_a_split_conversation_starts_where_it_was_sampled` asserts every
example's observation is the generation prompt for its own prefix. **27 failed with
`NotImplementedError` before this change, 27 pass after.**

Scoped to renderers using the default split; `kimi_k2` overrides it with a coarser one
(per user turn rather than per assistant message) and is skipped. That split does not hold
this invariant on tool conversations -- a pre-existing issue in kimi's own implementation,
untouched here.

`pytest tinker_cookbook/renderers/` -- 590 passed / 94 skipped.

---

### DIFF
--- a/tinker_cookbook/renderers/base.py
+++ b/tinker_cookbook/renderers/base.py
@@ -1694,7 +1694,10 @@ class Renderer(ABC):
         """Build tokens and per-token weights for supervised fine-tuning.
 
         Returns a list of (model_input, weights) tuples. Multiple examples are
-        needed when the renderer does not satisfy the extension property.
+        needed when the renderer does not satisfy the extension property: it writes an
+        assistant message one way as history and another as the turn being produced, so a
+        single sequence cannot show every turn the context it was sampled from. One example
+        per trained message can, each ending at the message it trains.
 
         Args:
             messages (list[Message]): The conversation to render.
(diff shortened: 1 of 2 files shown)


rllin:
Converted to ghstack; this is now #879.

PR #870, opened 2026-08-09
[renderers] kimi's generation prompt is the base's too

rllin:
`KimiK2Renderer.build_generation_prompt` was a hand-rolled copy of the base loop -- the
sibling of the `build_supervised_example` copy deleted two PRs below -- differing only in
where the turn starts and in writing its own suffix inline. The base's default suffix
already produces exactly what it wrote:

    '<|im_assistant|>assistant<|im_middle|>'

so what is left is the system-message normalisation, and **-51 lines** go.

The turn boundary is what kept the two apart. The paths ask the same question of different
lists: `build_generation_prompt` is prompting for a turn that comes *after* its messages,
while `build_supervised_example` is rendering a turn that *is* the last of its own. Kimi
answered that with two scans, `range(len - 1, ...)` and `range(len - 2, ...)`, which disagree
on three of the four message shapes that reach them.

One scan, and the caller says which list to look at:

```python
# build_generation_prompt: the turn comes after these
turn_start = self._produced_turn_start_index(messages)
# build_supervised_example: the turn is the last of these, and cannot start after itself
turn_start = self._produced_turn_start_index(messages[:-1])
```

The indices returned still address the full list either way, so nothing downstream changes.
This also removes the last reason `<think>` had to be written through the caller's `prefill`
parameter -- `_get_generation_suffix` is now reachable from kimi, which is what the PR below
this one worked around.

## Test Plan

`pytest tinker_cookbook/renderers/` -- 638 passed / 132 skipped, unchanged.

Verified by differential rather than by test count, the way the earlier boundary regression
should have been: 40 (renderer x conversation x `train_on_what`) cases dumped as tokens and
weights, plus generation prompts, before and after -- **0 differ**.

---

**Stack** (base → top): #866 → #864 → #859 → #865 → #862 → #867 → #868 → #869 → **#870**

### DIFF
--- a/tinker_cookbook/renderers/base.py
+++ b/tinker_cookbook/renderers/base.py
@@ -1796,7 +1796,10 @@ class Renderer(ABC):
             default=-1,
         )
 
-        turn_start = self._produced_turn_start_index(messages)
+        # Without the target: the produced turn is the last message here, and a turn cannot
+        # start after itself. build_generation_prompt passes the whole list, where the turn
+        # it is prompting for comes after the end.
+        turn_start = self._produced_turn_start_index(messages[:-1])
 
         for idx, message in enumerate(messages):
             if train_on_what == TrainOnWhat.CUSTOMIZED:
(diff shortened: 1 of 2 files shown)


rllin:
Converted to ghstack; this is now #881.

PR #871, opened 2026-08-09
[renderers] qwen3 with thinking disabled refuses a turn with no query

rllin:
Qwen3's template gates the empty think block on `loop.index0 > ns.last_query_index`, and
`last_query_index` defaults to the final index when there is no user message -- so nothing
is ever after it and the document carries no block. `add_generation_prompt` writes one
regardless:

```
[system, assistant]   prompt   <|im_start|>assistant\n<think>\n\n</think>\n\n
                      document <|im_start|>assistant\na<|im_end|>
```

Both are the template's, and they cannot both be honoured: the turn trains after a prompt
that always opens a block the training sequence does not contain. Rendering it anyway is a
silent train/serve gap, so it raises instead. A conversation with a user message is
unaffected, and observes the block it is sampled after.

The renderers below this one made the conflict visible by matching the template on both
sides; before, the block was written unconditionally, diverging from the document and
hiding the question.

## Test Plan

`test_qwen3_disable_thinking_refuses_a_turn_with_no_query` asserts the refusal, and asserts
the same turn renders with `observation == build_generation_prompt(...)` once a user message
is present.

`pytest tinker_cookbook/renderers/` -- 639 passed / 132 skipped.

The monorepo's `lib/chat_conformance` goes to **zero failures** against this stack (from 20
before the ledger update); the three cases this refuses become skips, which is how that
harness records "cannot render" rather than "disagrees".

---

**Stack** (base → top): #866 → #864 → #859 → #865 → #862 → #867 → #868 → #869 → #870 → **#871**

### DIFF
--- a/tinker_cookbook/renderers/qwen3.py
+++ b/tinker_cookbook/renderers/qwen3.py
@@ -14,6 +14,7 @@ from typing import cast
 
 import tinker
 
+from tinker_cookbook.exceptions import RendererError
 from tinker_cookbook.image_processing_utils import ImageProcessor
 from tinker_cookbook.renderers.base import (
     ContentPart,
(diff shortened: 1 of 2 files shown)


rllin:
Converted to ghstack; this is now #882.

PR #878, opened 2026-08-10
[renderers] the produced turn is a concept, not a lie about is_last

rllin:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.16.0) (oldest at bottom):
* #882
* #881
* #880
* #879
* __->__ #878
* #877
* #876
* #875
* #874
* #873
* #872

## Why is this change needed?

`KimiK2Renderer` carried its own copy of `build_supervised_example` — 100 lines duplicating the base loop — to change one thing: **which messages count as the turn being trained**.

It had to copy, because the base asks that question through `ctx.is_last`, and `is_last` means "the final message in the list". Kimi's turn spans several messages (assistant → tool → assistant), so it faked the flag instead:

```python
kimi_k2     is_last=is_last_assistant_turn            # not the final message
nemotron3   dataclasses.replace(ctx, is_last=True)    # same trick, to defeat the base's check
```

`ctx.in_last_assistant_turn` now carries that directly, computed once from `_last_assistant_turn_start_index` — the one thing kimi actually needed to override. Kimi's turn starts after the last assistant message that did not call a tool; the default is after the last user message.

So the copy goes: **`kimi_k2.py` −100/+21**.

Same story for the second thing kimi's copy existed to change — which messages get loss. That `TrainOnWhat` rule moved out of the loop into `_output_is_trained`, so overriding it no longer means copying the loop around it.

Divergences retired: **none directly** — this is what makes #879's two possible.

## Test Plan

`test_last_assistant_turn_observes_the_prefill_too` is new; only `LAST_ASSISTANT_MESSAGE` had boundary coverage before.

Verified by **differential**: 40 cases of tokens and weights before/after — **0 differ**. An earlier version of this changed 5 of them while all 551 tests stayed green, which is why the test exists.

### DIFF
--- a/tinker_cookbook/renderers/base.py
+++ b/tinker_cookbook/renderers/base.py
@@ -1789,8 +1789,6 @@ class Renderer(ABC):
                 )
 
             is_last_message = idx == len(messages) - 1
-            is_assistant = message["role"] == "assistant"
-            is_user_or_system = message["role"] in ["user", "system"]
             in_produced_turn = idx >= turn_start
 
             # only apply weight to header if train_on_what is ALL_TOKENS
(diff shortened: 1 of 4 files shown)


rllin:
/stamp-plz simplify abstraction and clarify meaning for is_last

PR #882, opened 2026-08-10
[renderers] qwen3 with thinking disabled refuses a turn with no query

rllin:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.16.0) (oldest at bottom):
* __->__ #882
* #881
* #880
* #879
* #878
* #877
* #876
* #875
* #874
* #873
* #872

## Why is this change needed?

A supervised example is normally the generation prompt plus what the model wrote — the finished conversation *starts with* the prompt, so the observation/action boundary falls out for free:

```
apply_chat_template(prefix, add_generation_prompt=True)   …assistant\n<think>\n\n</think>\n\n
apply_chat_template(whole,  add_generation_prompt=False)  …assistant\n<think>\n\n</think>\n\na<|im_end|>\n
                                                          ^--------- the prompt --------^
```

With no user message Qwen3's template stops doing that: it still writes the block into the prompt, but leaves it out of the finished conversation. There is then no split point, and we rendered one anyway —

```diff
  qwen3_disable_thinking, [system, assistant]
  sampled after  '…<|im_start|>assistant\n<think>\n\n</think>\n\n'
- observation    '…<|im_start|>assistant\n'                      ← block missing
- action         'a<|im_end|>'
+ RendererError: this conversation has no user message, so Qwen3's template writes no
+ empty think block on the produced turn -- but the generation prompt always opens one.
```

The example trained the turn after a context the model is never handed. A template can't arbitrate the boundary — it has no loss weights, and Qwen3's carries no `{% generation %}` markers — so there is no rendering to fall back to.

**The caller has a working alternative**; only the disable-thinking variant has this conflict:

| renderer | `[system, assistant]` |
|---|---|
| `qwen3` | renders, `observation == prompt` |
| `qwen3_instruct` | renders, `observation == prompt` |
| `qwen3_disable_thinking` | **refuses** |

Not fixable by always writing the block: tested, and it makes **plain `qwen3`'s prompt** diverge from the template — 4 harness failures, on a renderer this conflict never touched.

Divergences retired: **none** — it turns three silently mis-rendered conversations into refusals, which the harness records as skips.

## Test Plan

`test_qwen3_disable_thinking_refuses_a_turn_with_no_query` **fails on `main`** — without the refusal the render succeeds. Also asserts the turn still renders once a user message is present.

### DIFF
--- a/tinker_cookbook/renderers/qwen3.py
+++ b/tinker_cookbook/renderers/qwen3.py
@@ -14,6 +14,7 @@ from typing import cast
 
 import tinker
 
+from tinker_cookbook.exceptions import RendererError
 from tinker_cookbook.image_processing_utils import ImageProcessor
 from tinker_cookbook.renderers.base import (
     ContentPart,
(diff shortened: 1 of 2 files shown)


rllin:
/land --force fixed according to internal chat_conformance tests

PR #884, opened 2026-08-12
Pipeline DPO optimizer requests before awaiting results

svdrecbd:
## Summary

- submit the DPO optimizer request immediately after the custom backward request is enqueued
- consume the backward and optimizer futures together instead of blocking before optimizer submission
- add trace spans for custom-backward enqueue, optimizer enqueue, and result consumption
- propagate optimizer metrics and add a regression test for the required request ordering

## Root cause

The DPO loop previously called `.result()` on `forward_backward_custom()` before it submitted `optim_step()`. On Tinker's clocked worker pools, that ordering can miss the optimizer slot in the current cycle and increase step latency.

The SDK contract makes the safe ordering precise: `forward_backward_custom_async()` performs the policy-logprob forward and the client-side custom loss, enqueues the resulting backward request, and only then returns its future. Submitting `optim_step_async()` immediately afterward therefore preserves the gradient/update dependency while allowing both queued operations to occupy the intended worker-pool schedule.

This is also the established cookbook pattern for custom-loss training: `tinker_cookbook/distillation/sdft.py::_train_step_reverse_kl` already enqueues `forward_backward_custom_async()` and `optim_step_async()` back-to-back before consuming either result.

This patch changes request scheduling only. It passes the same ordered datums, custom loss callable, and Adam parameters, and it does not pipeline gradients across optimizer steps.

## Validation

- `uv run pytest tinker_cookbook/preference -q`
- `uv run ruff check tinker_cookbook/preference/train_dpo.py tinker_cookbook/preference/train_dpo_test.py`
- `uv run pyright tinker_cookbook/preference/train_dpo.py tinker_cookbook/preference/train_dpo_test.py`
- `uv run pre-commit run --files tinker_cookbook/preference/train_dpo.py tinker_cookbook/preference/train_dpo_test.py`
- upstream CI: pytest, pyright, pre-commit, downstream compatibility, and stack freshness all pass

The regression test fails if either result is consumed before both operations have been enqueued. No paid service run was used for this correctness validation.

### DIFF
--- a/tinker_cookbook/preference/train_dpo.py
+++ b/tinker_cookbook/preference/train_dpo.py
@@ -4,6 +4,7 @@ Direct Preference Optimization (DPO) training
 
 import asyncio
 import logging
+from collections.abc import Callable
 from pathlib import Path
 from typing import cast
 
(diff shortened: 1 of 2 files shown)


svdrecbd:
Live service follow-up on the scheduling change:

- Model: NVIDIA Nemotron 3 Ultra (550B total / 55B active)
- Same immutable DPO trajectory, batch size, loss, optimizer, and checkpoint lineage
- Sequential baseline segment: 150 updates
- Patched continuation segment: next 150 updates
- Backend `clock_cycle:unique` gap: median 6 → 4 cycles/update, mean 5.95 → 4.10, p90 7 → 5
- Training-step wall time: 15,119 s → 8,401 s (4h12m → 2h20m), a 44% reduction / 1.80× observed throughput

The backend-cycle result is the cleaner scheduling measure (1.5×); wall time also reflects provider/load variation across the sequential segments. The candidate artifact preserved the baseline batch history byte-for-byte and advanced the same optimizer/model lineage from step 151 to 301. No updates from different optimizer steps were overlapped.

PR #886, opened 2026-08-13
[renderers] message text cannot write Qwen turn markers

rllin:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.17.0) (oldest at bottom):
* __->__ #886

## Why is this change needed?

Qwen tokenizers recognize `<|im_start|>` and `<|im_end|>` wherever they appear. Encoding a message and the renderer-authored end marker as one string therefore lets quoted protocol text create real turn boundaries.

Encode message output with special-token splitting and append the renderer-authored end token separately.

| conformance measure | before | after |
|---|---:|---:|
| quoted-marker divergences across `qwen3`, `qwen3_instruct`, and `qwen3_disable_thinking` | 3 | 0 |
| all Qwen Rust/cookbook divergence cases | 7 | 4 |

## Test Plan

The new parametrized test fails before the change because the quoted markers add one start-token ID and one end-token ID. It now passes for all three renderers while preserving the quoted text.

`pytest tinker_cookbook/renderers/qwen3_test.py tinker_cookbook/renderers/tool_calling_test.py tinker_cookbook/renderers/renderers_test.py -q` — 360 passed, 132 skipped, 1 xfailed.

### DIFF
--- a/tinker_cookbook/renderers/qwen3.py
+++ b/tinker_cookbook/renderers/qwen3.py
@@ -283,14 +283,18 @@ class Qwen3Renderer(Renderer):
                 ]
             )
             output_content += ("\n" if has_text else "") + calls
-        output_content += "<|im_end|>"
         header = tinker.types.EncodedTextChunk(
             tokens=self.tokenizer.encode(header_str, add_special_tokens=False)
         )
         output: list[tinker.ModelInputChunk] = [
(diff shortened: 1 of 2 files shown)


rllin:
/stamp-plz fix in-content markers so they are not accidnetally made special

PR #890, opened 2026-08-14
[harbor_rl] Upload test fixtures as bytes so binary graders work

### LINKED ISSUE #889: [harbor_rl] Binary test fixtures silently zero out 7 Terminal-Bench tasks
### Summary

`HarborReward._upload_tests` reads every file under a task's `tests/` directory as text:

https://github.com/thinking-machines-lab/tinker-cookbook/blob/main/tinker_cookbook/recipes/harbor_rl/harbor_tools.py#L100

```python
content = file_path.read_text()
```

Terminal-Bench tasks whose graders compare against binary ground truth raise `UnicodeDecodeError` here. `HarborReward.__call__` catches it and returns `reward=0.0, grading_error=1.0`, so those tasks score zero no matter what the agent did.

### Affected tasks

7 of the 89 tasks in `terminal-bench@2.0`:

| Task | Binary fixture |
|---|---|
| `build-pov-ray` | `tests/reference_illum1.png` |
| `make-doom-for-mips` | `tests/reference.jpg` |
| `make-mips-interpreter` | `tests/reference.jpg` |
| `pytorch-model-recovery` | `tests/weights_gtruth.pt` |
| `sam-cell-seg` | `tests/test_img.png` |
| `train-fasttext` | `tests/private_test.tar.gz` |
| `video-processing` | `tests/test_video.mp4` |

These fixtures are load-bearing, not incidental — `sam-cell-seg/tests/test.sh:25` does `cp /tests/test_img.png .`, and `make-mips-interpreter/tests/test_outputs.py:71` reads `/tests/reference.jpg` to run an L2 image comparison against the agent's output.

### Observed

From a `train_terminal_bench.py` run on `moonshotai/Kimi-K2.6`:

```
harbor_tools:87 [ERROR] Harbor grading failed: 'utf-8' codec can't decode byte 0x89 in position 0: invalid start byte
harbor_tools:87 [ERROR] Harbor grading failed: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
harbor_tools:87 [ERROR] Harbor grading failed: 'utf-8' codec can't decode byte 0x8b in position 1: invalid start byte
```

`0x89` = PNG, `0xff` = JPEG, `0x8b` = gzip — matching the table above.

### Why this is easy to miss

The grading failure is swallowed rather than raised, so it never surfaces as an error. In `eval.py`:

```python
status = "ERROR" if result.error else ("PASS" if result.reward > 0 else "FAIL")
```

A returned `0.0` is classified `FAIL`, indistinguishable from a genuine miss. In the README's Terminal-Bench 2.0 results (14 PASS / 31 FAIL / 44 ERROR) these would sit inside the 31 FAILs.

For RL specifically it is worse than a lost eval sample: because the reward is constant zero across the group, advantages center to zero and the group contributes no gradient, so the sandboxes are paid for and produce nothing.

### Fix

Read as bytes. `SandboxInterface.write_file` already accepts `str | bytes`, and `ModalSandbox.write_file` encodes `str` to bytes before streaming to stdin, so bytes is the path it was already built for.

PR with a regression test to follow.

### Environment

- `tinker-cookbook` @ `main` (also present on `nightly`; the two are identical for this file)
- dataset: `harbor datasets download terminal-bench@2.0`

miguelrc-scale:
Fixes #889.

### Problem

`HarborReward._upload_tests` read every file under a task's `tests/` directory with `Path.read_text()`. Terminal-Bench tasks whose graders compare against binary ground truth raise `UnicodeDecodeError` there, and `__call__` catches it and returns `reward=0.0, grading_error=1.0` — so those tasks score zero regardless of what the agent did.

7 of the 89 tasks in `terminal-bench@2.0` are affected: `build-pov-ray`, `make-doom-for-mips`, `make-mips-interpreter`, `pytorch-model-recovery`, `sam-cell-seg`, `train-fasttext`, `video-processing`. Their fixtures are load-bearing — `sam-cell-seg/tests/test.sh:25` does `cp /tests/test_img.png .`, and `make-mips-interpreter/tests/test_outputs.py:71` reads `/tests/reference.jpg` for an L2 image comparison.

Because the failure is swallowed rather than raised, `eval.py` classifies the returned `0.0` as `FAIL` rather than `ERROR`, so it is indistinguishable from a genuine miss. In RL the group's reward is constant zero, so advantages center to zero and the group yields no gradient.

### Change

- `_upload_tests` reads with `read_bytes()` instead of `read_text()`. `SandboxInterface.write_file` already accepts `str | bytes`, and `ModalSandbox.write_file` encodes `str` to bytes before streaming to stdin, so this is the path it was already built for — no sandbox-layer change needed.
- `FakeSandbox` now stores written content verbatim instead of eagerly `.decode()`-ing it, so the test double round-trips binary the way the real sandbox does. `read_file` decodes on the way out to keep returning `str`.
- New regression test `test_upload_tests_with_binary_fixture`.

### Verification

The new test fails on `main` with the same error seen in training logs:

```
E       KeyError: '/tests/reference.png'
ERROR   Harbor grading failed: 'utf-8' codec can't decode byte 0x89 in position 0: invalid start byte
1 failed, 10 passed
```

and passes with the fix:

```
11 passed in 2.27s
```

Wider suite: `pytest tinker_cookbook/recipes/ tinker_cookbook/tool_use/` gives 194 passed, 10 skipped. The remaining failures in my environment are pre-existing and unrelated to these files (`sdft_test.py` needs `pytest-asyncio`; `math_rl` needs the `math-rl` extra). `ruff format --check` and `ruff check` are clean on both files.

### Note

#673 (draft) also touches `harbor_tools.py` and `harbor_tools_test.py`. It does not modify `_upload_tests`, but it does add its own `FakeSandbox` usage, so there may be a textual conflict to resolve depending on merge order.

### DIFF
--- a/tinker_cookbook/recipes/harbor_rl/harbor_tools.py
+++ b/tinker_cookbook/recipes/harbor_rl/harbor_tools.py
@@ -93,12 +93,18 @@ class HarborReward:
         Walks ``tests_dir`` recursively so nested fixture trees (e.g. a
         ``tests/repo_overlay/`` directory copied into the workspace by
         ``test.sh``) are preserved. ``Path.iterdir`` would silently drop them.
+
+        Files are read as bytes because test fixtures are not always text:
+        several Terminal-Bench tasks ship reference images, serialized model
+        weights, or tarballs that their graders compare against. ``read_text``
+        raises ``UnicodeDecodeError`` on those, which ``__call__`` swallows
+        into a zero reward, making the task permanently unsolvable.
         """
         await self.sandbox.run_command("mkdir -p /tests", workdir="/")
         for file_path in self.tests_dir.rglob("*"):
             if not file_path.is_file():
                 continue
-            content = file_path.read_text()
+            content = file_path.read_bytes()
             target = f"/tests/{file_path.relative_to(self.tests_dir)}"
             await self.sandbox.write_file(target, content, executable=(file_path.suffix == ".sh"))
 
--- a/tinker_cookbook/recipes/harbor_rl/harbor_tools_test.py
+++ b/tinker_cookbook/recipes/harbor_rl/harbor_tools_test.py
@@ -19,7 +19,9 @@ class FakeSandbox:
     """In-memory sandbox for testing."""
 
     def __init__(self) -> None:
-        self.files: dict[str, str] = {}
+        # Values are stored exactly as written so binary fixtures round-trip,
+        # matching ModalSandbox.write_file, which streams raw bytes.
+        self.files: dict[str, str | bytes] = {}
         self.executable_files: set[str] = set()
         self.commands_run: list[str] = []
         self._command_results: dict[str, SandboxResult] = {}
@@ -51,13 +53,15 @@ class FakeSandbox:
         self, path: str, max_bytes: int | None = None, timeout: int = 60
     ) -> SandboxResult:
         if path in self.files:
-            return SandboxResult(stdout=self.files[path], stderr="", exit_code=0)
+            content = self.files[path]
+            stdout = content if isinstance(content, str) else content.decode(errors="replace")
+            return SandboxResult(stdout=stdout, stderr="", exit_code=0)
         return SandboxResult(stdout="", stderr=f"No such file: {path}", exit_code=1)
 
     async def write_file(
         self, path: str, content: str | bytes, executable: bool = False, timeout: int = 60
     ) -> SandboxResult:
-        self.files[path] = content if isinstance(content, str) else content.decode()
+        self.files[path] = content
         if executable:
             self.executable_files.add(path)
         return SandboxResult(stdout="", stderr="", exit_code=0)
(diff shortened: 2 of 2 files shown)


miguelrc-scale:
Flagging a behavioral difference a reviewer might reasonably ask about, since `read_text()` and `read_bytes()` are not equivalent for text: `read_text()` opens in text mode with universal newlines, so it translates `\r\n` → `\n`, while `read_bytes()` preserves them.

I checked what that actually changes across all 280 files under `*/tests/**` in `terminal-bench@2.0`:

| | files |
|---|---|
| Byte-identical either way | 272 |
| Undecodable as text (the crash this PR fixes) | 7 |
| Decodable but bytes differ | 1 |

The 272 are unchanged because `read_text()` decodes with the locale encoding and `ModalSandbox.write_file` re-encodes via `.encode()`; for UTF-8 text that round-trip is the identity.

The single differing file is `sparql-university/tests/university_graph_test.ttl` — 310 CRLF pairs, no lone CRs, so line endings are the only difference. Its grader parses it with `rdflib.Graph().parse(format="ttl")`. Both variants yield **295 triples** and `rdflib.compare.isomorphic()` is **True**, so grading behavior is unchanged.

**No `.sh` fixture is affected.** That was the case I most wanted to rule out: a script with CRLF uploaded with `executable=True` would previously have been normalized to LF, and preserving CRLF would break it under bash. No test script in the dataset has CRLF endings.

For that one `.ttl`, the previous behavior was arguably the incorrect one — it silently rewrote a fixture's line endings on the way into the sandbox. Reading as bytes uploads the grader's ground truth verbatim. Text mode introduces two other latent cases in the same direction that this also closes: transcoding a non-UTF-8 fixture to UTF-8 when the locale differs, and translating lone `\r`.

PR #892, opened 2026-08-15
Fix IFEval scoring every "less than" constraint as satisfied

### LINKED ISSUE #891: IFEval scores every "less than" constraint as satisfied
### Summary

`_relation_check` in `tinker_cookbook/eval/benchmarks/_ifeval_verify.py` doesn't know the `less than` relation, and its fallback is `True`, so every `less than` constraint in IFEval is scored as satisfied no matter what the model wrote. ifeval strict and loose accuracy both come out too high.

```python
def _relation_check(count: int, relation: str, target: int) -> bool:
    checks = {"at least": count >= target, "at most": count <= target, "exactly": count == target}
    return checks.get(relation, True)
```

### Why it matters

`google/IFEval` only ever uses two relations, and the missing one is the second most common. Counting every `*relation` kwarg in the 541-prompt `input_data.jsonl`:

| relation | occurrences | in `checks` |
| --- | --- | --- |
| `at least` | 129 | yes |
| `less than` | **75** | **no → scored as satisfied** |
| `at most` | 0 | yes |
| `exactly` | 0 | yes |

So the two relations that are implemented never appear in the dataset, and the one that appears 75 times isn't. Those 75 constraints land on **72 of the 541 prompts (13%)**, arriving as `relation` (57), `let_relation` (11) and `capital_relation` (7).

The value comes straight from the dataset kwargs into seven call sites, so this affects word count, sentence count, paragraph count, keyword frequency, letter frequency and capital-word frequency.

This also matches the reference implementation, where `_COMPARISON_RELATION = ("less than", "at least")` — `less than` is a first-class IFEval relation, not an edge case.

### Reproduction

```python
from tinker_cookbook.eval.benchmarks._ifeval_verify import _relation_check, verify_instruction

# 3 is not less than 3
_relation_check(3, "less than", 3)          # -> True   (expected False)

# "zebra zoo" has 2 z's, the constraint asks for fewer than 2
verify_instruction(
    "keywords:letter_frequency",
    "zebra zoo",
    {"letter": "z", "let_relation": "less than", "let_frequency": 2},
)                                            # -> True   (expected False)
```

### Notes

Nothing imports `_ifeval_verify` from the test suite today, so relation semantics have no coverage — which is probably why this survived the grading pass in #643.

I have a fix with tests ready and will open a PR referencing this issue.

winklemad:
Fixes #891.

`_relation_check` built its lookup from `{at least, at most, exactly}` and fell back to `True`, so `less than` — which isn't in the table — scored as satisfied whatever the model produced.

```diff
-    checks = {"at least": count >= target, "at most": count <= target, "exactly": count == target}
+    checks = {
+        "at least": count >= target,
+        "less than": count < target,
+        "at most": count <= target,
+        "exactly": count == target,
+    }
```

`google/IFEval` uses exactly two relations across its 541 prompts — `at least` (129) and `less than` (75) — so the missing one covers **72 of 541 prompts (13%)**, arriving as `relation` (57), `let_relation` (11) and `capital_relation` (7). `at most` and `exactly` never appear in that dataset; I left them in place since other IF datasets may use them. The relation reaches seven call sites, so this touched word, sentence and paragraph counts as well as keyword, letter and capital-word frequency.

## Tests

Added `TestIFEvalRelationCheck` to `benchmark_test.py` — nothing imported `_ifeval_verify` before, so relation semantics had no coverage:

- `less than` at, below and above the boundary, plus `at least` as a no-regression check
- one case per real call site through `verify_instruction`: `length_constraints:number_words`, `number_sentences`, `keywords:frequency`, `keywords:letter_frequency`, `change_case:capital_word_frequency`
- a sweep asserting no instruction type accepts a violating response under `less than`

The six `less than` tests fail on `main` and pass with this change. `uv run pytest tinker_cookbook/eval/` is 82 passed / 5 skipped (the skips are pre-existing — antlr4/sympy not installed), and `ruff check`, `ruff format --check` and `pyright` are clean on both files.

## One open question

I deliberately left the `checks.get(relation, True)` fallback alone rather than making an unknown relation raise, and pinned the current behaviour with a test. Two reasons:

1. `verify_instruction` handles instruction ids that don't exist in `google/IFEval` (`letters:letter_counting`, `keywords:palindrome`, `count:count_unique`, …), so the set of relation strings that can reach this function isn't closed — some come from IFBench-style data I can't audit from here.
2. It wouldn't actually fail loudly anyway: `verify_instruction` wraps its body in `except Exception: return False`, so a raise would be swallowed into a silent zero, which seems worse than a lenient pass.

If you'd rather an unrecognised relation be a hard error, I think the change belongs at that `except` boundary rather than here, and I'm happy to do it in this PR or a follow-up — just say which you'd prefer.


tinker_cookbook/eval/benchmarks/_ifeval_verify.py
```diff
@@ -19,7 +19,14 @@
...
-    checks = {"at least": count >= target, "at most": count <= target, "exactly": count == target}
+    # "at least" and "less than" are the only two relations google/IFEval uses;
+    # "at most" and "exactly" are here for datasets that use them.
+    checks = {
+        "at least": count >= target,
+        "less than": count < target,
+        "at most": count <= target,
+        "exactly": count == target,
+    }
     return checks.get(relation, True)
```
caiotheodoro: risk: `checks.get(relation, True)` is exactly how "less than" went unscored — any relation string not in this dict silently counts as satisfied instead of failing loudly. All 7 call sites pull `relation` straight from dataset kwargs (`kwargs.get("relation", "at least")`), and this file also backs `ifbench.py`, which has more instruction variants than plain IFEval. A typo'd or new relation string here inflates scores with no signal. Worth raising on an unrecognized relation instead of defaulting to `True`.

caiotheodoro (commented):
Confirmed the fix — "less than" was silently defaulting to satisfied before this.

tinker_cookbook/eval/benchmarks/_ifeval_verify.py
winklemad: Good catch — you're right, and it's the same failure this PR is about: defaulting an unverifiable relation to `True` silently inflates scores exactly the way "less than" did. Keeping the lenient fallback while fixing "less than" was inconsistent.

Fixed in 04411e3: `_relation_check` now returns `False` and logs a warning on an unrecognized relation, so a typo'd or new relation string surfaces instead of quietly passing. I went with warn+`False` rather than a bare `raise` since `verify_instruction` already wraps everything in a `try/except` that would turn a raise into a generic "Error verifying instruction" + `False` anyway — logging the specific relation is clearer. Updated the test that had asserted the old lenient behavior to assert it now fails loudly.

Issue #895, opened 2026-08-15
extract_boxed picks the first \boxed{} in eval but the last in math_rl


winklemad:
### Summary

The repo has two `extract_boxed` implementations that disagree on which `\boxed{}` wins when a response contains more than one, and both are used for grading.

| function | picks | docstring |
| --- | --- | --- |
| `tinker_cookbook/eval/benchmarks/_common.py:326` | the **first** (`text.find("\\boxed{")`) | "content inside the outermost `\boxed{}`" |
| `tinker_cookbook/recipes/math_rl/math_grading.py:191` | the **last** (`boxed_strs[-1]`) | "the context of the last `\boxed{...}`" |

Same input, different answers:

```python
text = r"First attempt gives \boxed{5}. That is wrong; correcting, the answer is \boxed{7}."

eval._common.extract_boxed(text)          # -> '5'
recipes.math_rl.extract_boxed(text)       # -> '7'
```

`_common.extract_boxed` also backs `extract_gsm8k_answer`, so the same first-vs-last choice applies to gsm8k grading.

### Why I'm asking rather than sending a patch

The eval-side docstring says "outermost", which describes nesting rather than position, so I can't tell from the code whether taking the first match is deliberate or incidental. It matters when a model reasons out loud and boxes an intermediate or self-corrected value before its final one — the last box is the conventional reading, and it's what the math-RL path already does.

I have **not** measured how often real completions contain multiple `\boxed{}` spans in these benchmarks, so I'm not claiming a specific score impact. What is certain is that the two graders disagree on identical input.

### Question

Which is authoritative for grading — first or last? If last is intended for the eval path, I'm happy to send a small PR switching `_common.extract_boxed` to a last-occurrence scan and adding tests; I checked it against every existing boxed assertion in the repo (`benchmark_test.py` and `math_env_test.py`) plus nested, empty and unclosed inputs, and they all still pass. If first is intended, then the two docstrings could probably say so explicitly.

IgorGanapolsky:
## Repro note

Leaderboard and cookbook scores only stay comparable when **protocol is pinned** (system prompt, max_tokens, temperature, timeout, renderer, thinking-token handling). Tinker’s eval README already flags this sensitivity — linking each public result to the exact `BenchmarkConfig` + renderer avoids “verification” against a different protocol.

For agent post-training we also separate:

- **LLM-as-a-Judge / rubrics** for open-ended quality
[...]

PR #903, opened 2026-08-18
[eval] extract_boxed takes the last box, matching math_rl

mmjerge:
Addresses #895.

`tinker_cookbook/eval/benchmarks/_common.py` and `tinker_cookbook/recipes/math_rl/math_grading.py` both grade `\boxed{}` answers, and they disagreed on which span wins: `_common.extract_boxed` scanned with `text.find` and returned the **first** box, while `math_grading.extract_boxed` returns the **last**. A model that boxes an intermediate or self-corrected value before its final answer was graded on the attempt it had discarded.

```python
text = r"First attempt gives \boxed{5}. That is wrong; correcting, the answer is \boxed{7}."
eval._common.extract_boxed(text)     # '5'
math_rl.extract_boxed(text)          # '7'
```

## Why last-wins

- It is already the convention in this module: `extract_number`, `extract_gsm8k_answer` and `extract_mcq_answer` all take the last match. `extract_boxed` was the only exception.
- It is what the training path does, so eval and RL now grade identically.
- The docstring said "outermost", which describes nesting rather than position, so the first-box scan reads as an oversight rather than a decision. The new docstring states position and nesting separately.

This reaches past the math benchmarks: `extract_mcq_answer` calls `extract_boxed` (`_common.py:437`), so `mmlu_pro`, `gpqa`, `mmlu_redux`, `ceval` and `supergpqa` were subject to the same first-vs-last choice.

## What changed

`extract_boxed` scans for the last *closed* span. Nesting still resolves to the outermost (`\boxed{\boxed{3}}` → `\boxed{3}`), because a closed span resumes the scan past its own end. An unclosed span resumes just inside itself, so it no longer hides a well-formed span on either side of it — `\boxed{42} and \boxed{unclosed` and `\boxed{unclosed and \boxed{42}` both yield `42`, where the second returned `None` before.

This shifts grading output for any response that contains more than one box, so cached eval results produced before this change would move.

## Tests

- New `TestExtractBoxed`: single, multiple, self-correction, nested braces, nested boxes, empty, absent, unclosed, and unclosed-alongside-closed.
- New `test_extract_boxed_matches_math_rl` in `TestGradingConsistency`, asserting the two implementations agree on the same inputs.

The three existing `TestGradingConsistency` tests all `importorskip` `recipes.nemotron_cascade`, which is no longer in the repo, so the class skipped in full and guarded nothing. The new test runs whenever the math-rl extras are installed. It catches the bare `ImportError` that `math_grading` raises when those extras are absent, which `pytest.importorskip` does not intercept, and skips instead.

Every existing boxed assertion in `benchmark_test.py` and `math_env_test.py` still passes; `ruff check`, `ruff format --check` and `pytest tinker_cookbook/eval/benchmarks/` are clean locally.

## Note

#895 asked which behaviour is authoritative and that question is still open. This is the branch I offered there — if first-wins was in fact intended, close this and the fix is a docstring on each side saying so. Happy for @winklemad to take it over or for a maintainer to reshape it.

### DIFF
--- a/tinker_cookbook/eval/benchmarks/_common.py
+++ b/tinker_cookbook/eval/benchmarks/_common.py
@@ -324,24 +324,42 @@ class SandboxMixin:
 
 
 def extract_boxed(text: str) -> str | None:
-    r"""Extract content from ``\boxed{...}`` handling nested braces.
+    r"""Extract content from the last ``\boxed{...}`` handling nested braces.
+
+    When a response contains several ``\boxed{}`` spans -- e.g. a model boxes an
+    intermediate result, or a self-corrected value, before its final answer --
+    the last one is the answer. This matches the convention used by
+    ``recipes.math_rl.math_grading.extract_boxed`` and by the other extractors
+    in this module, which all take the last match.
+
+    Nested spans resolve to the outermost: ``\boxed{\boxed{3}}`` yields
+    ``\boxed{3}``. An unclosed span is skipped, so a well-formed earlier span is
+    still found.
 
     Returns:
-        The content inside the outermost ``\boxed{}``, or ``None`` if not found.
+        The content inside the last closed ``\boxed{}``, or ``None`` if there is none.
     """
-    idx = text.find("\\boxed{")
-    if idx == -1:
-        return None
-    start = idx + len("\\boxed{")
-    depth = 1
-    i = start
-    while i < len(text) and depth > 0:
-        if text[i] == "{":
-            depth += 1
-        elif text[i] == "}":
-            depth -= 1
-        i += 1
-    return text[start : i - 1] if depth == 0 else None
+    marker = "\\boxed{"
+    result: str | None = None
+    idx = text.find(marker)
+    while idx != -1:
+        start = idx + len(marker)
+        depth = 1
+        i = start
+        while i < len(text) and depth > 0:
+            if text[i] == "{":
+                depth += 1
+            elif text[i] == "}":
+                depth -= 1
+            i += 1
+        if depth == 0:
+            # Closed span: record it and resume past it so nesting keeps the outermost.
+            result = text[start : i - 1]
+            idx = text.find(marker, i)
+        else:
+            # Unclosed span: resume just inside it, in case a later span is well-formed.
+            idx = text.find(marker, start)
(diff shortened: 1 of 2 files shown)


winklemad:
Please keep it — you did the work and you found a better argument than the one I opened the issue with. I raised #895 as a question precisely because I couldn't show how often models emit multiple boxes, and "every other extractor in this module already takes the last match" settles the direction far more convincingly than incidence data would have.

I checked that claim rather than take it on trust, and it holds: `bold_numbers[-1]`, `numbers[-1]`, `letters[-1]` and `matches[-1]` are all last-wins, so `extract_boxed` was the lone exception among five. Your point about it not being math-only also holds — `extract_boxed` has two callers, the gsm8k path and the MCQ path, so the MCQ grader inherits the same behaviour.

I ran your implementation over the edge cases I'd worry about and it behaves:

| input | result |
| --- | --- |
| `\boxed{42}` | `42` |
| `\boxed{5}` … `\boxed{7}` | `7` |
| `\boxed{\boxed{3}}` | `\boxed{3}` (outermost) |
| `\boxed{\frac{3}{4}}` | `\frac{3}{4}` |
| `\boxed{}` | `""` |
| `\boxed{42` (unclosed) | `None` |
| `\boxed{5}` then `\boxed{7` | `5` |

That last row is the one I'd have got wrong: resuming past a *closed* span but only just inside an *unclosed* one is what makes a trailing broken box fall back to the good earlier answer instead of losing it. Worth keeping the comment that explains it.

Good catch on `TestGradingConsistency` too. If those tests `importorskip` a module that no longer exists, the class has been silently skipping rather than guarding, which is its own small bug.

One thing for whoever reviews: this changes grading output for any stored response containing more than one box, so if there are cached eval results anywhere they'd shift. Probably worth a sentence in the PR body so it isn't a surprise later.

mmjerge:
Thanks for going over this so carefully, particularly for re-checking the last-wins claim against the other four extractors rather than taking my word for it, and for running the edge cases independently. Good to have the closed-then-unclosed row confirmed by someone other than me. I've left the comment explaining the resume position in place.

Added a sentence to the PR body about cached eval results shifting, under "What changed".

PR #911, opened 2026-08-20
make ifeval number_bullet_lists require exactly N bullets

confiscate:
The detectable_format:number_bullet_lists instruction asks for "exactly N bullet points," but the verifier used >=, so extra bullets still passed.

Dataset source (google/IFEval, the prompts the model is graded against):

["…exactly 3 bullet points in the markdown format…" (`num_bullets=3`)](https://datasets-server.huggingface.co/rows?dataset=google/IFEval&config=default&split=train&offset=5&length=1)
["…exactly 6 bullet point in Markdown…" (`num_bullets=6`)](https://datasets-server.huggingface.co/rows?dataset=google/IFEval&config=default&split=train&offset=56&length=1)


confiscate:
cc @dphuang2

PR #939, opened 2026-09-10
Add validation AUC to the forecasting results tables

nealwu:
Adds validation AUC to the forecasting recipe, both as a column in the README results tables and as a metric logged during training.

**Computation**: `ForecastEvaluator`, an `RLTestSetEvaluator` subclass, reports Mann-Whitney AUC as `test/env/all/auc` and per category as `test/env/<tag>/auc`, on both the returned metrics and `last_result`. Because the training loop adds a plain `RLTestSetEvaluator` whenever the dataset builder returns a validation split, `ProphetArenaRLDatasetBuilder` gains `include_validation`, and `cli_main` passes a training-only copy to the loop while supplying `ForecastEvaluator` through `evaluator_builders`. The final-checkpoint evaluation uses the same class, so all nine evals report AUC.

**Results**: (256 validation questions × 8 forecasts, step 0 → 128): Qwen3.8-27B AUC `0.7959` → `0.8175`; GLM-5.3 `0.8088` → `0.8495`. Training improves discrimination for both models, with GLM's gain roughly twice Qwen's. AUC is calibration-blind, so it complements Brier rather than decomposing it.

**README**: Adds a `Validation AUC` column to both results tables along with a definition.


feifeiobama (approved):
lgtm!

tinker_cookbook/recipes/forecasting/env.py
```diff
@@ -194,6 +200,84 @@ def logging_tags(self) -> list[str]:
...
+        rollout_summary_export: RolloutSummaryExportConfig | None,
+        *,
+        store: TrainingRunStore | None = None,
+    ) -> dict[str, float]:
+        metrics = super()._collect_eval_metrics(results, rollout_summary_export, store=store)
+        pairs_by_tag: dict[str, list[tuple[float, int]]] = defaultdict(list)
+        for builder, group in zip(self.env_group_builders_P, results):
+            if group is None:
+                continue
+            pair = _group_forecast(group)
```
feifeiobama: `_group_forecast` averages the eight forecasts for each question before computing AUC. shall we just use row forecasts?

feifeiobama (approved):
lgtm

PR #941, opened 2026-09-11
Fix SL resume replaying the last checkpointed batch

### LINKED ISSUE #935: SL resume replays the last checkpointed batch, applying a duplicate optim_step on the same data
## Summary

In `tinker_cookbook/supervised/train.py`, the checkpoint saved during training records the batch that was *just applied*, but resume logic treats that same index as the *next* batch to run. Every resume of an SL training run re-executes (forward_backward + optim_step) the last batch that was trained before the checkpoint, applying a second gradient update on that data before continuing.

Tested against commit `1f962eda3a2cec8de284725f2adc9978e93dfcd3`.

## Root cause

In `main()`, the checkpoint's `batch` field is written from `finish_batch`, *after* that batch's `forward_backward`/`optim_step` requests have already been submitted to the training client:

```python
# tinker_cookbook/supervised/train.py
async def submit_batch(epoch_idx: int, batch_idx: int) -> SubmittedBatch:
    ...
    fwd_bwd_future = await training_client.forward_backward_async(data, loss_fn="cross_entropy")
    optim_step_future = await training_client.optim_step_async(adam_params)
    return SubmittedBatch(..., batch_idx=batch_idx, ...)

async def finish_batch(submitted: SubmittedBatch):
    ...
    elapsed_tokens += submitted.num_tokens
    await checkpoint_mgr.maybe_save_async(          # supervised/train.py:463
        step=submitted.step,
        loop_state={
            "epoch": submitted.epoch_idx,
            "batch": submitted.batch_idx,            # supervised/train.py:467
            "elapsed_tokens": elapsed_tokens,
        },
        elapsed_tokens=elapsed_tokens,
    )
    async with trace.scope_span("step"):
        fwd_bwd_result = await submitted.fwd_bwd_future.result_async()
        optim_step_result = await submitted.optim_step_future.result_async()
```

`checkpoint_mgr.maybe_save_async` ends up calling `training_client.save_state_async`, which — like `forward_backward_async` and `optim_step_async` — gets a monotonically increasing request id from `TrainingClient._get_request_id()`:

```python
# tinker/lib/public_interfaces/training_client.py
# Reserves a request id for a request. Requests are to be executed in the order of request ids.
def _get_request_id(self) -> int:
```

Since `forward_backward_async`/`optim_step_async` for `batch_idx` were already submitted (in `submit_batch`, which always runs to completion before `finish_batch` is invoked on the same `SubmittedBatch`), `save_state_async`'s request id is strictly greater, so the saved state reflects weights *after* `batch_idx`'s optim step has been applied.

On resume:

```python
# tinker_cookbook/supervised/train.py:290-299
resume_info = checkpoint_utils.get_last_checkpoint(config.log_path)
if resume_info:
    start_epoch = resume_info.epoch or 0
    start_batch = resume_info.batch
    ...
# (start_epoch, start_batch) now represent the next batch to execute if resuming.
...
start_batch_idx = start_batch if epoch_idx == start_epoch else 0
for batch_idx in range(start_batch_idx, n_batches):
```

`start_batch` is used unmodified as "the next batch to execute", but it actually names a batch whose gradient update is already baked into the loaded weights. `range(start_batch_idx, n_batches)` re-includes it, so it gets trained on a second time on top of a model that already learned from it once, and `elapsed_tokens` (also saved pre-increment-consistent, at the same instant) gets double-counted for that batch too.

## Why this is wrong, contrasted with the rest of the repo

The RL trainer (`tinker_cookbook/rl/train.py`) has the identical "checkpoint after the batch has already been submitted" structure, but explicitly corrects for it:

```python
# tinker_cookbook/rl/train.py:1676
# NOTE: saving the checkpoint as the i + 1 step
sampling_client, full_batch_metrics = await compute_full_batch_metrics_and_get_sampling_client(
    training_client, checkpoint_mgr,
    i_batch + 1,   # <-- +1, not i_batch
    ...
)
```
and in the rolling-checkpoint path (`rl/train.py:1237-1238`):
```python
await checkpoint_mgr.maybe_save_rolling_async(
    step=i_batch + 1, loop_state={"batch": i_batch + 1}
)
```

DPO (`tinker_cookbook/preference/train_dpo.py:302-309`) uses the opposite, also-correct convention: it checkpoints *before* fetching/processing the batch, so `batch_idx` genuinely is the next batch to run:

```python
step = epoch_idx * n_batches + batch_idx
with trace.trace_iteration(step=step) as window:
    if checkpoint_mgr is not None:
        save_result = checkpoint_mgr.maybe_save(
            step=step, loop_state={"epoch": epoch_idx, "batch": batch_idx}
        )
    ...
    data = dataset.get_batch(batch_idx)   # fetched AFTER the checkpoint
```

`supervised/train.py` is the outlier: it checkpoints after submission (like RL) but records the raw `batch_idx` (like DPO), combining the wrong halves of both correct patterns.

## Reproduction (offline, no TINKER_API_KEY / network needed)

The script below drives the real `tinker_cookbook.supervised.train.main()` event loop against a fake in-process `TrainingClient`/`ServiceClient` (no network) that just logs call order. It simulates a 10-batch, `save_every=5` SL run that crashes right after checkpointing batch 5, then resumes.

```python
# repro_checkpoint_offbyone.py
import asyncio, logging, shutil
from pathlib import Path
from typing import Any
import chz, tinker
from tinker_cookbook.supervised import train
from tinker_cookbook.supervised.types import SupervisedDataset, SupervisedDatasetBuilder

logging.basicConfig(level=logging.WARNING)
EVENTS: list[tuple[str, Any]] = []

def make_datum(batch_idx: int) -> tinker.Datum:
    return tinker.Datum(
        model_input=tinker.ModelInput.from_ints([1]),
        loss_fn_inputs={
            "weights": tinker.TensorData(data=[1.0], dtype="float32", shape=[1]),
            "target_tokens": tinker.TensorData(data=[100 + batch_idx], dtype="int64", shape=[1]),
        },
    )

class SimulatedCrash(Exception): pass

class FakeDataset(SupervisedDataset):
    def __init__(self, n, crash_at_batch=None):
        self.n, self.crash_at_batch = n, crash_at_batch
    def get_batch(self, index):
        if self.crash_at_batch is not None and index == self.crash_at_batch:
            raise SimulatedCrash(f"crash before batch {index}")
        EVENTS.append(("get_batch", index))
        return [make_datum(index)]
    def __len__(self): return self.n
    def set_epoch(self, seed=0): pass

@chz.chz
class FakeDatasetBuilder(SupervisedDatasetBuilder):
    n_batches: int = 10
    crash_at_batch: int | None = None
    def __call__(self):
        return FakeDataset(self.n_batches, self.crash_at_batch), None

class FakeTokenizer:
    def decode(self, ids, **kw): return " ".join(str(i) for i in ids)
    def encode(self, s, add_special_tokens=True): return [0]

class FakeResult:
    def __init__(self, **kw): self.__dict__.update(kw)
class FakeFuture:
    def __init__(self, r): self._r = r
    async def result_async(self): return self._r

class FakeTrainingClient:
    model_id = "fake-model"
    async def forward_backward_async(self, data, loss_fn, loss_fn_config=None):
        idx = data[0].loss_fn_inputs["target_tokens"].data[0] - 100
        EVENTS.append(("forward_backward_submit", idx))
        n = len(data[0].loss_fn_inputs["target_tokens"].data)
        return FakeFuture(FakeResult(loss_fn_outputs=[
            {"logprobs": tinker.TensorData(data=[0.0]*n, dtype="float32", shape=[n])}]))
    async def optim_step_async(self, adam_params):
        EVENTS.append(("optim_step_submit", None))
        return FakeFuture(FakeResult(metrics={}))
    async def save_state_async(self, name, ttl_seconds=None):
        EVENTS.append(("save_state_submit", name))
        return FakeFuture(FakeResult(path=f"tinker://fake-run/state/{name}"))
    async def save_weights_for_sampler_async(self, name, ttl_seconds=None):
        return FakeFuture(FakeResult(path=f"tinker://fake-run/sampler/{name}"))

class FakeServiceClient:
    def __init__(self, *a, **kw): pass
    async def create_lora_training_client_async(self, base_model, rank, user_metadata=None):
        return FakeTrainingClient()
    async def create_training_client_from_state_with_optimizer_async(self, state_path, user_metadata=None):
        EVENTS.append(("resume_from_state", state_path))
        return FakeTrainingClient()

async def run(log_path, label, crash_at_batch):
    config = train.Config(
        log_path=log_path, model_name="fake/model", recipe_name="repro",
        renderer_name=None, dataset_builder=FakeDatasetBuilder(n_batches=10, crash_at_batch=crash_at_batch),
        num_epochs=1, save_every=5, eval_every=0, infrequent_eval_every=0,
        wandb_project=None, submit_ahead=0,
    )
    try:
        await train.main(config)
    except SimulatedCrash as e:
        EVENTS.append(("simulated_crash", str(e)))

def main():
    log_dir = "/tmp/tc_resume_repro"
    shutil.rmtree(log_dir, ignore_errors=True); Path(log_dir).mkdir(parents=True)
    tinker.ServiceClient = FakeServiceClient
    train.tinker.ServiceClient = FakeServiceClient
    train.get_tokenizer = lambda model_name: FakeTokenizer()
    asyncio.run(run(log_dir, "run1", crash_at_batch=6))
    asyncio.run(run(log_dir, "run2 (resume)", crash_at_batch=None))
    submitted = [idx for name, idx in EVENTS if name == "forward_backward_submit"]
    print("batches sent to forward_backward, in order:", submitted)
    print("duplicated batches:", {b for b in submitted if submitted.count(b) > 1})

if __name__ == "__main__":
    main()
```

Actual captured output:

```
batches sent to forward_backward, in order: [0, 1, 2, 3, 4, 5, 5, 6, 7, 8, 9]
duplicated batches: {5}
```

Batch 5 is trained twice: once in the run that gets interrupted (its `forward_backward`+`optim_step` are submitted, then the checkpoint recording `"batch": 5` is written), and again immediately after resuming from that same checkpoint (`resume_from_state`), before batch 6 is ever reached.

## Related: the regression test for this never runs

`tinker_cookbook/supervised/resume_test.py` contains a hand-written checkpoint/resume correctness check (`def checkpoint_resume():`) that compares losses across a simulated crash+resume — exactly the scenario above — but the function is not named `test_checkpoint_resume`, so `pytest` (`python_functions` defaults to `test_*`, per `pyproject.toml`'s `[tool.pytest.ini_options]`) collects zero tests from this file:

```
$ python -m pytest tinker_cookbook/supervised/resume_test.py --collect-only
collected 0 items
```

`pytest tinker_cookbook/` — the documented unit-test command in `AGENTS.md`/`CONTRIBUTING.md` — silently never runs this check, and it also isn't runnable from `tests/` since it needs `TINKER_API_KEY` unlike the rest of `tinker_cookbook/`'s colocated tests. That's a separate defect but explains why this off-by-one was never caught in CI: the one test written specifically to catch it has never executed.

## Suggested fix

Either move the checkpoint save in `finish_batch` to before `submit_batch` fires the batch's requests (DPO's pattern), or record `submitted.batch_idx + 1` in `loop_state["batch"]` (RL's pattern) so the invariant documented at `supervised/train.py:299` ("(start_epoch, start_batch) now represent the next batch to execute if resuming") actually holds. Separately, renaming `checkpoint_resume` to `test_checkpoint_resume` (or adding a `pytest.mark` entry point) would let this class of regression get caught automatically going forward.

## Environment

- tinker-cookbook commit: `1f962eda3a2cec8de284725f2adc9978e93dfcd3`
- Python 3.14.7, `tinker==0.27.2`, `torch==2.14.0`
- macOS (Apple Silicon), no GPU/TINKER_API_KEY used for this reproduction — the repro fakes the `tinker` service client entirely and never touches the network.

### LINKED ISSUE #940: SL checkpoint under default submit_ahead=1 can snapshot the next batch
## Summary

Follow-up to #935. That issue documents the off-by-one: SL `loop_state["batch"]` names a batch whose `optim_step` is already on Tinker's ordered request queue, and resume treats it as the next batch to run.

There is a second, independent failure under the **default** pipeline. Cookbook SL uses `submit_ahead=1`. #935's repro set `submit_ahead=0`, so it did not catch this.

A naive `batch + 1` write (RL's convention) fixes #935's `submit_ahead=0` case and still retrains a batch after resume from a default run.

## Root cause

Tinker executes training RPCs in request-id order (`TrainingClient._get_request_id`). Periodic SL checkpoints call `save_state_async`, which takes a request id like `forward_backward_async` / `optim_step_async`.

With `submit_ahead=1`, `submit_batch(N+1)` runs **before** `finish_batch(N)`. If `maybe_save_async` stays in `finish_batch`, `save_state` for a checkpoint labeled N can be queued **after** N+1's train ops. The weights named `save_state:00000N` then include N **and** N+1.

Resume then does `range(start_batch, n_batches)`:

- Recording `batch = N` (today) retrains N, and the loaded weights may already include N+1.
- Recording `batch = N + 1` (RL-style, the #935 suggestion) skips N but still retrains N+1 on weights that already applied it.

DPO is not pipelined this way: it checkpoints **before** fetching the batch, so `batch_idx` really is the next batch. RL writes `i_batch + 1` and is not using this SL `submit_ahead` queue. SL combined the wrong halves of both patterns.

## Why CI missed it

`tinker_cookbook/supervised/resume_test.py` defined `checkpoint_resume`, not `test_*`. Pytest never collected it (`python_functions` defaults to `test_*`). The NLL comparison in that helper was also the wrong assertion for this bug.

## Suggested fix

1. Enqueue periodic `save_state` at the end of `submit_batch`, immediately after that batch's `optim_step_async`, so the checkpoint sits on the request queue before any later batch.
2. Persist the **next** `(epoch, batch)` in `loop_state` (epoch wrap when `batch == n_batches`), matching the comment in `main()` and RL.
3. Collect real `test_*` coverage with `submit_ahead=0` **and** `1`, using an in-process fake `TrainingClient` (no `TINKER_API_KEY` / network).

Do not also bump the resume *read* path to `start_batch + 1` if the write side already stores next-batch — that would skip a batch.

## Related

- #935 (off-by-one write vs resume-as-next; `submit_ahead=0` repro)
- Rolling checkpoints (`rolling_save_every`) and `async_periodic_saves=True` still defer `save_state` onto a later request id; that is a separate follow-up, not required to fix default SL resume.

rohith500:
## Summary

Supervised resume trains the last checkpointed batch a second time, applying a duplicate `optim_step` on data already baked into the saved weights (#935). Under the default `submit_ahead=1` pipeline the failure is worse: `save_state` queued from `finish_batch(N)` can run after batch N+1's train ops, so a naive `batch + 1` write still retrains N+1.

## Problem

`checkpoints.jsonl` recorded the batch that had already been enqueued (`loop_state["batch"] = submitted.batch_idx`), while resume treats `(epoch, batch)` as the **next** batch to run (`range(start_batch, n_batches)`). Tinker executes by monotonic request id, so the saved weights already include that batch.

Cookbook default is `submit_ahead=1`. `submit_batch(N+1)` can run before `finish_batch(N)`, so a checkpoint labeled N can contain N **and** N+1.

The colocated check in `resume_test.py` was named `checkpoint_resume`, so pytest never collected it.

## Fix

- Enqueue periodic `save_state` at the end of `submit_batch`, immediately after that batch's `optim_step_async`, so the checkpoint sits on the request queue before any later batch.
- Persist the next `(epoch, batch)` via `next_supervised_loop_state` (epoch wrap when `batch == n_batches`), matching RL and the comment in `main()`.
- Replace the uncollected helper with `test_*` cases that drive `train.main` against an in-process fake `TrainingClient` (no network / no `TINKER_API_KEY`), covering `submit_ahead=0` and `1`, epoch wrap, and a complete run.

No `Config` or public API changes. Rolling checkpoints and `async_periodic_saves=True` still have deferred request-id risk; left as follow-up.

## Tests

```
uv run pytest tinker_cookbook/supervised
```

After a crash at `save_every=5`, resume trains `[6, 7, 8, 9]`, not 5. `save_state:000005` is immediately after that batch's `forward_backward` / `optim_step`. The checkpoint record has `batch == 6`, `epoch == 0`.

## Checklist

- [x] Feature branch from `main`
- [x] Colocated `test_*` (no network)
- [x] CHANGELOG `fix` entry
- [x] `ruff` / `pyright` / `pytest tinker_cookbook/supervised` / `pre-commit` on changed files

Fixes #935
Fixes #940


rohith500:
CI note: `pytest (unit)` is red for the same reason as current `main` (#939): `test_sdk_client_instrumentation_covers_all_async_methods` — `TrainingClient.save_weights_external_async` is not in `_instrument_sdk_clients`. That is unrelated to this change.

This PR's new tests are green: `tinker_cookbook/supervised/resume_test.py .......` (7 passed). `pre-commit`, `type-check`, `downstream-compat`, `check-links`, and `stack-freshness` are green.

rohith500:
Pushed a one-line follow-up for the pytest tripwire: tinker 0.28 added `TrainingClient.save_weights_external_async`, and `_instrument_sdk_clients` did not wrap it. That is the same red test as current `main` (#939), not caused by the resume change.

PR #944, opened 2026-09-11
Stop Harbor retries from leaking a full sandbox group

### LINKED ISSUE #943: Harbor retry make_envs() leaks a full sandbox group (and drops the previous generation from cleanup)
## Summary

`RetryOnFailure` and `MinViableGroup` (Inkling's default) retry a failed rollout by calling `EnvGroupBuilder.make_envs()` and keeping only `[0]`. Harbor's `make_envs()` also does `self._sandboxes = []` **without** cleaning the previous list. Together that leaks live Modal (or other) sandboxes on every retry: unused extras from the new group, plus the entire previous generation that `cleanup()` can no longer see.

## Why this matters

Harbor RL retries are common (sandbox flakes, command timeouts). Inkling defaults to `MinViableGroup`. Each retry currently:

1. Provisions a **full `group_size`** of new sandboxes to replace **one** env.
2. Drops Harbor's pointer to every sandbox from the original `make_envs()` call, so `do_group_rollout`'s final `cleanup()` never terminates them.

The comment in `RetryOnFailure` says unused sandbox containers "get GC'd". That is false for Harbor: sandboxes require explicit `cleanup()`. Result is quota burn, cost, and eventual create failures mid-run.

## Root cause

```python
# tinker_cookbook/rl/rollout_strategy.py (RetryOnFailure / MinViableGroup)
new_envs = await env_group_builder.make_envs()
new_env = new_envs[0]
```

```python
# tinker_cookbook/recipes/harbor_rl/harbor_env.py
async def make_envs(self) -> Sequence[Env]:
    self._sandboxes = []   # previous generation is no longer cleaned
    ...
    for _ in range(self.group_size):
        sandbox = await self.sandbox_factory(...)
        self._sandboxes.append(sandbox)
```

You cannot "fix" this by cleaning all `_sandboxes` at the start of `make_envs`: the original group's envs are still in flight. The list must accumulate until the builder's `cleanup()`.

## Repro (offline, no Modal / no API key)

A Harbor-like builder that resets `_sandboxes = []` without cleanup, `group_size=4`, one scripted failure, `RetryOnFailure(max_retries=2)`:

- **Before:** `created=8 cleaned=4 leaked=4` (or `created=12` if a second retry runs)
- **After:** retry calls `make_env()` (one sandbox) and Harbor appends; `created == cleaned`

## Suggested fix

1. Add `EnvGroupBuilder.make_env()` — default delegates to `make_envs()[0]`; Harbor overrides to provision **one** sandbox and append it.
2. `RetryOnFailure` / `MinViableGroup` call `make_env()` instead of `make_envs()[0]`.
3. Harbor must **not** replace `_sandboxes` at the start of `make_envs()`.

No `Config` / public training API change. Existing builders keep working via the default `make_env`.

rohith500:
## Summary

Harbor RL retries leak live sandboxes. `RetryOnFailure` / `MinViableGroup` (Inkling default) called `make_envs()` for a single replacement env and kept `[0]`. Harbor also did `self._sandboxes = []` without cleanup, so `cleanup()` never saw the previous generation. The "unused containers get GC'd" comment is false for Modal.

Fixes #943.

## Fix

- Add `EnvGroupBuilder.make_env()` (default: `make_envs()[0]`). Harbor overrides it to provision **one** sandbox.
- Retry strategies call `make_env()`.
- Harbor **appends** sandboxes until `cleanup()`; it no longer drops the previous list while original envs are still in flight.

No `Config` or public training API change.

## Tests

- Retry creates 3 envs for a group of 2 + one failure, not 4
- Harbor `make_envs()` twice still cleans both generations
- Harbor `make_env()` after a group is one extra sandbox, all cleaned

```
uv run pytest tinker_cookbook/recipes/harbor_rl/harbor_tools_test.py \
  tinker_cookbook/rl/rollout_error_resilience_test.py \
  tinker_cookbook/rl/min_viable_group_test.py
```

## Checklist

- [x] Feature branch from `main`
- [x] Colocated `test_*` (no network / no Modal)
- [x] CHANGELOG `fix` entry


rohith500:
Cherry-picked the `save_weights_external_async` tracing tripwire so this PR is not red for the same pre-existing `main` failure as #939 / #941.

PR #950, opened 2026-09-16
Name both models in the forecasting README intro

nealwu:
Update in the first line that the recipe fine-tunes Qwen3.8-27B or GLM-5.3, and drop the note that `glm5_3_max_reasoning` was untested now that it has been run.


feifeiobama (approved):
lgtm

simonguozirui (approved):

